PR #2, opened 2015-09-17
Update debugger interop with stuff from latest dbgeng.h/cvconst.h headers.

jazzdelightsme:
Also collapse the managed definition of WINDBG_EXTENSION_APIS32 and
WINDBG_EXTENSION_APIS64, since they are defined the same.


leculver:
This is a bit tough.  I like the changes, but this is a breaking change for people using the APIs.  They will need to update their code if calling one of the changed apis.  I don't see a good way to keep the old code and mark it obsolete though, so this will have to be ok.

PR #5, opened 2015-09-23
Expose System and Shared Domain similar to other AppDomains

stefangossner:
To allow enumeration so modules loaded into System and Shared Domain it
is required to expose these two App Domains

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/Desktop/runtimebase.cs
+++ b/src/Microsoft.Diagnostics.Runtime/Desktop/runtimebase.cs
@@ -284,40 +284,6 @@ namespace Microsoft.Diagnostics.Runtime.Desktop
         }
 
         /// <summary>
-        /// The system domain in CLR.
-        /// </summary>
-        public DesktopAppDomain SystemDomain
-        {
-            get
-            {
-                if (_domains == null)
-                    InitDomains();
-
-                if (_system == null)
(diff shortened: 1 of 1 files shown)


leculver:
This pull request adds and removes a change, netting +/- 0 lines.  =)

I'm not opposed to this change.  I like that they are not enumerated with the other domains.  I'd like to look and see if there is anything missed with shared/system domains.  However, please submit a new pull request which also captures additions to public.cs.

stefangossner:
Hi Lee,
thats why I removed it as I noticed that public.cs should also be included.
Will send a new pull tomorrow.
Cheers,
Stefan

Von meinem Windows Phone gesendet
________________________________
Von: Lee Culver<mailto:notifications@github.com>
Gesendet: ‎23.‎09.‎2015 19:25
An: Microsoft/clrmd<mailto:clrmd@noreply.github.com>
Cc: stefangossner<mailto:webmaster@stefan-gossner.de>
Betreff: Re: [clrmd] Expose System and Shared Domain similar to other AppDomains (#5)


This pull request adds and removes a change, netting +/- 0 lines. =)

I'm not opposed to this change. I like that they are not enumerated with the other domains. I'd like to look and see if there is anything missed with shared/system domains. However, please submit a new pull request which also captures additions to public.cs.

PR #18, opened 2015-09-27
Apply DllImportSearchPath.LegacyBehavior to dbgeng/dbghelp PInvoke defs.

jazzdelightsme:
Some projects that use ClrMd have two modes of operation:

1.	Standalone mode.
2.	Hosted as an extension inside another program (such as windbg).

In standalone mode, I distribute a copy of the core debugger binaries
(dbgeng.dll, dbghelp.dll, et al) along with my binaries. But when I’m in
"hosted" mode, I need my code, including ClrMd, to use the
dbgeng.dll/dbghelp.dll that are already loaded into the process (instead of
loading a separate copy--the ones I redist).

To make this happen, I added a DefaultDllImportSearchPaths attribute to key
PInvoke definitions.


leculver:
I have merged this pull request manually as change 818486861d3de21c4a4b39f1517aebe8d72cea12.  Thanks for the submission, and sorry for the delay!  I couldn't find documentation on this while I was working on it on vacation, so it got put on pause.

Issue #24, opened 2015-10-08
Object and field representation (dynamic and otherwise)


leculver:
ClrMD should provide a more sensible way to access fields, values, and so on.  Instead of relying on "object GetValue".  Here is some of the previous discussion (edited for brevity and clarity).  I will refrain from putting my opinion here.

There are several primary topics of discussion for this design:
1. Interior flag.
2. Getting a single field/object value.
3. Getting a chain of values.
4.  Enumeration array contents.
5.  Enumeration of advanced containers. (List/Dictionary/etc.)
6.  Converting more complex structures such as DateTime, Guid, etc.

@Zvirja starts:

I analyzed the way with the `dynamic` keyword and liked it (even added to todo), however decided to proceed with a "static" direction. I defined the `IPtr` interface that provides address, type and heap and build a set of different extension methods around that. One kind of extensions allows to read fields from the object.
Below are examples of code that is used to achieve the different tasks.

Read fields:

```
IPtr ptr; //here we somehow get the object.
int intValue = ptr.GetInt("intField", /*interior true/false*/)
string strValue = ptr.GetString("stringField");
IPtr refValue = ptr.GetRef("ptrField");
IPtr structPtr = ptr.GetStruct("structField");
```

Read chains:

```
int intValue = ptr.GetSimpleChained<int>("refFld.someStruct.intFld");
IPtr ref = ptr.GetRefChained("refFld.innerRef");
```

To read collections (example of supported: `Array`, `List<T>`, `Dictionary<TKey,TValue>`, `ConcurrentDictionary<TKey, TValue>`) I created the specialized wrappers. API usage example:

```
List<int> listOfInts = ptr.AsCollection().List()/*Wrapper is returned. I can get elements, size, etc.*/.Read<int>();
List<KeyValuePair<int, IPtr>> dictionary = ptr.AsCollection().Dictionary().Read<int, IPtr>(); //For Dictionary<int, object>
```

To read complex values:

```
DateTime dateTime = ptr.GetThisComplexValue().DateTime();
Guid guid = ptr.GetThisComplexValue().GuidValue();
```

@JeffCyr comments:
I like the idea of making extensible convertions. However, I prefer an approach where you don't have to specify the type of the field you are accessing or whether it's interior or not. It can be resolved quickly by inspecting the ClrInstanceField without much performance hit.

e.g.

``` c#
ClrObject o = ...;
int value = o["fieldName"]["innerField"]["intField"].AsInt32();
```

``` c#
[...]

PR #30, opened 2015-10-30
Implement an object model for ClrMD

leculver:
This change implements an object model for ClrMD.  Additionally we now support enumerating arguments and local variables for v4 runtimes and beyond.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrObject.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrObject.cs
@@ -1,7 +1,7 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
+
+#pragma warning disable 1591
 
 namespace Microsoft.Diagnostics.Runtime
 {
@@ -48,6 +48,7 @@ namespace Microsoft.Diagnostics.Runtime
         /// </summary>
         public bool IsValid { get { return _type != _type.Heap.ErrorType; } }
 
+        #region Fields
         /// <summary>
         /// Gets an object reference field from ClrObject.  Any field which is a subclass of System.Object
         /// </summary>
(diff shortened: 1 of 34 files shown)


src/Microsoft.Diagnostics.Runtime/ClrValue.cs
```diff
@@ -0,0 +1,1077 @@
...
+
+        /// <summary>
+        /// Returns the result of ReadPointer(Address) if this value is an object.
+        /// </summary>
+        protected abstract ulong Object { get; }
+
+        /// <summary>
+        /// Returns whether Address is an interior value.
+        /// </summary>
+        public abstract bool Interior { get; }
```
noahfalk: It seems like this would be hard to know in general, unless you constrain how ClrValue objects can be created. I'd figure people would love to just point at random addresses in the target and say treat that as a ClrValue?

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
```diff
@@ -0,0 +1,1077 @@
...
+            return result;
+        }
+
+
+        /// <summary>
+        /// Converts the value to an IntPtr.  Note that this method is a conversion, not a cast.  Similarly sized data will be converted.
+        /// </summary>
+        /// <returns>This value as an IntPtr.</returns>
+        public virtual IntPtr AsIntPtr()
+        {
```
noahfalk: This makes the assumption that target pointer and host pointer have the same size. Its always true today for DAC, but I don't know if you want to encourage people to lock themselves into that? I'd encourage people to map IntPtr as a long or ulong so that things keep working if we enable debugging 64 bit processes with 32 bit debuggers

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
```diff
@@ -0,0 +1,1077 @@
...
+            ulong obj = Object;
+            return new ClrObject(obj, obj != 0 ? heap.GetObjectType(obj) : heap.NullType);
+        }
+
+        /// <summary>
+        /// Converts the value to a boolean.  Note that this method is a conversion, not a cast.  Similarly sized data will be converted.
+        /// </summary>
+        /// <returns>This value as a boolean.</returns>
+        public virtual bool AsBoolean()
+        {
```
noahfalk: C# does support overloading the implicit or explicit casting operators. Rather writing:
bool b = value.AsBoolean()

Customers could write:
bool b = (bool)value;
OR:
bool b = value;

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
```diff
@@ -0,0 +1,1077 @@
...
+        /// will not do type coercion.  This method will throw an ArgumentException if no field matches
+        /// the given name.  It will throw a NullReferenceException if the target object is null (that is,
+        /// if (IsNull returns true).  It will throw an InvalidOperationException if the field is not
+        /// of the correct type.  Lastly, it will throw a MemoryReadException if there was an error reading
+        /// the value of this field out of the data target.
+        /// </summary>
+        /// <param name="fieldName">The name of the field to get the value for.</param>
+        /// <returns>The value of the given field.</returns>
+        public bool GetBoolean(string fieldName)
+        {
```
noahfalk: You might be able to collapse many of these methods into something like T GetField<T>(string fieldname)?

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
```diff
@@ -0,0 +1,1077 @@
...
+                    _value.GetSize(out size);
+                    _size = (int)size;
+                }
+
+                return _size.Value;
+            }
+        }
+
+        public override bool Interior
+        {
```
noahfalk: I haven't looked enough at the surrounding code, but this seemed a surprising implementation

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
```diff
@@ -0,0 +1,1077 @@
+﻿using System;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    /// <summary>
+    /// A ClrValue represents the value of a field or variable.  This value may be of any type (or null).
+    /// </summary>
+    public abstract class ClrValue
```
noahfalk: You might consider making an interface IClrValue which is implemented by both ClrObject and ClrValue? Alternatively if ClrObject was a reference type you could derive from ClrValue. In my mind if I wanted a really high performance object enumerator I'd just use the raw addresses. As soon as you are willing to pay the cost to look up the method table and create a reference for it, it doesn't feel like I am saving much having ClrObject be a struct. All that said, I have 0 perf data in front of me and this is wild speculation : )

JeffCyr:
Nice work!

What is the purpose of having a separate ClrValue class? I thought everything would be implemented in ClrObject with the interior flag masked in the address field. With the current impl we would not be able to inspect struct fields without allocating ClrValues in the heap.

JeffCyr:
@noahfalk You should take a look at issue #24, there has been discussions about some of your comments.

noahfalk:
@JeffCyr Thanks Jeff! Indeed Lee pointed me at the review before I knew of that issue. Skimming through I get the impression you guys have spent a lot more time thinking about the problems than I have so don't put too much weight behind my comments. Those are just the things that struck while taking a first look.

src/Microsoft.Diagnostics.Runtime/ClrObject.cs
```diff
@@ -0,0 +1,884 @@
...
+        /// if (IsNull returns true).  It will throw an InvalidOperationException if the field is not
+        /// of the correct type.  Lastly, it will throw a MemoryReadException if there was an error reading
+        /// the value of this field out of the data target.
+        /// </summary>
+        /// <param name="fieldName">The name of the field to get the value for.</param>
+        /// <returns>The value of the given field.</returns>
+        public char GetChar(string fieldName)
+        {
+            ulong address = GetFieldAddress(fieldName, ClrElementType.Char, "char");
+            char result;
```
leculver: char result.

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
leculver: In general this came out of a realization that you always do the following every time you enumerate the heap:

    foreach (ulong address in heap.EnumerateObjectAddresses())
    {
        ClrType type = heap.GetObjectType(address);  //  <-  This is basically required code.

Having them wrapped up into one struct of two pointers in size is just more efficient both size-wise (structs are generally faster when they are 16 bytes or smaller) and in the sense that the user *has* to get the type to do anything meaningful with the object at all.

As for inheritance/interface, I agree I probably need to add an interface here, but I didn't want to add one until I have some perf data on how much it will cost.  (The answer is probably nothing, performance related...or at least so small it doesn't matter.)  I plan to look into that this week if I have time.

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
leculver: Interior is tough to keep track of, but users of ClrMD already have to keep track of it to use the API now...  Also I'm ok with people pointing at random addresses to threat as a ClrValue...as at that point they likely know what they are doing.  They'd have to implement their own subclass of ClrValue though.

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
leculver: Thanks, that's good feedback.  I'm pretty sure the cast operator would want a stronger check (AsBoolean will convert Int8/UInt8 as well).  I'll probably add cast operators in the future.

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
leculver: Good point.

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
leculver: I thought of that, but didn't for two reasons:

1.  Intellisense.  Feedback was that it would be better to have named fields.
2.  C# doesn't have template specialization, so it's tough to get an efficient implementation of this.  Not impossible though, and perf it's exactly the largest concern.  So this was secondary.

I'm still a bit on the fence about this honestly, especially after seeing how similar the implementations turned out to be.

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
leculver: It's not correct.  Good catch.  =)

src/Microsoft.Diagnostics.Runtime/ClrValue.cs
leculver: Ah so the problem is generally the last step.  Let's say you already have the raw address of where to read from and you know the type T of the variable you are trying to convert to.  There's not a great way to way to convert a byte[] in a generic Foo<T> to a T.

Like let's say you have a byte[] of the value already read, and you have to convert it to T.  The general way to convert a byte[] to (let's say) a boolean is to call BitConverter.ToBoolean().  As far as I can tell you will have to either convert that boolean to an object and convert it on the way out, or do something funny with the dynamic keyword.

Either of which aren't the end of the world, just slower.  I assume that BitConverter is the class I should model behavior after, and thus we have "GetBoolean, GetInt16, etc).

leculver:
@JeffCyr 

> With the current impl we would not be able to inspect struct fields without allocating ClrValues in the heap.

In the current implementation that's correct.  After looking everything over I've basically made this tradeoff:

1.  ClrObject is a fast, lightweight design which will allow you to enumerate the heap without creating garbage.  It only works for reference types.
2.  ClrValue can represent *any* type of value in the target runtime, including objects, structs, and primitives.  It's implemented as a class since they require more bookkeeping (depending on how they are constructed) and need to be sub-classed so I can stitch together things like ICorDebug (which is how we get local variables) and fields (pure ClrMD implementation).

If you are dereferencing fields, you will can go through ClrObject for object fields (so no garbage there), but yes one tradeoff here is that you do have to create an object when passing through a struct field.  I think these will be the lesser case than dealing with objects, and hopefully will be cleaned in a gen0 GC if you aren't hanging onto them.

You could imagine we could also implement a "struct ClrStruct" which was the original design we discussed (address/type/interior, lighter weight), but it's not something I'm going to implement in the first iteration of this feature.

PR #40, opened 2016-02-17
Serialized Exception support for .NET native

cshung:
This change added the support for reading the serialized exceptions in .NET native dumps.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
@@ -19,13 +19,18 @@ namespace Microsoft.Diagnostics.Runtime
     public abstract class ClrHeap
     {
         /// <summary>
+        /// Returns a ClrType from an EE address. Available in .NET native only. Used for constructing ClrType object from an EEType value encoded in a .NET native serialized exception.
+        /// </summary>
+        public abstract ClrType GetClrTypeFromEE(ulong eeType);
+
+        /// <summary>
         /// And the ability to take an address of an object and fetch its type (The type alows further exploration)
         /// </summary>
         abstract public ClrType GetObjectType(Address objRef);
 
         /// <summary>
         /// Returns whether this version of CLR has component MethodTables.  Component MethodTables were removed from
-        /// desktop CLR in v4.6, and do not exist at all on .Net Native.  If this method returns false, all componet
+        /// desktop CLR in v4.6, and do not exist at all on .Net Native.  If this method returns false, all component
         /// MethodTables will be 0, and expected to be 0 when an argument to a function.
         /// </summary>
         virtual public bool HasComponentMethodTables { get { return true; } }
--- a/src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
@@ -19,11 +19,6 @@ namespace Microsoft.Diagnostics.Runtime
     public abstract class ClrHeap
     {
         /// <summary>
-        /// Returns a ClrType from an EE address. Available in .NET native only. Used for constructing ClrType object from an EEType value encoded in a .NET native serialized exception.
-        /// </summary>
-        public abstract ClrType GetClrTypeFromEE(ulong eeType);
-
-        /// <summary>
         /// And the ability to take an address of an object and fetch its type (The type alows further exploration)
         /// </summary>
         abstract public ClrType GetObjectType(Address objRef);
--- a/src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Diagnostics.Runtime
         /// <summary>
         /// In .NET native crash dumps, we have a list of serialized exceptions objects. This property expose them as ClrException objects.
         /// </summary>
-        public abstract IEnumerable<ClrException> GetSerializedExceptions();
+        public abstract IEnumerable<ClrException> EnumerateSerializedExceptions();
 
         /// <summary>
         /// The ClrInfo of the current runtime.
--- a/src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
@@ -21,6 +21,11 @@ namespace Microsoft.Diagnostics.Runtime
     public abstract class ClrRuntime
     {
         /// <summary>
+        /// In .NET native crash dumps, we have a list of serialized exceptions objects. This property expose them as ClrException objects.
+        /// </summary>
+        public abstract IEnumerable<ClrException> GetSerializedExceptions();
+
+        /// <summary>
         /// The ClrInfo of the current runtime.
         /// </summary>
         public ClrInfo ClrInfo { get; protected set; }
--- a/src/Microsoft.Diagnostics.Runtime/Desktop/heap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/Desktop/heap.cs
@@ -1667,6 +1667,10 @@ namespace Microsoft.Diagnostics.Runtime.Desktop
             return type;
         }
 
+        public override ClrType GetClrTypeFromEE(ulong eeType)
+        {
+            throw new NotSupportedException("GetClrTypeFromEE is only supported for .NET native.");
+        }
     }
 
     internal class V46GCHeap : DesktopGCHeap
@@ -1799,6 +1803,11 @@ namespace Microsoft.Diagnostics.Runtime.Desktop
                 ret.ComponentType = TryGetComponentType(obj, 0);
 
             return ret;
+        }
+
+        public override ClrType GetClrTypeFromEE(ulong eeType)
+        {
+            throw new NotSupportedException("GetClrTypeFromEE is only supported for .NET native.");
         }
     }
 }
--- a/src/Microsoft.Diagnostics.Runtime/Desktop/heap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/Desktop/heap.cs
@@ -1666,11 +1666,6 @@ namespace Microsoft.Diagnostics.Runtime.Desktop
 
             return type;
         }
-
-        public override ClrType GetClrTypeFromEE(ulong eeType)
-        {
-            throw new NotSupportedException("GetClrTypeFromEE is only supported for .NET native.");
-        }
     }
 
     internal class V46GCHeap : DesktopGCHeap
@@ -1803,11 +1798,6 @@ namespace Microsoft.Diagnostics.Runtime.Desktop
                 ret.ComponentType = TryGetComponentType(obj, 0);
 
             return ret;
-        }
-
-        public override ClrType GetClrTypeFromEE(ulong eeType)
-        {
-            throw new NotSupportedException("GetClrTypeFromEE is only supported for .NET native.");
         }
     }
 }
--- a/src/Microsoft.Diagnostics.Runtime/Desktop/legacyruntime.cs
+++ b/src/Microsoft.Diagnostics.Runtime/Desktop/legacyruntime.cs
@@ -55,6 +55,11 @@ namespace Microsoft.Diagnostics.Runtime.Desktop
             }
         }
 
+        public override IEnumerable<ClrException> GetSerializedExceptions()
+        {
+            throw new NotSupportedException("GetSerializedExceptions is only supported for .NET native.");
+        }
+
         internal override ulong[] GetAssemblyList(ulong appDomain, int count)
         {
             return RequestAddrList(DacRequests.ASSEMBLY_LIST, appDomain, count);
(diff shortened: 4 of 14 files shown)


src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
```diff
@@ -19,13 +19,18 @@ namespace Microsoft.Diagnostics.Runtime
     public abstract class ClrHeap
     {
         /// <summary>
+        /// Returns a ClrType from an EE address. Available in .NET native only. Used for constructing ClrType object from an EEType value encoded in a .NET native serialized exception.
+        /// </summary>
+        public abstract ClrType GetClrTypeFromEE(ulong eeType);
```
leculver: You should not create a function for this.  Instead override GetTypeByMethodTable and provide your implementation there.

src/Microsoft.Diagnostics.Runtime/Desktop/heap.cs
```diff
@@ -1667,6 +1667,10 @@ public override ClrType GetObjectType(Address objRef)
             return type;
```
leculver: Changes in this file should be reverted.  You should implement GetTypeByMethodTable for .Net Native instead.

src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
```diff
@@ -21,6 +21,11 @@ namespace Microsoft.Diagnostics.Runtime
     public abstract class ClrRuntime
     {
         /// <summary>
+        /// In .NET native crash dumps, we have a list of serialized exceptions objects. This property expose them as ClrException objects.
+        /// </summary>
+        public abstract IEnumerable<ClrException> GetSerializedExceptions();
```
leculver: Can you rename this to "EnumerateSerializedExceptions"?

src/Microsoft.Diagnostics.Runtime/Desktop/legacyruntime.cs
```diff
@@ -55,6 +55,11 @@ internal override DesktopVersion CLRVersion
             }
         }
 
+        public override IEnumerable<ClrException> GetSerializedExceptions()
+        {
+            throw new NotSupportedException("GetSerializedExceptions is only supported for .NET native.");
```
leculver: This should return an empty collection (`return new ClrException[0];`).

Generally speaking, the design of ClrMD is such that the consumer of the API expects empty collections when something is not supported on that platform, and acts accordingly...instead of wrapping try/catches or checking what version of the runtime is loaded.

src/Microsoft.Diagnostics.Runtime/Desktop/v45runtime.cs
```diff
@@ -550,6 +550,11 @@ internal override IRCWData GetRCWData(ulong rcw)
         }
         #endregion
 
+        public override IEnumerable<ClrException> GetSerializedExceptions()
```
leculver: See the comment in LegacyRuntime.cs.  This should be overridden in DesktopBaseRuntime instead.

src/Microsoft.Diagnostics.Runtime/Native/NativeHeap.cs
```diff
@@ -23,6 +23,76 @@ internal class NativeHeap : HeapBase
...
+                _ip = ip;
+                _symbolName = symbolName;
+                _module = module;
+            }
+
+            public override string DisplayString
+            {
+                get
+                {
+                    return $"{_symbolName}";
```
leculver: This should just be `return _symbolName;`.

src/Microsoft.Diagnostics.Runtime/Native/NativeHeap.cs
```diff
@@ -23,6 +23,76 @@ internal class NativeHeap : HeapBase
...
+                {
+                    return _ip;
+                }
+            }
+
+            public override ClrStackFrameType Kind
+            {
+                get
+                {
+                    throw new NotImplementedException("This information is unavailable when debugging Native code");
```
leculver: As mentioned above, the design of ClrMD is to not throw exceptions except in unrecoverable cases.  Can you add a new entry to ClrStackFrameType called Unknown, the `return ClrStackFrameType.Unknown;` here?

    public enum ClrStackFrameType
    {
        /// <summary>
        /// Indicates that information on this stack frame type is unavailable
        /// </summary>
        Unknown = -1,

        /// <summary>
        /// Indicates this stack frame is a standard managed method.
        /// </summary>
        ManagedMethod = 0,

        /// <summary>
        /// Indicates this stack frame is a special stack marker that the Clr runtime leaves on the stack.
        /// Note that the ClrStackFrame may still have a ClrMethod associated with the marker.
        /// </summary>
        Runtime = 1
    }

src/Microsoft.Diagnostics.Runtime/Native/NativeHeap.cs
```diff
@@ -23,6 +23,76 @@ internal class NativeHeap : HeapBase
...
+                }
+            }
+
+            public override ClrMethod Method
+            {
+                get
+                {
+                    throw new NotImplementedException("This information is unavailable when debugging Native code");
```
leculver: `return null;` instead

src/Microsoft.Diagnostics.Runtime/Native/NativeHeap.cs
```diff
@@ -23,6 +23,76 @@ internal class NativeHeap : HeapBase
...
+                {
+                    throw new NotImplementedException("The SP is not serialized, hence it is unavailable when debugging Native code");
+                }
+            }
+
+            public override ClrThread Thread
+            {
+                get
+                {
+                    throw new NotImplementedException("The Thread is not serialized, hence it is unavailable when debugging Native code");
```
leculver: `return null;`

src/Microsoft.Diagnostics.Runtime/Native/NativeHeap.cs
```diff
@@ -23,6 +23,76 @@ internal class NativeHeap : HeapBase
...
+                {
+                    throw new NotImplementedException("This information is unavailable when debugging Native code");
+                }
+            }
+
+            public override ulong StackPointer
+            {
+                get
+                {
+                    throw new NotImplementedException("The SP is not serialized, hence it is unavailable when debugging Native code");
```
leculver: `return 0;` instead

src/Microsoft.Diagnostics.Runtime/Native/NativeHeap.cs
```diff
@@ -84,6 +154,121 @@ private void CreateFreeType()
...
+            public override ClrType Type
+            {
+                get { return _type; }
+            }
+
+            public override string Message
+            {
+                get
+                {
+                    return "<Invalid Object>"; // we do not expect a Message in serialized exceptions, however, returning <Invalid Object> as the SOS does
```
leculver: Better to `return null;` here.  This matches the rest of what ClrMD does.  It tells the user we do not have this message, and they can decide what to show ("<Invalid Object>" or something of their own choosing).  If we return a special string here, consumers of the API will have to handle this case.

src/Microsoft.Diagnostics.Runtime/Native/NativeHeap.cs
```diff
@@ -84,6 +154,121 @@ private void CreateFreeType()
...
+            int index;
+            if (_indices.TryGetValue(eeType, out index))
+                clrType = _types[index];
+            else
+                clrType = ConstructObjectType(eeType);
+
+            return clrType;
+        }
+
+        internal class NativeException : ClrException
```
leculver: Can you move this class to its own file?  I realize the desktop one doesn't do that, but it'd be better...

src/Microsoft.Diagnostics.Runtime/Native/NativeRuntime.cs
```diff
@@ -46,12 +50,22 @@ protected override void InitApi()
 
                 _sos = (ISOSNative)dac;
             }
+            if (_sosNativeSerializedExceptionSupport == null)
+            {
+                var dac = _library.DacInterface;
```
leculver: You can replace all of this with: `_sosNativeSerializedExceptionSupport = _library.DacInterface as ISOSNativeSerializedExceptionSupport;`.

src/Microsoft.Diagnostics.Runtime/Native/NativeRuntime.cs
```diff
@@ -261,6 +275,95 @@ private void InitThreads()
             _threads = threads.ToArray();
         }
 
+        public override IEnumerable<ClrException> GetSerializedExceptions()
+        {
+            if (!this.ClrInfo.DacInfo.PlatformAgnosticFileName.Contains("mrt100"))
+                throw new ClrDiagnosticsException("Serialized exceptions are only available for a Native data target");
+
+            if (_sosNativeSerializedExceptionSupport == null)
+                throw new ClrDiagnosticsException("This version of mrt100 is too old.", ClrDiagnosticsException.HR.DataRequestError);
```
leculver: As mentioned above, this should be `return new ClrException[0];`.  In cases where data is not available we return an empty list instead of throwing.

src/Microsoft.Diagnostics.Runtime/Native/NativeRuntime.cs
```diff
@@ -261,6 +275,95 @@ private void InitThreads()
             _threads = threads.ToArray();
         }
 
+        public override IEnumerable<ClrException> GetSerializedExceptions()
+        {
+            if (!this.ClrInfo.DacInfo.PlatformAgnosticFileName.Contains("mrt100"))
```
leculver: You should remove this if statement.  The only way to construct this class (NativeRuntime) is if we are debugging a .Net Native process.  This is also the wrong way to check for .Net Native...but since this check is completely not needed, you should just remove this if statement.

src/Microsoft.Diagnostics.Runtime/Native/NativeRuntime.cs
```diff
@@ -261,6 +275,95 @@ private void InitThreads()
...
+                {
+                    ISerializedStackFrame serializedStackFrame = serializedStackFrameEnumerator.Next();
+
+                    NativeModule nm = ((NativeHeap)this.GetHeap()).GetModuleFromAddress(serializedStackFrame.IP);
+                    if (nm == null)
+                    {
+                        Trace.WriteLine(String.Format("Unable to resolve module for IP {0}", serializedStackFrame.IP));
+                        continue;
+                    }
+                    if (nm.Pdb == null) {
```
leculver: The `{` needs to be on its own line to match the rest of clrmd.  =)

src/Microsoft.Diagnostics.Runtime/Native/NativeRuntime.cs
```diff
@@ -261,6 +275,95 @@ private void InitThreads()
...
+                    try
+                    {
+                        resolver = this.DataTarget.SymbolProvider
+                                        .GetSymbolResolver(nm.Pdb.FileName, nm.Pdb.Guid, nm.Pdb.Revision);
+                    }
+                    catch (Exception e)
+                    {
+                        Trace.WriteLine($"Error in finding the symbol resolver for PDB [Filename:{nm.Pdb.FileName}, GUID:{nm.Pdb.Guid}, Revision:{nm.Pdb.Revision}]: {e.Message}");
+                        Trace.WriteLine("Check previous traces for additional information");
+                        continue;
```
leculver: Continuing is incorrect.  That will lead to a stack trace with missing frames when we can't resolve the name.  Instead `string symbolName` should simply be null when we cannot resolve the name.  The stack trace will contain the IP (but not its resolved name) and the consumer of the API can try again to resolve that IP themselves.

src/Microsoft.Diagnostics.Runtime/Native/NativeRuntime.cs
```diff
@@ -261,6 +275,95 @@ private void InitThreads()
...
+                    }
+                    if (resolver == null)
+                    {
+                        Trace.WriteLine($"Unable to find symbol resolver for PDB [Filename:{nm.Pdb.FileName}, GUID:{nm.Pdb.Guid}, Revision:{nm.Pdb.Revision}]");
+                        continue;
+                    }
+                    string symbolName = resolver.GetSymbolNameByRVA((uint)(serializedStackFrame.IP - nm.ImageBase));
+
+                    NativeHeap.NativeStackFrame nativeStackFrame = new NativeHeap.NativeStackFrame(serializedStackFrame.IP, symbolName, nm);
+                    if (nativeStackFrame != null)
```
leculver: nativeStackFrame can never be null...

src/Microsoft.Diagnostics.Runtime/Native/NativeRuntime.cs
```diff
@@ -261,6 +275,95 @@ private void InitThreads()
...
+                ISerializedStackFrameEnumerator serializedStackFrameEnumerator = serializedException.StackFrames;
+                while (serializedStackFrameEnumerator.HasNext())
+                {
+                    ISerializedStackFrame serializedStackFrame = serializedStackFrameEnumerator.Next();
+
+                    NativeModule nm = ((NativeHeap)this.GetHeap()).GetModuleFromAddress(serializedStackFrame.IP);
+                    if (nm == null)
+                    {
+                        Trace.WriteLine(String.Format("Unable to resolve module for IP {0}", serializedStackFrame.IP));
+                        continue;
```
leculver: We should not continue when we encounter a null module.  This leads to an incorrect stack trace.  Not having the module simply means we can't resolve the symbol name...we should still produce a stack frame to our user...even if the only useful thing in the ClrStackFrame is the ip.

cshung:
My fork is updated and should have addressed all the concerns.

PR #41, opened 2016-03-03
Fixing a couple bugs in the serialized exception implementation

cshung:
Hit these bugs when I tried to integrate the serialized exception into PerfView.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
@@ -831,12 +831,12 @@ namespace Microsoft.Diagnostics.Runtime
             _dacInterface.Flush();
 
             IGCInfo data = GetGCInfo();
-            if (data == null)
-                throw new ClrDiagnosticsException("This runtime is not initialized and contains no data.", ClrDiagnosticsException.HR.RuntimeUninitialized);
-
-            ServerGC = data.ServerMode;
-            HeapCount = data.HeapCount;
-            CanWalkHeap = data.GCStructuresValid && !dataTarget.DataReader.IsMinidump;
+            if (data != null)
+            {
+                ServerGC = data.ServerMode;
+                HeapCount = data.HeapCount;
+                CanWalkHeap = data.GCStructuresValid && !dataTarget.DataReader.IsMinidump;
+            }
             _dataReader = dataTarget.DataReader;
         }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
```diff
@@ -831,12 +831,12 @@ public RuntimeBase(ClrInfo info, DataTargetImpl dataTarget, DacLibrary lib)
             _dacInterface.Flush();
 
             IGCInfo data = GetGCInfo();
```
leculver: I don't think this is the right fix.  This check (GetGCInfo returning null) for desktop CLR is checking for a bad crash dump.  Basically we always expect that we can get the GCInfo out of a crash dump, otherwise that means the crash dump was generated incorrectly.  So this is a real and valid exception in that case.  From you changing it, I assume that means that this returns null in .Net Native sometimes?

Instead of this change, I would first suggest debugging the dac when GetGCInfo returns null, and see why that happens and if it's expected.  If you are 100% sure it's ok that .Net Native dac can return null here and that's not a bug, then removing "if (data == null)" should only be done by 'NativeRuntime'.

(And if you do that, I would suggest pulling the code dealing with IGCInfo here into a virtual function called 'InitGCInfo' that NativeRuntime overrides.)

src/Microsoft.Diagnostics.Runtime/Native/NativeTypes.cs
```diff
@@ -26,9 +26,12 @@ public NativeType(NativeHeap heap, int index, NativeModule module, string name,
             _eeType = eeType;
             _index = index;
```
leculver: Why are these changes necessary?  ConstructObjectType checks that mtData is not null before calling here.

The only place that could generate a null mtData for this function is CreateFreeType:

            IMethodTableData mtData = NativeRuntime.GetMethodTableData(free);
            _free = new NativeType(this, _types.Count, _mrtModule, "Free", free, mtData);

However, if getting method table data from the free EEType is returning null...then again we have a bad crash dump.  If you are sure we want to handle cases where this fails (I assume for micro crash dumps?), then I think this is ok.

cshung:
Fixed - the check for bad crash dump is now preserved for desktop runtimes.

In .NET native minidumps, the only thing that we preserve in the WER buffer. That is why we cannot get to GCData or MethodTableData.

Issue #46, opened 2016-08-14
x64 stack walking is broken


goldshtn:
The NuGet version works fine, but with `master` I'm getting UNKNOWN frames for any managed method in a 64-bit process. After running `git bisect` I was able to pinpoint https://github.com/Microsoft/clrmd/commit/3a6a4303c54092ad2e9efee6fad5d68e33663b6e as the problematic commit. I assume it's part of the move to the `CorDebug` interfaces.

weltkante:
It's suspicious that as part of that commit it only does `BitConverter.ToUInt32` on what looks to be a pointer. Might be truncating 64 bit adresses. Old code used `BitConverter.ToUInt64` at least in one of the places.

mitikov:
Switching to `BitConverter.ToUInt64` did the trick for me 👍 

Thanks guys.

PR #48, opened 2016-08-31
Bug fix for "#3a6a4303 2015-10-21 AM 5:36:02 GerICorDebugILFrame for …

stjeong:
There is a bug in #3a6a4303 commit.

That code is not work for x64 environment. (Call stack's method is "UNKNOWN")

The issue "x64 stack walking is broken #46" is relevant to this.


leculver:
The stack walking issue is fixed, and I will look into the PEBuffer memory leaks.  I can't accept these changes back into the master branch though.  We are supporting the most recent C# compiler, so I can't accept the PR as a whole.

PR #74, opened 2017-03-28
Add links to MemoScope.NET and DumpMiner

KirillOsenkov:


KirillOsenkov:
Ping @leculver :)

leculver:
Sorry, I don't want to add these links here.  I don't want to be on the hook for keeping a list of projects using ClrMD.

Issue #66, opened 2017-03-13
[Bug] Incomplete signature for some generic methods


Fread75:
I've made a simple test by adding the method :

`public T3 GenericBar<T1, T2, T3>(Func<T1, T2, T3> a) { return a(default(T1), default(T2)); }`

in the unit tests, in the SharedLibrary file, class `Foo`.

A call to the `GetFullSignature` method on the object `DesktopMethod` returns an incomplete signature. It seems to be the case when several generic parameters are implicated, but it is not systematic.

The error seems to come from `v45runtime.GetNameForMD(ulong md)`. In this case, the first call to `_sos.GetMethodDescName` set an incorrect `pNeeded` value. In this case, a second call to `_sos.GetMethodDescName` just to have the correct value solves the issue.

Fread75:
Hello,

I am currently using ClrMd for a professional project, and my company doesn't allow us *yet* to contribute to open source projects (mostly for legal reasons). They are working on it and in a few months, I would be authorized to submit a pull request to your repo.

In the mean time, I am working on a local clone of your repository. And I create here all the tickets that correspond to the commit I have made on my local repo, mostly to have your feedback in those points, and also in case that you want to correct them directly.

Regards,
Fred

PR #75, opened 2017-04-03
Patch for a bug on SOS side when retrieving a method's signature

Fread75:
#66 Patch for a bug on SOS side when retrieving a method's signature

Sometimes, when the target method has parameters with generic types the first call to GetMethodDescName sets an incorrect value into pNeeded.

In those cases, a second call directly after the first returns the correct value.

+ unit test


EricDales:
Hello, 

can you please provide some feedback on this request ? 

Thank you !
Eric

PR #80, opened 2017-06-23
Enable official builds, signing, packaging

chcosta:
Changes to enable official builds.

The most consistent way we have to provision dependencies to build machines, is via build tools.  I noticed that this repo was using some C# 7 features, but the project system is based on the older `packages.config`.

To address using a later Roslyn compiler, and ensure package references are available, I moved this repo to use BuildTools (which is using a newer Roslyn version than what is available in the build machine VS installs.  Additionally, I updated the two test projects which were using packages.config to use project.json.  I debated just moving all of the projects to build on the latest CLI (2.0), but it isn't really necessary.

build.cmd now runs init-tools.cmd (install buildtools), restores packages, and builds.  It uses the CLI for some of this (like package restores), but pass through to msbuild for building so that we can use project.json instead of packagereferences (you should consider moving to packagereferences via `dotnet migrate`.  

I've added a `pkg` folder which has the nuspec I created to match very closely to the package I downloaded from nuget.org.  

Just to really make my point, of all the changes in this PR, the one's I'd expect you to want to pay the most attention to are the changes to the csproj's / build.cmd.


chcosta:
@leculver, it just occurred to me that I didn't include the xml file in the nuspec. I can add this on Monday. Sorry about that

leculver:
No problem.

PR #91, opened 2017-12-07
Implement create process snapshot and attach

AviAvni:
This pr is create at SELA hackathon with Sasha Goldshtein @goldshtn

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/datatarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/datatarget.cs
@@ -881,7 +881,7 @@ namespace Microsoft.Diagnostics.Runtime
             IDataReader reader;
             if (attachFlag == AttachFlag.Passive)
             {
-                reader = new LiveDataReader(pid);
+                reader = new LiveDataReader(pid, createSnapshot: false);
             }
             else
             {
@@ -901,7 +901,7 @@ namespace Microsoft.Diagnostics.Runtime
         /// <returns>A DataTarget instance.</returns>
         public static DataTarget CreateSnapshotAndAttach(int pid)
         {
-            IDataReader reader = new LiveDataReader(pid, true);
+            IDataReader reader = new LiveDataReader(pid, createSnapshot: true);
             DataTargetImpl dataTarget = new DataTargetImpl(reader, null);
             return dataTarget;
         }
@@ -2060,32 +2060,32 @@ namespace Microsoft.Diagnostics.Runtime
         #region Variables
         private int _originalPid;
         private IntPtr _snapshotHandle;
-        private IntPtr _vaCloneHandle;
+        private IntPtr _cloneHandle;
         private IntPtr _process;
         private int _pid;
         #endregion
 
         private const int PROCESS_VM_READ = 0x10;
         private const int PROCESS_QUERY_INFORMATION = 0x0400;
-        public LiveDataReader(int pid, bool createSnapshot = false)
+        public LiveDataReader(int pid, bool createSnapshot)
         {
             if (createSnapshot)
             {
                 _originalPid = pid;
-                var process = Process.GetProcessById(pid);
-                var hr = PssCaptureSnapshot(process.Handle, PSS_CAPTURE_FLAGS.PSS_CAPTURE_VA_CLONE, IntPtr.Size == 8 ? 0x0010001F : 0x0001003F, out _snapshotHandle);
+                Process process = Process.GetProcessById(pid);
+                int hr = PssCaptureSnapshot(process.Handle, PSS_CAPTURE_FLAGS.PSS_CAPTURE_VA_CLONE, IntPtr.Size == 8 ? 0x0010001F : 0x0001003F, out _snapshotHandle);
                 if(hr != 0)
                 {
                     throw new ClrDiagnosticsException(String.Format("Could not create snapshot to process. Error {0}.", hr));
                 }
 
-                hr = PssQuerySnapshot(_snapshotHandle, PSS_QUERY_INFORMATION_CLASS.PSS_QUERY_VA_CLONE_INFORMATION, out _vaCloneHandle, IntPtr.Size);
+                hr = PssQuerySnapshot(_snapshotHandle, PSS_QUERY_INFORMATION_CLASS.PSS_QUERY_VA_CLONE_INFORMATION, out _cloneHandle, IntPtr.Size);
                 if (hr != 0)
                 {
                     throw new ClrDiagnosticsException(String.Format("Could not query the snapshot. Error {0}.", hr));
                 }
 
-                _pid = GetProcessId(_vaCloneHandle);
+                _pid = GetProcessId(_cloneHandle);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


goldshtn:
To clarify this PR a bit: the [Windows Process Snapshotting API](https://msdn.microsoft.com/en-us/library/dn457825(v=vs.85).aspx) provides a facility for creating a virtual address space clone of the target process. By creating a clone and then attaching the CLRMD live data reader to the clone we benefit from the best of both worlds:

- The target process is not suspended at all (just like `AttachMode.Passive`), nor do we generate a dump file
- We get a consistent snapshot of the target process (like `AttachMode.NonInvasive` or a dump file)

Creating the clone is a relatively cheap operation, however the cloned pages are copy-on-write. If the original process modifies a page, it is duplicated in RAM for the clone.

One final caveat is that Process Snapshotting is supported on Windows 8.1/Windows Server 2012 R2 and later.

goldshtn:
Hi @leculver, I was wondering if we could get some feedback on this PR, or ideally get it merged. Thanks!

leculver:
Overall this looks good.  I'd like to double check a few things but I plan to merge this later in the week.

goldshtn:
@Alois-xx I don't think it's a good idea to change the default behavior of `AttachToProcess` to use a snapshot. Attaching in passive mode consumes virtually no resources on the target -- there is no clone, there is no extra memory usage. OTOH creating a process snapshot has the potential for using more resources. Also, using a snapshot would not support the Windows debugging interfaces, unlike `AttachFlag.Invasive`. All in all, I think process snapshotting warrants a separate mode -- it would be confusing otherwise.

Alois-xx:
@goldshtn: When I call ClrMD from a command line tool like MemAnalyzer I would always need to think if I can use PSS or not and pass a different command line flag to it. Procdump follows this route but what are the reasons that I would not want to use the faster approach always? 
I have tried PSS years ago on Server 2012 machines which resulted in dumps which were not usable at all.  Since I want to reliably get full memory dumps I have never tried the PSS feature again. Is this now working reliably? 

Yes it would be confusing if not supported functionality will not work on a newer OS because a PSS is now used. But that should be a decision of the caller of the API if he wants to use an AttachFlag  like InvasiveOrSnapshot if he can live without dbgeng support.

goldshtn:
@Alois-xx Sorry, I disagree. It is already a decision of the API client to use the snapshot API. I don't think we should be making this choice automatically because of its potential cost and complications. And yes, PSS is more stable but the whole thing is not as thoroughly tested as CLRMD's main features, which have been around for years.

src/Microsoft.Diagnostics.Runtime/datatarget.cs
```diff
@@ -2045,16 +2058,40 @@ public unsafe bool GetThreadContext(uint threadID, uint contextFlags, uint conte
...
+        private IntPtr _snapshotHandle;
+        private IntPtr _vaCloneHandle;
         private IntPtr _process;
         private int _pid;
         #endregion
 
         private const int PROCESS_VM_READ = 0x10;
         private const int PROCESS_QUERY_INFORMATION = 0x0400;
-        public LiveDataReader(int pid)
+        public LiveDataReader(int pid, bool createSnapshot = false)
```
leculver: Can you make this a non-default parameter?  I'd rather it be clear there's a choice here at the callsite.

src/Microsoft.Diagnostics.Runtime/datatarget.cs
```diff
@@ -893,6 +894,18 @@ public static DataTarget AttachToProcess(int pid, uint msecTimeout, AttachFlag a
...
         }
 
+        /// <summary>
+        /// Attaches to a snapshot process (see https://msdn.microsoft.com/en-us/library/dn457825(v=vs.85).aspx).
+        /// </summary>
+        /// <param name="pid">The process ID of the process to attach to.</param>
+        /// <returns>A DataTarget instance.</returns>
+        public static DataTarget CreateSnapshotAndAttach(int pid)
+        {
+            IDataReader reader = new LiveDataReader(pid, true);
```
leculver: Can you make this:

    IDataReader reader = new LiveDataReader(pid, createSnapshot:true);

src/Microsoft.Diagnostics.Runtime/datatarget.cs
```diff
@@ -2045,16 +2058,40 @@ public unsafe bool GetThreadContext(uint threadID, uint contextFlags, uint conte
...
         private const int PROCESS_QUERY_INFORMATION = 0x0400;
-        public LiveDataReader(int pid)
+        public LiveDataReader(int pid, bool createSnapshot = false)
         {
-            _pid = pid;
-            _process = OpenProcess(PROCESS_VM_READ | PROCESS_QUERY_INFORMATION, false, pid);
+            if (createSnapshot)
+            {
+                _originalPid = pid;
+                var process = Process.GetProcessById(pid);
```
leculver: Can you change these 'var' statements to use the actual types?  (Throughout this change.)  It's not completely consistent throughout the code, but the only time var should be used is generally in statements where it reduces a ton of redundancy in a new statement.  Having 'Process process', 'int hr' makes it clear what we are working with.

src/Microsoft.Diagnostics.Runtime/datatarget.cs
```diff
@@ -2045,16 +2058,40 @@ public unsafe bool GetThreadContext(uint threadID, uint contextFlags, uint conte
     internal unsafe class LiveDataReader : IDataReader
     {
         #region Variables
+        private int _originalPid;
+        private IntPtr _snapshotHandle;
+        private IntPtr _vaCloneHandle;
```
leculver: Can this be renamed "_cloneHandle" for consistency?  ClrMD doesn't really use Hungarian notation for variable names.

leculver:
@AviAvni @goldshtn Really sorry about the delay.  I was out sick the second half of last week and wasn't able to get around to this.  Thanks for the submission.  I have a few modification requests to match the overall style of the code in ClrMD, but functionally it looks good.

@Alois-xx I do not think this should be the default for AttachToProcess with a fallback.  The user of the API should make a clear choice between using snapshots or attaching to a live process.  Remember, ClrMD supports a bunch of odd scenarios.  The LiveDataTarget can be used in places where a 'real' debugger is already attached and has the process paused (such as WinDbg/VS already attached to a process and you want to do extra inspection).  Trying to clone the process is just a waste in those scenarios.

AviAvni:
@leculver all fixed the style issues thanks for the review

leculver:
Merged.  Unfortunately the internal build sever where nuget packages are produced seems to be down, and most of CLR is out until Jan 2.  I'm going to try to put together a nuget release as soon as they are back and can take a look at why ClrMD builds aren't getting produced.  (Also hopefully adding strong name signing.)

AviAvni:
@leculver did you publish it to nuget?

leculver:
@AviAvni I pushed a new version to NuGet last week.  Unfortunately policies changed here and I had to go get re-approved for signing/shipping changes.  Sorry for the delay.

This build should contain the change:  https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/0.9.180305.1

PR #97, opened 2018-03-26
Make sure handles are returned only once

kevingosse:
When dealing with AsyncPinned handles with a state, the v45runtime.EnumerateHandleWorker method returned the interior handle twice


src/Microsoft.Diagnostics.Runtime.Tests/Targets/GCHandles.cs
```diff
@@ -13,11 +14,22 @@ public static void Main()
         string weak = "weak";
         GCHandle.Alloc(weak, GCHandleType.Weak);
 
+        //var a = new System.Net.Sockets.SocketAsyncEventArgs();
```
leculver: This shouldn't make it to the final change.

kevingosse: My bad, went too fast. Fixed.

PR #104, opened 2018-05-26
 Fix docs and comments referencing old type name. GcHeapType to ClrType

felipepessoto:


leculver:
Thank you for the change, and I'm sorry it took so long to get to this.

PR #112, opened 2018-07-05
Update WalkingTheHeap tutorial to latest API shape

karelz:
Remaining questions:
* The text about `Committed` should be likely updated.
* Not sure if `ClrHeap.EnumerateObjectAddresses` should be rather replaced with example using `EnumerateObject`
* Examples of `GCHeapType.Enumerate` should be likely replaced - not quite sure with what.


karelz:
@leculver thanks for merging - any thoughts on the remaining questions above?

Issue #108, opened 2018-06-27
Bug when DesktopInstanceField.HasSimpleValue is called on a Int32[] field


chrisnas:
While writing the ClrMD-based code to list the elements of a **ConcurrentQueue<int>**, it seems that **HasSimpleValue** is triggering a buggy state.
`    var t = heap.GetObjectType(0x00000145071476f0);  // --> System.Collections.Concurrent.ConcurrentQueue+Segment<System.Int32>
`
this segment has a int[] field called **m_array** --> we need to figure out if it is an array of simple values:
```
   var f = t.GetFieldByName("m_array");
   var r = f.HasSimpleValue;
```
When **f.HasSimpleValue** is called, the underlying type gets created and **GetTypeByMethodTable** is called with 0/null as its **obj** parameter: a new **DesktopHeapType** instance is added into **_types** (+ its **mt** into **_indices**) with a **ComponentType** set to **null**.
 	

[...]

PR #121, opened 2018-08-02
Provide optimized heap method to compute class histogram

chrisnas:
After profiling on large heaps, the expensive calls to GetObjectType and GetSize are responsible for most of the work. This code change avoids duplicated calls to these expensive methods.
--> ~30% gain on large heaps during tests on 16+ GB memory dumps

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
@@ -214,6 +214,12 @@ namespace Microsoft.Diagnostics.Runtime
         abstract public IEnumerable<ulong> EnumerateObjectAddresses();
 
         /// <summary>
+        /// Enumerates all objects on the heap and for each returns its address, type and size in an optimized way.
+        /// </summary>
+        /// <returns>An enumerator for all objects on the heap.</returns>
+        abstract public IEnumerable<Tuple<ulong, ulong, ClrType, ulong>> EnumerateObjectDetails();
+
+        /// <summary>
         /// Enumerates all objects on the heap.
         /// </summary>
         /// <returns>An enumerator for all objects on the heap.</returns>
@@ -584,6 +590,14 @@ namespace Microsoft.Diagnostics.Runtime
         /// 0 when there are no more objects.   (Or enumeration is not possible)  
         /// </summary>
         abstract public ulong NextObject(ulong objRef, out ClrType type);
+
+        /// <summary>
+        /// Given an object on the segment given its address, type and size,
+        /// return the 'next' object in the segment.
+        /// Returns 0 when there are no more objects.   (Or enumeration is not possible)
+        /// This is used for specific class histogram workflow.
+        /// </summary>
+        abstract internal ulong NextObject(ulong objRef, ClrType type, ulong size);
 
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
```diff
@@ -976,6 +990,43 @@ public override IEnumerable<ulong> EnumerateObjectAddresses()
...
         }
 
+
+        public override IEnumerable<Tuple<ulong, ulong, ClrType, ulong>> EnumerateObjectDetails()
+        {
+            ClrHeap heap = _segments[0].Heap;
+
+            // optimization for large dumps/apps with 30.000+ types:
+            //    keep track of ClrType given a method table address
+            Dictionary<ulong, ClrType> typeCache = new Dictionary<ulong, ClrType>(32768);
```
0xd4d: This is a pretty big allocation that will also always end up on the LOH. The dictionary allocates an int[] buckets list and an Entry[] entries list, both will be on the LOH. It's almost 1MB of data in 64-bit mode.

chrisnas: the goal is to avoid triggering garbage collections due to dictionary internal allocations due to array increase: this is important for large memory dumps/address space

src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
```diff
@@ -213,6 +213,12 @@ virtual public ClrType GetTypeByMethodTable(ulong methodTable)
         /// <returns>An enumerator for all objects on the heap.</returns>
         abstract public IEnumerable<ulong> EnumerateObjectAddresses();
 
+        /// <summary>
+        /// Enumerates all objects on the heap and for each returns its address, type and size in an optimized way.
+        /// </summary>
+        /// <returns>An enumerator for all objects on the heap.</returns>
+        abstract public IEnumerable<Tuple<ulong, ulong, ClrType, ulong>> EnumerateObjectDetails();
```
leculver: We should return a specific, documented type instead of a tuple.

src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
```diff
@@ -976,6 +990,43 @@ public override IEnumerable<ulong> EnumerateObjectAddresses()
...
+            //    keep track of ClrType given a method table address
+            Dictionary<ulong, ClrType> typeCache = new Dictionary<ulong, ClrType>(32768);
+
+            for (int iSegment = 0; iSegment < _segments.Length; ++iSegment)
+            {
+                ClrSegment segment = _segments[iSegment];
+
+                // unlike the default implementation, we are trying to avoid
+                // duplicate calls to GetObjectType and GetSize that are used by segment.NextObject()
+                // --> 25% faster on large memory dumps
```
leculver: That's a pretty surprising finding.  It's been a loooong time since I've profiled this (so are probably right), but the original intent of some of the implementation of ClrHeap.GetObjectType is to cache the previous call, and if the MethodTable is known we do a dictionary lookup anyway.  I'm surprised that this provides value outside of increasing the dictionary size for methodtables.

On the one hand, this makes me think that the solution here might not be to add another public method, but instead for me to figure out the underlying source of why this would be a 25% perf improvement and fix the underlying perf issue.

On the other hand, I'm actually not opposed to a more detailed EnumerateHeap method with these tradeoffs, assuming there's not something I can tweak in the current implementation to fix the perf problem you report here.

davidfowl: How do you stop things like this from regressing without performance tests? We could add some microbenchmarks with BDN (@adamsitnik )

src/Samples/ClassHistogram/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ClassHistogram
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            // build a "class histogram" (i.e. !dumpheap -stat in WinDBG+sos)
```
davidfowl: cc @anurse

src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
leculver: Agreed.  ClrMD is in dire need of more tests, perf and otherwise.  :)

src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
chrisnas: You're right: I did not know where to define it...

src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
chrisnas: @davidfowl: I've added a test just for this purpose
@leculver: caching in ClrMD is... complicated and I already found an issue related to caching (https://github.com/Microsoft/clrmd/issues/108). I did not feed confident enough to touch it: this is why I preferred to implement the dedicated new method

leculver:
I am closing some old pull requests.  I'm hoping where we land in ClrMD v2 will make this easier to implement on your end and not require a change to clrmd later.  If that turns out to not be the case let's chat and see if we can open up the library a bit more to suit your needs.

chrisnas:
> I am closing some old pull requests. I'm hoping where we land in ClrMD v2 will make this easier to implement on your end and not require a change to clrmd later. If that turns out to not be the case let's chat and see if we can open up the library a bit more to suit your needs.

Any timeline for ClrMD v2 @leculver ?

leculver:
@chrisnas I can't fully commit to any timeline since my schedule isn't completely my own.  My _plan_ if all goes well is something usable in pre-release on GitHub by January with most of the major architectural changes completed.

As for when to ship it as non prerelease I plan to either:

1.  Leave it prerelease for 3-4 months as I use it heavily to make sure I'm satisfied with the surface area and that it doesn't need further changes, then ship it for real.  Essentially it'd be usable in Jan, but wouldn't stop evolving/changing until April/May when it goes live.
2.  Ship 2.0 in mid January as a regular nuget package, but it would be a short release.  I'd plan to ship 2.1 ~6 months after that with further breaking changes.  (Basically all of the stuff I would have changed giving myself a longer window in 1.)

I'm leaning towards the first option, but I'll put this on the roadmap and in issue #311 for more feedback.


I fully intend to leave clrmd v2 in prerelease mode for a while as I make sure I haven't missed anything and to use it myself for

chrisnas:
Thanks for these details and I fully understand your "constraints".
We have a few tools based on ClrMD (such as what is available at https://github.com/chrisnas/DebuggingExtensions) and we'll try to test the different ClrMD pre-versions from github.
I'll follow https://github.com/microsoft/clrmd/issues/311

PR #130, opened 2018-11-06
Update to the latest version of arcade

anurse:
This is the first phase in getting some real signed builds up to MyGet ;).

This just updates the build to the latest version of arcade. No Azure DevOps (AzDO) support yet, but this will make it a lot easier to use the standard AzDO tools we've already got in place.

I believe I preserved the necessary MSBuild settings. Arcade generally uses a top-level `Directory.Build.props`, so I moved the one from `src` up. We can add lower level files if we need to configure individual folders differently.

I removed some content that I don't believe are necessary with the latest arcade:
* `BuildToolsVersion.txt` - Not used anymore
* `DotnetCLIVersion.txt` - Not used anymore
* `artifacts/toolset` directory - Not used anymore
* `eng/Versions.props` had some extra stuff specifying dependency versions but I didn't see those dependencies in use
* `netci.groovy` - AFAICT there isn't a Jenkins build active on this repo, so this is unnecessary, it's in source control history if needed again :)
* `sign.proj` - Arcade handles signing (or *will* handle, once we get AzDO up and running)
* `AssemblyInfo._` - Wasn't used (file extension was changed) and all that info is in `Directory.Build.props` now
* `pkg/Microsoft.Diagnostics.Runtime.nuspec` - Not needed, SDK-style projects don't need a NuSpec, the metadata is in the project! 🎉 	

Some stylistic choices I made based on other repos (but am happy to revert if you prefer other patterns):

* Moved the strong-name key up to the `eng` directory.
* Moved the License file in to the project itself (since the new Sdk-style project handles packaging).
* Changed `build.cmd` to produce packages by default (so now it does Restore, Build *and* Pack now). This is also just a style choice. The package is the primary artifact IMO and it takes very little time to build once the DLL is ready, so we generally make it part of the main build. Totally happy to revert that if you feel differently!

cc @pakrym @davidfowl

Next steps:
* Add source-mirroring to dev.azure.com/dnceng
* Add AzDO PR build from GitHub
* Add AzDO Official build from AzDO mirror with real signing and publishing to MyGet


leculver:
Thanks for your help here!

Issue #126, opened 2018-10-22
Access violation in finalizers


0xd4d:
It seems the bug is that DacLibrary's dtor is called before other dtors that depend on that library being present in memory.

https://github.com/Microsoft/clrmd/blob/e37fb3a9e419ae09c7e2ac0a66ba0afb4c264b6c/src/Microsoft.Diagnostics.Runtime/datatarget.cs#L1261

vs

https://github.com/Microsoft/clrmd/blob/e37fb3a9e419ae09c7e2ac0a66ba0afb4c264b6c/src/Microsoft.Diagnostics.Runtime/ComWrappers/CallableCOMWrapper.cs#L82

which tries to do this:

https://github.com/Microsoft/clrmd/blob/e37fb3a9e419ae09c7e2ac0a66ba0afb4c264b6c/src/Microsoft.Diagnostics.Runtime/ComWrappers/CallableCOMWrapper.cs#L42

leculver:
Much appreciated, this was on my todo list to tackle today.

Also, please note the repo isn't in a fully finished and tested state.  I had to merge it over early to do some (internal) build work to support the new build system.  Please expect bugs and crashes for a short while as I finish build work and get this fully tested.  (Always happy for bug reports though!)

leculver:
This should be fixed with the most recent pull request merge.  I haven't specifically gone to test this so please do reopen this issue if you hit it again.  It's still on my todo list to ensure proper addref/release counts and that the daclibrary doesn't free the library until everything is unloaded.

0xd4d:
I think it can only happen when the application is exiting in which case the dtors should check `Environment.HasShutdownStarted` and not do any cleanup if it's true.

leculver:
Thank you, I've reopened the issue and will take a closer look this week.

0xd4d:
I added a PR https://github.com/Microsoft/clrmd/pull/137

leculver:
Thank you for the PR!  I decided to go a slightly different direction to fix this:  https://github.com/Microsoft/clrmd/commit/4a16cfa3c42a1c5f8899aa772543a72dd9b63622

Instead of putting Dispose on ClrRuntime, I wanted to put that on DacLibrary instead, as I use DacLibrary directly in some of my projects.  I've also split out the code to actually release the library into its own class with its own lifetime.

[...]

PR #136, opened 2018-11-11
Don't free the dac lib when the app is exiting

0xd4d:
https://github.com/Microsoft/clrmd/issues/126

This issue happens when the app is exiting. Other dtors assume that their dtor is called before the dac lib dtor but this isn't true when the app is exiting.


0xd4d:
I'm closing this, a better fix is to add ref counting to the dac lib. `Environment.HasShutdownStarted` can also be true if the appdomain is being unloaded.

PR #137, opened 2018-11-11
Add ref counting to DacLibrary to prevent AV when app is exiting

0xd4d:
#126

This fixes an AV that can happen when the application is shutting down. CallableCOMWrapper holds a ref to the DAC lib, but the CallableCOMWrapper dtor assumes that the DAC lib dtor is always called after its own dtor. This is not always true when the app is shutting down.

ClrRuntime now implements IDisposable and has a dtor so DacLibrary can be freed.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrInfo.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrInfo.cs
@@ -68,21 +68,27 @@ namespace Microsoft.Diagnostics.Runtime
         public ClrRuntime CreateRuntime(object clrDataProcess)
         {
             DacLibrary lib = new DacLibrary(_dataTarget, DacLibrary.TryGetDacPtr(clrDataProcess));
-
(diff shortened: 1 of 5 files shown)


leculver:
Thank you for the PR!  I decided to go a slightly different direction to fix this:  https://github.com/Microsoft/clrmd/commit/4a16cfa3c42a1c5f8899aa772543a72dd9b63622

Instead of putting Dispose on ClrRuntime, I wanted to put that on DacLibrary instead, as I use DacLibrary directly in some of my projects.  I've also split out the code to actually release the library into its own class with its own lifetime.

I'm not done with lifetime management here.  I still want the call to DataTarget.Dispose to clean up everything instead of waiting for finalizers.  I also need to go through and fully test all the addref/release to make sure everything is clean.

For now though, this should stop the AVs.  If it doesn't please let me know and I'll reopen the issue and take another swing.

PR #151, opened 2018-11-28
extract class declarations into separate files, update project structure

isakiev:
Extracted all as discussed. Please re-adjust if you feel that something is out of place.


leculver:
There’s a few changes I’d ask though before accepting this PR, if you don’t mind:

1.	Can you rename all “Src” to lower case “src”?
2.	Can you add #pragma warning disable for 0649 and 0169 to ISOSNativeVTable (and any other locations that are generating warnings)?

A couple other questions/notes about this PR:
1.	Do you mind if I edit the history to set myself as the author (with you credited in the comments section)?  With these files potentially not being touched for years at a time I think this would make things a bit clearer that this was just a file move operation.
2.	I checked in a change today that conflicts, but it should be easy to resolve:  https://github.com/Microsoft/clrmd/commit/54aea35870f284a7bb24a53dd653852ca007f661

isakiev:
@leculver:
1.1. Done.
1.2. Done. No compile-time warnings anymore.
2.1. Totally agree. It would be great to have you as an author of all files. I don't think that such a change requires any mentions of myself. I'm just interested in the result.
2.2. Synched.

drewnoakes:
This PR removes the copyright headers from pretty much every file.

```c#
// Copyright (c) Microsoft. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```

drewnoakes:
It will also probably break every open pull request.

@isakiev you can explicitly set @leculver as the author when committing. Or you can use `git filter-branch` as described here: https://help.github.com/articles/changing-author-info/

leculver:
@drewnoakes This is change 1 of 2, I'm coordinating with Ruslan via email.  The copyright at the top of files are wrong in general with ClrMD.  They need to be changed over to look like the ones in the corefx branch.  A code-cleanup checkin will follow this one.

It's unfortunate that this breaks some outstanding code, but this is essentially a one-time cleanup that I've been putting off doing for 5+ years.  All of the recent changes to the guts of ClrMD likely broke a lot of outstanding PRs to begin with, this is basically the only window of time I foresee being able to make structural changes to the repo.

PR #152, opened 2018-11-28
Formatting, typos and update some information

drewnoakes:
Mostly just typos and formatting, however I also updated the section that claimed AppDomain data was limited to just a few properties as you can now do quite a bit more with them.

`FinalizerQueue` was changed from a property to the method `EnumerateFinalizerQueueObjectAddresses()`, so the code sample was updated.


leculver:
Awesome, thanks for the fixes.

PR #160, opened 2018-12-06
Expose context of stack frames

tidecarv:
This change is to expose the context of stack frames. This change is not applicable to stack frames extracted from an exception where this new field is set to null.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrStackFrame.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrStackFrame.cs
@@ -35,6 +35,11 @@ namespace Microsoft.Diagnostics.Runtime
     public abstract class ClrStackFrame
     {
         /// <summary>
+        /// Gets this stack frame context.
+        /// </summary>
+        public abstract byte[] Context { get; }
+
+        /// <summary>
         /// Returns the thread this stack frame came from.
         /// </summary>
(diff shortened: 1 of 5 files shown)


leculver:
Sorry for the delay.  I'm mildly concerned about returning a raw byte array since the caller can mutate the array and affect other callers.

However, we don't (currently) have a way to return IReadOnlyArray (or similar) in ClrMD.  I'll merge this now but probably will change the return based on https://github.com/Microsoft/clrmd/issues/163.

tidecarv:
Note that mutating the array will not affect the unwinding logic, the array is a copy of the original array because that one is reused during unwinding. Where do you see changes to the array causing side effects?

leculver:
@tidecarv Sorry, I missed this reply earlier.  There's a lot of comments and traffic to keep up with, but if you "@leculver" I will get the message.

Imagine two libraries in one process sharing one instance of ClrMD:

1.  Library A calls ClrStackFrame.Context, uses it, and for some reason mutates it (like zeroes it out).
2.  Library B later calls into the same ClrStackFrame.Context and receives the zero'ed out byte[] that Library A modified, causing issues.

Marking the value as IReadOnlyArray (or similar) solves that problem.  But it's a problem for another day, we'll use byte[] for now.  I may change the return type later to be read-only, but that would be in a much later release.

PR #168, opened 2018-12-11
Introduce NotNull and HasType asserts

mitikov:


mitikov:
@leculver please take a look at this PR.
Thanks.

leculver:
There's no description here so I'm not sure what the intent of this PR is.  I don't think this is generally applicable to many users of ClrMD.  This seems more like something that should go in the library/application that consumes ClrMD and not the library itself.  Sorry I can't accept this one.

Unrelated, if I were to give some feedback on the code here:  Assert _generally_ implies that it's only done under DEBUG builds.  Also, asserts generally should never be triggered unless there is some kind of code issue in the implementing library, which means you would want to pop a dialog, tear down the process, or otherwise halt (similar to how Debug.Assert is done).  In this case throwing a catchable exception is fine if that's the intent, but I'd just not call the class "*Assert" in that case.  =)

mitikov:
@leculver, sorry for an empty description, gimme a change to make it right this time :)

[Framework design guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/parameter-design) recommend having argument validation for parameters.

Doing validation with Debug set would leave us with phantom `NullRefs` (at best) in production code or letting objects be initialized with improper state = instance state would [not allow it to do what it is designed to do](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/exception-throwing) = bring unexpected results.
The choice is between paying `tiny` price for null check and have solid data if things go wrong in prod.
A good example is insurance - you never dream of using it, but still pay for it :)

Considering the library performance is Disk-bound over CPU, buying ensurance for tiny CPU sounds reasonable to me.

Options on the desk now for doing null-checks are:

- manually call `if (...) throw new ArgumentException...` in every method

- Use `Contract.Requires()` with repeatable conditions `(ClrObject.Type !=null)`

Both options are smelly as have 'code duplications' - should we want to introduce own `ClrObjectTypeMissingException`, we'll have to do tons of copy-paste then.

As for the class naming - I'll be happy to pick/accept any name you prefer, we can go with ClrContract class, or extension methods for `ClrObject` that will accept argument name.

I was trying to mimic well-known `Assert` syntax from unit testsing frameworks to make API easy discoverable.

Do let me know your thoughts :)

mitikov:
@leculver In the prev. ClrMD version (i'm migrating from it now) asserts were widely used as we could easily get an object without type.

Checking commit history it seems we [no longer welcome/expect ClrObject to be typeless](https://github.com/Microsoft/clrmd/commit/ac9f7b981bd8aab51f04093f38fff04431b0d49d), is that correct? If so, I would agree with not letting ClrObject construction without type 👍 

Keeping in mind struct-based nature, `ClrObject` still could be created without constructor execution (like `default(ClrObject)` or array) -> to properly implement [NotNull] contract for Type, we might need to rework Type getter to test each time.

Key questions:
A) We do not expect ClrObjects without types +/- ? 
B) If so, do we want to improve Type getter to guarantee its not null ? +/- 

Thanks.
P.s. pls do not merge the PR now.

mitikov:
It seems better now to introduce contracts requiring/enforcing types for base entity mentioned in  #207

leculver:
I'd like to close this one out now that I'm about to undergo a big refactor of ClrMD for v2.  I'm not opposed to these kinds of checks and contracts in the future, but I'd like to give them some more thought after the main refactor of ClrMD is done.

PR #169, opened 2018-12-11
Introduce Extension methods for locating chained field value

mitikov:


mitikov:
@leculver Please take a look at this PR.
Thanks.

leculver:
I'm sorry, I don't think I want to add this to ClrMD.  I used to have a method like this one in the library itself, but I intentionally removed it.

On the one hand, this is a very difficult algorithm to get right, which almost demands we put something into the library.  For instance, your code here would work fine if all of the fields in question are objects, but if you are dereferencing "m_foo.m_bar.m_baz" and "m_bar" happens to be a struct, I don't think that code will work.

On the other hand, in building code like this you have to make a lot of "policy" decisions about how the code should work.  I'm a bit hesitant to build that kind of policy into ClrMD itself.  I attempted that a few times before, even using "dynamic" in places to make this more user friendly, but at the end of the day I'd like to keep ClrMD pretty low level and make the person building on top of it decide how they want to use it.

mitikov:
@leculver  I understand your point.

While I can bring the extension method to my library and will use it from there, I want to give a use-case before doing so :)

The App I built is a mapper between memory snapshot values and hand-written models.
At first I need to locate a `ClrObject` in memory which is usually found by navigating an object chain, like staticObj => fieldManager -> fieldProvider -> fieldCache -> fieldInnerStorage

This time I know for sure all are the objects in the chain and using the extension method saves the day.

Policies that work for me:

- As soon as field is not there -> exception -> assembly loaded into captured process has different version = field layout changed = the chain we use is likely outdated + must report what has changed
- As soon as null object met -> return null -> something was not yet assigned
- As soon as valueType met -> exception -> API documentation clearly stands for `only objects` use case

Thanks.

PR #174, opened 2018-12-12
extension point for custom dac search strategies, minor improvements

isakiev:
The PR:
- attempts to organize knowledge about Dac and responsibilities for ClrRuntime creation
- provides extension point for custom Dac search policies (e.g. in the same folder with the dump file, in a zip, etc.)
- provides extension point for custom logging

In the ideal case, I'd prefer to make DacLocator and SymbolLocator readonly and pass them only once during DataTarget creation.
But the changes were made to match the current approach of SymbolLocator initialization in DataTarget.

There are some breaking changes in api:
- DacInfo removed
- ClrInfo.LocalMatchingDac removed
- SymbolLocator.FindBinary(DacInfo) removed

the PR can be splitted into smaller ones and is open for comments


src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
```diff
@@ -26,122 +31,17 @@ public class ClrInfo : IComparable
...
         /// <summary>
         /// Returns module information about the ClrInstance.
         /// </summary>
         public ModuleInfo ModuleInfo { get; }
 
-        /// <summary>
-        /// Returns the location of the local dac on your machine which matches this version of Clr, or null
-        /// if one could not be found.
-        /// </summary>
-        public string LocalMatchingDac { get; }
```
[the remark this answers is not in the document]
leculver: I don't believe the DacFileName/DacRequestFileName options provide this info as well, this would be removing functionality, but something the user could theoretically implement themselves.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
```diff
@@ -26,122 +31,17 @@ public class ClrInfo : IComparable
...
-            else
-            {
-                // Assume future versions will all work on the newest runtime version.
-                return new V45Runtime(this, _dataTarget, lib);
-            }
-
-            return new LegacyRuntime(this, _dataTarget, lib, ver, Version.Patch);
-        }
+        public string DacFileName => ClrInfoProvider.GetDacFileName(Flavor, IsLinux);
+        public string DacRequestFileName => ClrInfoProvider.GetDacRequestFileName(Flavor, Architecture, Architecture, Version, IsLinux);
```
leculver: I see now why you want IsLinux, but you should go through DataTarget.GetPlatform() == Linux (or something similar).

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
```diff
@@ -152,52 +52,7 @@ public override string ToString()
...
-            if (Version.Minor != rhs.Minor)
-                return Version.Minor.CompareTo(rhs.Minor);
-
-            if (Version.Revision != rhs.Revision)
-                return Version.Revision.CompareTo(rhs.Revision);
-
-            return Version.Patch.CompareTo(rhs.Patch);
-        }
+        //for backward compatibility only
+        internal DataTarget DataTarget { get; }
```
leculver: Marking this public isn't a bad idea actually.  I like having "back-references" up the object chain to the object's logical parent.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfoProvider.cs
```diff
@@ -0,0 +1,86 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    internal static class ClrInfoProvider
```
leculver: I like this code, much cleaner than what I was doing.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfoProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        private const string c_desktopModuleName1 = "clr";
+        private const string c_desktopModuleName2 = "mscorwks";
+        private const string c_coreModuleName = "coreclr";
+        private const string c_linuxCoreModuleName = "libcoreclr";
+        private const string c_nativeModuleName = "mrt100_app";
+
+        private static bool TryGetModuleName(ModuleInfo moduleInfo, out string moduleName)
+        {
+            moduleName = Path.GetFileNameWithoutExtension(moduleInfo.FileName);
```
leculver: Nit:

    moduleName = Path.GetFileNameWithoutExtension(moduleInfo.FileName)?.ToLower();
    return moduleName != null;

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfoProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    internal static class ClrInfoProvider
+    {
+        private const string c_desktopModuleName1 = "clr";
+        private const string c_desktopModuleName2 = "mscorwks";
+        private const string c_coreModuleName = "coreclr";
+        private const string c_linuxCoreModuleName = "libcoreclr";
+        private const string c_nativeModuleName = "mrt100_app";
```
leculver: You can remove .Net Native stuff.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -176,6 +175,20 @@ public SymbolLocator SymbolLocator
             set => _symbolLocator = value;
         }
 
+        private IDacLocator _dacLocator;
+
+        public IDacLocator DacLocator
```
leculver: Have to document public properties.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DefaultDacLocator.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            _dataTarget = dataTarget ?? throw new ArgumentNullException(nameof(dataTarget));
+        }
+
+        public string FindDac(ClrInfo clrInfo)
+        {
+            if (clrInfo == null) throw new ArgumentNullException(nameof(clrInfo));
+
+            string moduleDirectory = Path.GetDirectoryName(clrInfo.ModuleInfo.FileName);
+            if (moduleDirectory == null)
```
leculver: This shouldn't happen, we use the FIleName to create clrInfo right?  More importantly we shouldn't throw here.  Theoretically we may still find the right dac in _dataTarget.SymbolLocator.FindBinary.

src/Microsoft.Diagnostics.Runtime/src/Utilities/Logging/DefaultLogger.cs
```diff
@@ -0,0 +1,19 @@
+namespace Microsoft.Diagnostics.Runtime.Utilities
```
leculver: This isn't actually used for anything in ClrMD, it's just a disconnected class.

In general I'm not opposed to adding logging capabilities to ClrMD.  However, if I accept a change to add logging capabilities it would have to meet this criteria:

1.  It should go in its own pull request.
2.  The checkin to add a logger should also go through the runtime and add the actual calls to the logger itself.  I don't like disconnected interfaces/classes without actually using them.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/IDacLocator.cs
```diff
@@ -0,0 +1,12 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    public interface IDacLocator
```
leculver: Do we really need this?  The idea behind the CreateRuntime method is that there's an overload for you to provide a path to the dac.  If you want to bypass all dac location in ClrMD, just find the dac yourself before calling CreateRuntime, then pass that local path to it.

leculver: Also, should "category" be a string or a well defined enum?  If it's an enum users can filter on standardized categories.

isakiev:
@leculver thank you for the detailed response!

there were several thoughts behind the IDacLocator:
- it's nice to have some a way to check if there is suitable dac from the ClrMD's point of view in advance, before the actual CreateRuntime call. I don't think that wrapping the call into try/catch is a good solution. Manual check will mean creating a copy of the logic elsewhere, which is also not that good and can easily get out of sync.
- same applies to the attempt to find matching dac among multiple libraries in a custom folder (try/catch vs in-advance-check)
- it allows to keep all dac location logic in one place
- it allows to fully override ClrMD behaviour (I had an issue with the current way of "dac request file names" creation which required such override, but this is probably a topic for a separate PR)

currently knowledge about dac is spread across the following entities:

`DataTargetImpl.InitVersions`
- knows where to find dac on local machine (not only knows location but also checks versions and has some other logic)
- knows about dac request file names ("agnostic" and regular ones)
`DacInfo`
- for some reason extends ModuleInfo but the only thing that is actually used there is file name;
other parameters, e.g. Timestamp and Size are set to fake values, SymbolLocator should remember to ignore them during requests.
- has some internal statics for file name manipulations
`ClrInfo.CreateRuntime`
- checks LocalMatchingDac and attempts to download dac from public servers
`SymbolLocator`
- remembers to ignore Timestamp and Size for dac requests

it looks like it can be improved

other comments:
- ClrInfo.CreateRuntime vs DataTarget.CreateRuntime: I agree, it's possible to make an intermediate step without breaking the api.
- ILogger: agree, no need for that until there will be any other usages except SymbolLocator. As for now it's possible to get away with just making SymbolLocator.Trace virtual.
- no need in making data readers public, true

what is not that clear right now however:
- why some of the readers are selected by the `CrashDumpReader` enum and others by the use of a separate method (`LoadCrashDump` vs `LoadCoreDump`)?
- assuming that there is an abstract .dmp file there is no convenient way to check what kind of dump is that and what method should be used to create a DataTarget.

leculver:
> why some of the readers are selected by the CrashDumpReader enum and others by the use of a separate method (LoadCrashDump vs LoadCoreDump)?

This is due to a (poor) design decision I made 7+ years ago.  `LoadCrashDump` actually accepts a wide variety of inputs if you use the DbgEng reader.  You can feed it .zip or .cab files that contain a crash dump, and the underlying DbgEng APIs will extract it automatically for you.  I should have only accepted raw minidumps to begin with.

`LoadCoreDump` does something more sensible:  It expects an ELF core dump to look at.  If I ever supported OS X, it would go through this codepath too and open the core dump to see if it is ELF/Mach-O and go from there.

It's a fair point that maybe LoadCrashDump and LoadCoreDump should be the same function that automatically figures out what you are attempting to load, but you run into the issue of: "I have a ZIP file and now I am obligated to open it and find a minidump (windows) or coredump (linux) and choose the right library...oh crap there's both a coredump and a minidump, what do I do now?"  It makes the behavior of the library very odd.

As to why not use an ENUM, it's because in order to use linux support you'd have to call the new overload `LoadCrashDump(WindowsOrLinuxEnum)` anyway....at that point might as well name it descriptively.

> [regarding dac loading] it looks like it can be improved

I definitely agree there.  A lot of this logic was written (and rewritten) 7+ years ago and I've left it alone mostly since then.  I very happy with a cleanup if we can find the right design.

> it's nice to have some a way to check if there is suitable dac from the ClrMD's point of view in advance, before the actual CreateRuntime call. I don't think that wrapping the call into try/catch is a good solution.
> it allows to fully override ClrMD behaviour

I think all of these points are covered by the original design with this overload of CreateRuntime, unless I'm missing something? https://github.com/Microsoft/clrmd/blob/6095cd0fbff574454e7af8a4731a9f626e6924be/src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs#L94

Let's say you have a method called `MyAlgorithmToFindTheDac` which has all of your `IDacLocator.FindDac` logic.  If I were trying to solve your issue above, I would write the following code.  I think this hits all of your points:  Doesn't need try/catch, can be done without calling CreateRuntime:

    class Current
    {
        void SomeFunction(DataTarget dataTarget)
        {
            foreach (ClrInfo clrInfo in dataTarget.ClrVersions)
            {
                string dacPath = MyAlgorithmToFindTheDac(clrInfo);
                if (string.IsNullOrWhiteSpace(dacPath))
                    continue;

                // You found the right dac at this point, CreateRuntime will never throw.
                ClrRuntime runtime = clrInfo.CreateRuntime(dacPath);
                ProcessOneClrVersion(runtime);
            }
        }
    }

Compare that to the proposed change you have below.  Unless I'm missing something here, this proposed design fails your try/catch criteria above:

    class Proposed : IDacLocator
    {
        void SomeFunction(DataTarget dataTarget)
        {
            dataTarget.DacLocator = this;
            foreach (ClrInfo clrInfo in dataTarget.ClrVersions)
            {
                // This code WILL have to throw if DacLocator.FindDac returns null...
                ClrRuntime runtime = clrInfo.CreateRuntime();
                ProcessOneClrVersion(runtime);
            }
        }

        public string FindDac(ClrInfo clrInfo) => MyAlgorithmToFindTheDac(clrInfo);
    }

I guess what I'm saying is, I'm not opposed to changing the design here to allow for custom Dac location strategy.  I DO like that it's more explicit than CreateRuntime(string, bool).  However, In order to change course here I basically want to understand the following things:

1.  What part of CreateRuntime(string, bool) is failing the criteria you've outlined?  It too allows you to fully replace the logic ClrMD uses to locate a runtime, avoid try/catch, and avoid calling CreateRuntime until you are certain you've found the right dac.
2.  What do we gain by adding IDacLocator?  In parallel to the first point, I don't see any new functionality enabled by adding IDacLocator, it's just a different approach to do roughly the same thing.

Or have I completely misunderstood this feature?  :)

PR #178, opened 2018-12-13
make DataTarget responsible for the ClrRuntime creation

isakiev:
follow-up for #174


leculver:
This is much better.  I still need to make a decision around marking ClrInfo.CreateRuntime obsolete or not.  I'll probably defer that until 1.1, and remove it in 1.2.  Definitely not something to deal with here.

Thanks!

isakiev:
Thank you!

PR #181, opened 2018-12-14
Feature/175 introduce unit tests

mitikov:
@leculver , adding unit testing project per #181 discussion.

Adding most popular unit testing framework set to begin with: [AutoFixture](https://github.com/AutoFixture/AutoFixture), [xUnit](https://xunit.github.io), [nSubstitute](http://nsubstitute.github.io).

As a next step simple test with `ClrObject` (thus needs fake `ClrType`) is introduced.

An exception (`GCDesc` property is internal = cannot be touched = failure) takes place during `ClrType` mock built process.

Further commits lift visibility for from `internal` to `protected internal` allowing those be touched by derived = mock generation.

Thanks.


leculver:
@mitikov I've merged just the access changes to make sure clrmd 1.0.1 has these implemented.

PR #182, opened 2018-12-14
simplify dac location logic

isakiev:
follow-up for #174

All dac location logic is put into one place.
The PR is an intermediate step for possible further improvements.


src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
```diff
@@ -33,21 +32,15 @@ internal ClrInfo(DataTarget dt, ClrFlavor flavor, ModuleInfo module, DacInfo dac
...
         /// </summary>
         public ModuleInfo ModuleInfo { get; }
 
         /// <summary>
-        /// Returns the location of the local dac on your machine which matches this version of Clr, or null
-        /// if one could not be found.
+        /// Target platform.
         /// </summary>
-        public string LocalMatchingDac { get; }
+        internal Platform Platform { get; }
```
leculver: Should this be public?  Is it useful to know the target coredump/crashdump is from Linux/Windows?

Maybe not in the library's current state...

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
```diff
@@ -33,21 +32,15 @@ internal ClrInfo(DataTarget dt, ClrFlavor flavor, ModuleInfo module, DacInfo dac
         /// </summary>
         public ClrFlavor Flavor { get; }
 
-        /// <summary>
-        /// Returns module information about the Dac needed create a ClrRuntime instance for this runtime.
-        /// </summary>
-        public DacInfo DacInfo { get; }
```
leculver: You've gotten rid of "DacInfo" here, but tracking information about the dac needed to debug the runtime is 100% something that I want to expose from this library.  For most external users, this is just an implementation detail to get ClrMD functional.

For internal users (myself included), I actually need a property that exposes the filename, timestamp, and filesize that the dac is expected to be indexed under.

In retrospect I shouldn't have made DacInfo inherit from ModuleInfo, but we do need to keep this class.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -256,15 +256,28 @@ internal ClrRuntime CreateRuntime(ClrInfo clrInfo)
...
+
+            string dac;
+            if (File.Exists(localDacFilePath) && PlatformFunctions.IsEqualFileVersion(localDacFilePath, clrInfo.Version))
+            {
+                dac = localDacFilePath;
+            }
+            else
+            {
+                string requestFileName = ClrInfoProvider.GetDacRequestFileName(clrInfo.Flavor, Architecture, Architecture, clrInfo.Version, clrInfo.Platform);
+                dac = SymbolLocator.FindBinary(requestFileName, 0, 0, false);
```
leculver: Why do you request the dac with properties `0, 0`?  This would not find the correct Dac.  The purpose of DacInfo is to track the dac long-name as well as the filesize/timestamp of clr.dll itself (these are the properties required to request the dac).

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DacInfo.cs
```diff
@@ -1,59 +0,0 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
```
leculver: This class must exist in some form...

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
isakiev: marked internal in order to minimize changes to the public api at the current moment

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
isakiev: thanks for the details, it was intended to inline DacInfo into ClrInfo, but I'll restore this as requested

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
isakiev: thanks for the details, looks like a misinterpretation from my side

PR #191, opened 2018-12-18
Add logic contracts for ClrObject APIs

mitikov:
@leculver Produce sample unit tests for ClrObject APIs.

The outcome is not only to have test harness for safety, but also to have code-by-example documentation (how would the API behave in that scenario).

Should I want to use the API, I have to test it on my own now with runtime values ( f.e. attach to simple console app & inspect values in debugger).

Once we have a solid coverage, one may use tests as example to speed up clrMD-based development.

I recall my last PR affected your build server - please be carefull (might need to inject specific project props).

Thanks :)


leculver:
Thank you for resubmitting this!  There's one issue here before I can merge this.  I'm hitting a warning (converted to error) in Microsoft.Diagnostics.Runtime.UnitTests.csproj:  `Castle.Core 4.0.0 depends on System.ComponentModel.TypeConverter (>= 4.0.1) but System.ComponentModel.TypeConverter 4.0.1 was not found. An approximate best match of System.ComponentModel.TypeConverter 4.1.0 was resolved.`

Even though it's just a warning, it will cause the build server to fail when that warning is converted to an error (which is what build.cmd does).  Do you have any idea how to resolve that (without suppressing it)?

mitikov:
@leculver Referencing latest Castle.Core should resolve the warning as we explicitly tell which version we want.

Please let me know if that works for you.

Thanks.

leculver:
Looks good to me, thanks for the PR!

In the future, do note that I prefer to use the concrete type instead of `var` for ClrMD itself.  (I don't care in tests though, so I am happy to merge this as-is.)

PR #196, opened 2018-12-20
VersionInfo equality and comparison

isakiev:
- make VersionInfo immutable
- provide IComparable, IEquatable, GetHashCode etc.
- simplify ClrInfo comparison


leculver:
This looks good, but it might break some external code.  I need to build a 1.1 branch so we can merge this there.

I'm currently on vacation and won't be back until the 26th or 27th, so I will take a look at this then.

leculver:
I feel like my previous reply was a bit open ended:  This change looks good, but since it's a breaking change I plan to merge it into a branch for the 1.1 release, whenever I create that.  I'm leaving this open until I'm ready for that, but I do plan to merge it.

isakiev:
Sure) Thank you for the response!
That's not something critical, just a small "nice-to-have" thing.

PR #200, opened 2018-12-20
fix dac request file name

isakiev:
Throughout ClrMD there are multiple checks which allow dump processing only when current process architecture is equal to the dump architecture.
Dac info construction logic however uses current architecture to create dac request file name.

This may lead to the undesired effects.
E.g. you may want to try creating a DataTarget for a x86 dump in a x64 process.
Just to get general info and check if suitable dac can be found, without actual `ClrRuntime` creation (which will be done afterwards in a separate process with matching architecture)
But this will not work properly because of the incorrect file name in the dac info.

Seems to be a valid usecase. Or am I missing something?

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTargetImpl.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTargetImpl.cs
@@ -136,13 +136,12 @@ namespace Microsoft.Diagnostics.Runtime
 
                 VersionInfo version = module.Version;
                 string dacAgnosticName = DacInfo.GetDacRequestFileName(flavor, Architecture, Architecture, version);
-                string dacFileName = DacInfo.GetDacRequestFileName(flavor, IntPtr.Size == 4 ? Architecture.X86 : Architecture.Amd64, Architecture, version);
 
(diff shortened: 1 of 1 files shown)


leculver:
The "long-named dac" has two architectures in it, one for the debugger's architecture and one for the target process (crash dump) architecture.  Some of these combinations actually exist, for example we created an x86 dac that can target an ARM32 process.  The dac for that version is something like 
`mscordacwks_x86_arm_version.dll`.

So the code you see there is intentional, but it clearly doesn't do the right thing in the case where you are debugging an arm32 process on an actual arm32 machine (`...arm_arm_version.dll`)...  The most common use case for internal folks with ClrMD targeting arm was looking at crash dumps on their regular x86/x64 machines, so this code was "right" in most cases.

Ignoring arm for a moment, CLR has theoretically talked about doing an x64_x86 version of the dac (allowing you to use x64 versions of the debugger to debug a 32bit runtime), but every time someone tried to do this eventually they said "no, this is too much work and too difficult" and gave up.

The summary of all of this is:  The code as written is correct (other than arm <==> arm) and intentional.  It's _theoretically_ possible we could build some of those other cross-architecture dacs to support other scenarios, but if we ever did the x86 <=> x64 one I'd have to modify ClrMD to handle it anyway (too many uses of IntPtr.Size).

PR #206, opened 2018-12-24
Feature/define array processing

mitikov:
Define how arrays are treated here.

Key aspects - only primitive struct types arrays are read.
Neither DateTime, nor Guid, nor any other hand-written structs must be processed manually.
While the pointer to struct start is located perfectly


mitikov:
@leculver Please take a look.

Thanks.

leculver:
Looks good to me, thank you!

PR #207, opened 2018-12-30
Feature/unify clr object and clr value class

mitikov:
ClrObject and ClrValueClass has common API that highlights same logical operations are driven by these entities:

1.  Read the primitive value from known field

1.  Read the known reference field

1. Report your type

1. Report your Address

Even though similar API semantics, calling code has to deal those differently which complicates consumer code.

Extracting the common functionality allows unifying those operations and provides a background for exposing less implementation detail to calling code.


mitikov:
Hi @leculver , please take a look.

Thanks & Happy Holidays!

src/Microsoft.Diagnostics.Runtime/src/Common/IAddressableTypedEntity.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    /// <summary>
+    /// Represents an addressable entity (class or struct) with associated type.
+    /// </summary>
+    public interface IAddressableTypedEntity: IEquatable<IAddressableTypedEntity>
+    {
+        ulong Address { get; }
```
leculver: Can you copy/paste the documentation strings out of ClrObject onto the properties/methods in this interface?

src/Microsoft.Diagnostics.Runtime/src/Common/IAddressableTypedEntity.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        string HexAddress { get; }
```
leculver: Can you remove HexAddress?  It's something I shouldn't have added in the first place, and I'd like to not carry it forward.

src/Microsoft.Diagnostics.Runtime/src/Extensions/IAddressableTypedEntityExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Gets the field value from <see cref="IAddressableTypedEntity"/> with respect to field nature (either reference, or value type).
+        /// </summary>
+        /// <param name="entity">The entity to read field value from.</param>
+        /// <param name="fieldName">The name of the field to get value.</param>
+        /// <exception cref="ArgumentNullException">if entity has no type.</exception>
+        /// <exception cref="ArgumentException">if field with matching name was not found.</exception>
+        /// <returns></returns>
+        public static IAddressableTypedEntity GetFieldFrom(this IAddressableTypedEntity entity, string fieldName)
```
leculver: Please replace all instances of `var` with a concrete type in ClrMD itself.  (I don't care about the test projects.)

It makes the code a LOT more readable to show what type we are dealing with.

src/Microsoft.Diagnostics.Runtime/src/Extensions/IAddressableTypedEntityExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            var entityType = entity?.Type;
```
leculver: Nitpick:  Use ?? expression here:  `ClrType entityType = entity?.Type ?? throw new ArgumentNullException("No associated type", nameof(entity));`

Same below with GetFieldByName.

src/Microsoft.Diagnostics.Runtime/src/Extensions/IAddressableTypedEntityExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Diagnostics.Runtime.Extensions
```
leculver: I don't think we should add a new namespace here.  No one will find this extension method.  Put this in `Microsoft.Diagnostics.Runtime` instead.

leculver: With that said, I do like putting in "src\Extensions", that does seem like the right place for it, even if not using a new namespace.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrValueClass.cs
```diff
@@ -142,5 +142,8 @@ private ulong GetFieldAddress(string fieldName, ClrElementType element, string t
             ulong address = field.GetAddress(Address, _interior);
             return address;
         }
+
+        public bool Equals(IAddressableTypedEntity other)
+            => Address == other?.Address && Type == other?.Type;
```
leculver: NitPick:  `=> other != null && Address == other.Address && Type == other.Type`.  This avoids creating a `ulong?` with that comparison.

src/Microsoft.Diagnostics.Runtime/src/Common/IAddressableTypedEntity.cs
mitikov: I added code documentation slightly changed from ClrObject.

The difference is how exceptions are documented - I used to [follow MS guides](https://docs.microsoft.com/en-us/dotnet/csharp/codedoc) to let intelisence/resharper pick exceptions.

Please let me know if you prefer differently.

Thanks.

src/Microsoft.Diagnostics.Runtime/src/Common/IAddressableTypedEntity.cs
mitikov: Yup, agree. Interface already gives Address, and HexAddress seem to be like syntax sugar to print user-friendly addresses whenever needed.

src/Microsoft.Diagnostics.Runtime/src/Extensions/IAddressableTypedEntityExtensions.cs
mitikov: Yep, I had similar idea. Thanks.

src/Microsoft.Diagnostics.Runtime/src/Extensions/IAddressableTypedEntityExtensions.cs
mitikov: Good one, thanks.

I try my best to keep readability as top priority, and this gotcha looks really good :)

src/Microsoft.Diagnostics.Runtime/src/Common/ClrValueClass.cs
mitikov: Agree, done.

mitikov:
@leculver  Thanks for the comments, agree & resolved.

You might want to take a look at code comments and exception documentaiton style.

leculver:
Looks good now.  I'm not back in the office until Wednesday, so I will merge the changes then.

PR #212, opened 2019-01-08
Get name of JIT helper func / method table

### LINKED ISSUE #165: Add APIs to get name of JIT helper funcs and method tables
A feature request to get names of some things that is useful when showing a disassembly of a method body.

These could be exposed on ClrRuntime.

string GetJitHelperFunctionName(ulong address): use DacRequests.JIT_HELPER_FUNCTION_NAME & GetJitHelperFunctionName()

string GetMethodTableName(ulong address): use GetNameForMT()

They should return null if input isn't a jit helper function or a method table.

I can send a PR if needed.

0xd4d:
Closes #165

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
@@ -47,6 +47,7 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
         private DacGetGCInfoData _getGCHeapData;
         private DacGetCommonMethodTables _getCommonMethodTables;
         private DacGetCharArrayWithArg _getMethodTableName;
+        private DacGetByteArrayWithArg _getJitHelperFunctionName;
         private DacGetCharArrayWithArg _getPEFileName;
         private DacGetCharArrayWithArg _getAppDomainName;
         private DacGetCharArrayWithArg _getAssemblyName;
@@ -430,6 +431,12 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             return GetString(_getMethodTableName, mt);
         }
 
+        public string GetJitHelperFunctionName(ulong addr)
+        {
+            InitDelegate(ref _getJitHelperFunctionName, VTable->GetJitHelperFunctionName);
+            return GetAsciiString(_getJitHelperFunctionName, addr);
+        }
+
         public string GetPEFileName(ulong pefile)
         {
             InitDelegate(ref _getPEFileName, VTable->GetPEFileName);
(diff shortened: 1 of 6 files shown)


src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
```diff
@@ -462,6 +469,32 @@ private string GetString(DacGetCharArrayWithArg func, ulong addr, bool skipNull
...
+            
+            byte[] buffer = AcquireBuffer(needed);
+            hr = func(Self, addr, needed, buffer, out needed);
+            if (hr != S_OK)
+            {
+                ReleaseBuffer(buffer);
+                return null;
+            }
+
+            if (skipNull)
```
leculver: Maybe we could remove skipNull and be more permissive with what we accept?  I'm thinking something like:

    int len = Array.IndexOf(buffer, 0);
    if (len == -1)
        len = needed;
    string result = Encoding.ASCII.GetString(buffer, 0, len);

It's _slightly_ less efficient than assuming we know if the string will be null terminated or not, but the dac is a really fickle library (meaning, very poorly implemented).  I prefer to not assume that strings will/won't be properly null terminated.

src/Microsoft.Diagnostics.Runtime/src/Desktop/V45Runtime.cs
```diff
@@ -695,5 +695,15 @@ internal override uint GetExceptionHROffset()
...
         }
+
+        public override string GetJitHelperFunctionName(ulong addr)
+        {
+            return _sos.GetJitHelperFunctionName(addr);
+        }
+
+        public override string GetMethodTableName(ulong addr)
+        {
+            return GetNameForMT(addr);
```
leculver: Maybe consider just renaming GetNameForMT to be GetMethodTableName instead of carrying two functions?  I'm indifferent either way.  It's fine as is if you want to leave it.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
0xd4d: I've removed the `skipNull` arg. The method is a copy of the other GetString() method which reads UTF-16 data. If needed, the methods could be merged but would IMO be a little ugly with an extra 'isUnicode' or similar parameter.

src/Microsoft.Diagnostics.Runtime/src/Desktop/V45Runtime.cs
0xd4d: Fixed

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
leculver: I like having two separate methods.

PR #218, opened 2019-02-07
Fix problem with DAC name on Linux.

mikem8361:
The current code forgot to prepend "lib" to the name of the DAC module.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTargetImpl.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTargetImpl.cs
@@ -119,10 +119,7 @@ namespace Microsoft.Diagnostics.Runtime
 
                 bool isLinux = clrName == "libcoreclr";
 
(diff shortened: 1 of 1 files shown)


leculver:
LGTM, thank you!

mikem8361:
When and what will the package version be with this change?

leculver:
Packages are created manually.  If you need one with this change I can go have one built.

mikem8361:
Yes, I need one.  Most .NET Core repos publish the package in their official build automatically.

Thanks.

leculver:
I will put this together tomorrow.

ClrMD isn't one of the ones that get daily builds. =(

mikem8361:
You probably should wait to do this build/publish.  Now I’m running into the “LTTng-UST: Error (-17) while registering tracepoint probe. Duplicate registration of tracepoint probes having the same name is not allowed” crash because of the DAC and the hosting coreclr load/registering the tracepoints twice.  I worked around this in SOS by creating a symbol link of the DAC into a temp dir/file and loading that.  I know you ran into this a while back when you first were porting clrmd and I thought you worked around it.

In order for clrmd on Linux to work, it will need to do something similar.  I look into this and get back to you.

PR #222, opened 2019-02-11
GCRoot refactor

drewnoakes:
Lots of small commits to show what's going on. Hopefully they speak for themselves.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/GCRoot.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/GCRoot.cs
@@ -163,7 +163,14 @@ namespace Microsoft.Diagnostics.Runtime
 
                 if (parallel)
                 {
-                    Task<Tuple<LinkedList<ClrObject>, ClrRoot>> task = PathToParallelAsync(rootObject, rootFunc);
+                    Task<Tuple<LinkedList<ClrObject>, ClrRoot>> task = Task.Run(
+                        () =>
+                            {
+                                LinkedList<ClrObject> path = PathsTo(processedObjects, knownEndPoints, rootObject, target, unique, cancelToken).FirstOrDefault();
+                                return new Tuple<LinkedList<ClrObject>, ClrRoot>(path, path == null ? null : rootFunc());
+                            },
+                        cancelToken);
+
                     if (initial < tasks.Length)
                     {
                         tasks[initial++] = task;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/GCRoot.cs
```diff
@@ -100,37 +100,77 @@ public IEnumerable<GCRootPath> EnumerateGCRoots(ulong target, CancellationToken
         public IEnumerable<GCRootPath> EnumerateGCRoots(ulong target, bool unique, CancellationToken cancelToken)
         {
             Heap.BuildDependentHandleMap(cancelToken);
+
```
leculver: I can't reason about the changes in this file.  The GCRoot code is very sensitive to changes, as the main consideration is "what happens when this change is thrown at a process with a 32gb heap and tens of thousands of roots?"

There's not enough information in your pull request.  What problem are you trying to solve with this change?  What are your changes here doing?  Have you tested this on a sufficiently sized heap?

My primary concern with accepting this change is I don't have time to test it right now...

leculver: This applies to PR #226 as well.

drewnoakes: Apologies for the size of the diff here. I appreciate you're busy and have limited time for review. My intent was to provide a series of small commits that detail the steps one by one, making it easier to see intent and evaluate correctness.

While working on https://github.com/drewnoakes/string-theory I've been developing some heap scanning algorithms. That's grown into code that builds reference graphs, and I will offer that as a contribution to ClrMD in future.

When I started that work I wanted to understand how the GCRoot code worked, so was looking at it. In doing so I found what I believe to be mostly readability improvements via consolidation of similar code paths.

I do not have a 32gb dump to test on but if you can point me to one I'm happy to do that testing.

Non-readability changes:

- a5ce6d68c10180130861cdc8eeeda2c0ecce382c replaces `new Task().Start()` with `Task.Run` per guidance [here](https://blogs.msdn.microsoft.com/pfxteam/2010/06/13/task-factory-startnew-vs-new-task-start/)
- 0a37902db47ef5e96ed1fab0ce15abc43809c4dd adds `ClrHeap.GetObject` for convenience and discoverability

leculver:
Sorry, I will get back to this PR probably this weekend.  I'm buried under a mountain of work.

src/Microsoft.Diagnostics.Runtime/src/Common/GCRoot.cs
leculver: Ahh ok.  If these are mostly cosmetic changes (with the Task.Run change), I'm happy with the result then.  Thanks for the clarifications!

PR #223, opened 2019-02-11
Small bits

drewnoakes:

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ParallelObjectSet.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ParallelObjectSet.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Diagnostics.Runtime
             {
                 int offset = GetOffset(obj, seg);
 
-                lock (seg.Objects.SyncRoot)
+                lock (seg.Objects)
                     return seg.Objects[offset];
             }
 
@@ -31,7 +31,7 @@ namespace Microsoft.Diagnostics.Runtime
             {
                 int offset = GetOffset(obj, seg);
 
-                lock (seg.Objects.SyncRoot)
+                lock (seg.Objects)
                 {
                     if (seg.Objects[offset])
                     {
@@ -52,7 +52,7 @@ namespace Microsoft.Diagnostics.Runtime
             if (GetSegment(obj, out HeapHashSegment seg))
             {
                 int offset = GetOffset(obj, seg);
-                lock (seg.Objects.SyncRoot)
+                lock (seg.Objects)
                 {
                     if (seg.Objects[offset])
                     {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Runtime/src/Desktop/HandleRoot.cs
```diff
@@ -3,25 +3,32 @@
...
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Diagnostics.Runtime.Desktop
 {
     internal class HandleRoot : ClrRoot
     {
-        public HandleRoot(ulong addr, ulong obj, ClrType type, HandleType hndType, GCRootKind kind, ClrAppDomain domain)
+        private readonly Dictionary<HandleType, string> s_nameByHandleType = Enum.GetValues(typeof(HandleType))
```
leculver: This needs to be marked static.

src/Microsoft.Diagnostics.Runtime/src/Desktop/DesktopHeapType.cs
```diff
@@ -477,26 +477,24 @@ public override bool IsFinalizable
...
+                    if (field.ElementType == ClrElementType.Unknown)
+                        break;
 
-                        if (offset < field.Offset + size)
-                        {
-                            childField = field;
-                            childFieldOffset = offset - field.Offset;
-                            return true;
-                        }
+                    if (offset >= field.Offset &&
```
leculver: Nitpick:  This is much easier to read as:  `if (field.Offset <= offset && offset < field.Offset + field.Size)`.  That follows the rest of ClrMD's convention as well.

src/Microsoft.Diagnostics.Runtime/src/Common/ParallelObjectSet.cs
```diff
@@ -18,7 +18,7 @@ public override bool Contains(ulong obj)
             {
                 int offset = GetOffset(obj, seg);
 
-                lock (seg.Objects.SyncRoot)
+                lock (seg.Objects)
```
leculver: What's the reasoning behind this change?  I don't mind using the object itself instead of a syncroot, but I'm not clear why you made the change here.

src/Microsoft.Diagnostics.Runtime/src/Desktop/HandleRoot.cs
drewnoakes: Thanks, fixed.

src/Microsoft.Diagnostics.Runtime/src/Desktop/DesktopHeapType.cs
drewnoakes: Updated.

src/Microsoft.Diagnostics.Runtime/src/Common/ParallelObjectSet.cs
drewnoakes: I see `ICollection.SyncRoot` as essentially deprecated (based on the fact it's only implicitly implemented in newer BCL collections such as `List<T>`.

There's little functional difference in this change, though `BitArray.SyncRoot` does lazily allocate an object to return which this change avoids.

PR #224, opened 2019-02-11
Documentation tweaks

drewnoakes:
I'm not sure why so much of the documentation XML file was regenerated. Should this file even be in version control?


leculver:
Nope, that file shouldn't be in source control...  I will add it to the gitignore list at some point soon.

PR #225, opened 2019-02-11
Enable object reference enumeration with field offset

drewnoakes:
I wanted the ability to access field offsets during enumeration of object references and couldn't find a way to do that without doing an inner scan. It seems the information is available during the outer scan, and this adds API to expose that.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
@@ -12,6 +12,8 @@ namespace Microsoft.Diagnostics.Runtime
     internal abstract class HeapBase : ClrHeap
     {
         protected static readonly ClrObject[] s_emptyObjectSet = new ClrObject[0];
+        protected static readonly ClrObjectReference[] s_emptyObjectReferenceSet = new ClrObjectReference[0];
+
         private ulong _minAddr; // Smallest and largest segment in the GC heap.  Used to make SegmentForObject faster.  
         private ulong _maxAddr;
         private ClrSegment[] _segments;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
```diff
@@ -257,6 +259,34 @@ protected internal override IEnumerable<ClrObject> EnumerateObjectReferences(ulo
...
         }
 
+        protected internal override IEnumerable<ClrObjectReference> EnumerateObjectReferencesWithFields(ulong obj, ClrType type, bool carefully)
+        {
+            if (type == null)
+                type = GetObjectType(obj);
+            else
+                Debug.Assert(type == GetObjectType(obj));
+
+            if (!type.ContainsPointers)
```
leculver: `GetObjectType` can return null if the heap is in an inconsistent state.  This if check should be `if (type == null || !type.ContainsPointers)`.

src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
```diff
@@ -257,6 +259,34 @@ protected internal override IEnumerable<ClrObject> EnumerateObjectReferences(ulo
...
+
+            GCDesc gcdesc = type.GCDesc;
+            if (gcdesc == null)
+                return s_emptyObjectReferenceSet;
+
+            ulong size = type.GetSize(obj);
+            if (carefully)
+            {
+                ClrSegment seg = GetSegmentByAddress(obj);
+                if (seg == null || obj + size > seg.End || !seg.IsLarge && size > 85000)
```
leculver: Nitpick:  I prefer to put the unnecessary parenthesis in boolean expressions so that it's always obvious of the intent:

`if (seg == null || obj + size > seg.End || (!seg.IsLarge && size > 85000))`

drewnoakes:
Thanks for the review. I pushed updates for your comments, and applied the same changes to `EnumerateObjectReferences` which I based this new method on.

PR #226, opened 2019-02-11
Public and cancellation

drewnoakes:
This PR includes #222 so only review 50d9f7787d1d1a9507b5a154717cd315d0e1d064, 4355947f68d7d8cc9aa8392867c7a18e90311376, 628288ca2b5eba1c7c71fc105b8752eeeb81835b here.

Makes various APIs public

- ClrHeap.EnumerateStackRoots
- ClrHeap.EnumerateStrongHandles
- ~ParallelObjectSet~
- HandleRoot

Adds cancellation support to:

- ClrHeap.EnumerateStackRoots
- ClrHeap.EnumerateStrongHandles

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHeap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHeap.cs
@@ -160,11 +160,23 @@ namespace Microsoft.Diagnostics.Runtime
         {
         }
 
+        /// <summary>
+        /// Returns an enumerable sequence of stack roots from this heap.
+        /// </summary>
+        /// <param name="cancelToken">A token that allows cancellation of the operation.</param>
+        /// <returns>An enumerable sequence of stack roots from this heap.</returns>
+        /// <exception cref="NotImplementedException">Raised if the operation is not supported for this heap type.</exception>
         public virtual IEnumerable<ClrRoot> EnumerateStackRoots(CancellationToken cancelToken)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns an enumerable sequence of strong handles from this heap.
+        /// </summary>
+        /// <param name="cancelToken">A token that allows cancellation of the operation.</param>
+        /// <returns>An enumerable sequence of strong handles from this heap.</returns>
+        /// <exception cref="NotImplementedException">Raised if the operation is not supported for this heap type.</exception>
         public virtual IEnumerable<ClrHandle> EnumerateStrongHandles(CancellationToken cancelToken)
         {
             throw new NotImplementedException();
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHeap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHeap.cs
@@ -21,6 +21,17 @@ namespace Microsoft.Diagnostics.Runtime
         public abstract ClrType GetObjectType(ulong objRef);
 
         /// <summary>
+        /// Gets a <see cref="ClrObject"/> for the given address on this heap.
+        /// </summary>
+        /// <remarks>
+        /// The returned object will have a <c>null</c> <see cref="ClrObject.Type"/> if objRef does not point to
+        /// a valid managed object.
+        /// </remarks>
+        /// <param name="objRef"></param>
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ParallelObjectSet.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.Diagnostics.Runtime
 {
-    internal class ParallelObjectSet : ObjectSet
+    public class ParallelObjectSet : ObjectSet
```
leculver: Why is this public?  It's an implementation detail of the GCRoot code.  It was not meant for general consumption...  I don't expect external people to rebuild the GCRoot code from scratch, and if they did I would imagine they'd use a different abstraction.

Also, it's not supported to call ClrMD in any multithreaded environment (other than very narrow cases, like the GCRoot code), it would be a bit confusing to mark public something that I didn't intend others to use and only has a very narrow acceptable usage.

src/Microsoft.Diagnostics.Runtime/src/Desktop/HandleRoot.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.Diagnostics.Runtime.Desktop
 {
-    internal class HandleRoot : ClrRoot
+    public class HandleRoot : ClrRoot
```
leculver: Why is this class marked public?  HandleRoot only makes it so the space complexity of enumerating roots is lower (letting us generate objects that only contain unique data for the particular root kind).

src/Microsoft.Diagnostics.Runtime/src/Common/GCRoot.cs
```diff
@@ -100,37 +100,77 @@ public IEnumerable<GCRootPath> EnumerateGCRoots(ulong target, CancellationToken
         public IEnumerable<GCRootPath> EnumerateGCRoots(ulong target, bool unique, CancellationToken cancelToken)
         {
             Heap.BuildDependentHandleMap(cancelToken);
+
```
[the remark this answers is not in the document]
drewnoakes: This PR is based off the tip of #222. If that PR isn't merged I would re-work this PR from master. Let's just discuss 50d9f7787d1d1a9507b5a154717cd315d0e1d064 here.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrHeap.cs
```diff
@@ -149,12 +160,12 @@ protected internal virtual void BuildDependentHandleMap(CancellationToken cancel
         {
         }
 
-        protected internal virtual IEnumerable<ClrRoot> EnumerateStackRoots()
+        public virtual IEnumerable<ClrRoot> EnumerateStackRoots(CancellationToken cancelToken)
```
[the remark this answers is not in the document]
drewnoakes: Good point, I've added comments.

Note that `ClrHeap.EnumerateStackRoots` and `ClrHeap.EnumerateStrongHandles` currently throw `NotImplementedException` for all but `DesktopGCHeap`.

src/Microsoft.Diagnostics.Runtime/src/Common/ParallelObjectSet.cs
drewnoakes: I ended up backing out use of this class so personally don't need it to be public anymore. I've reverted this to be internal again.

src/Microsoft.Diagnostics.Runtime/src/Desktop/HandleRoot.cs
drewnoakes: The reason was that I wanted to project a sequence of `ClrHandle`s as `ClrRoot`s, to merge with another sequence of `ClrRoot`s.

Looking at this again I'm wondering why I didn't just use `ClrHeap.EnumerateRoots`. It seems I was influenced by the implementation of `GCRoot.EnumerateGCRoots`. The latter includes strong handles and stack roots, while the former also includes the finalizer queue and (optionally) static fields and thread locals.

I believe `ClrHeap.EnumerateRoots` will be a better fit for me. I'll close this PR.

PR #227, opened 2019-02-11
COM interop helper cleanup

leculver:
- Move COM interop helpers to Utilities namespace.
- Add manual AddRef/Release to COMCallableIUnknown.
- Ensure COMCallableIUnknown will only free data once.
- Allow different return types in VTableBuilder. Now only validate what would cause a crash.


leculver:
@mikem8361 This fixes the requested issues from this morning.

PR #232, opened 2019-02-21
Replacing PdbReader with SharpPdb nuget.

southpolenator:
Removing all PDB parsing from ClrMD in favor of existing library SharpPdb.
This will add support for Portable PDBs too. Embedded Portable PDBs support is still pending.


leculver:
Unfortunately we cannot add a library dependency to ClrMD.  I will look at marking PdbReader obsolete with a note pointing people to SharpPdb.

PR #236, opened 2019-03-06
[WIP]Support loading 64-bit dumps of 32-bit processes

poizan42:
Since 64-bit task manager always generates 64-bit dumps you often end up with those from users. It would be nice if clrmd could support those.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs
@@ -140,8 +140,8 @@ namespace Microsoft.Diagnostics.Runtime
         public Architecture GetArchitecture()
         {
             SetClientInstance();
-
-            int hr = _control.GetExecutingProcessorType(out IMAGE_FILE_MACHINE machineType);
+            SetEffectiveProcessorType();
+            int hr = _control.GetEffectiveProcessorType(out IMAGE_FILE_MACHINE machineType);
             if (0 != hr)
                 throw new ClrDiagnosticsException($"Failed to get processor type, HRESULT: {hr:x8}", ClrDiagnosticsException.HR.DebuggerError);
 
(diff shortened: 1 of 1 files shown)


poizan42:
This works right now by just hardcoding a change of the effective processor architecture, so won't work on other architectures.

I have to figure out how to detect that the process is really a 32-bit process, and also where in the code that logic belongs.

poizan42:
So I did a bit of research. I don't think there is any direct way to see that the process is of a 32-bit process. I can see 3 solutions:

* Get the main module (either through IDebugSystemObjects::GetCurrentProcessExecutableName by name or reading the $exentry pseudo-register), then check the PE header to see if is 32-bit. I don't think we are 100% guaranteed that this is included in the dump though.
* Check if $teb[0] != null - [this is a pointer to the TEB32 from Vista onwards](https://docs.microsoft.com/en-us/windows/desktop/api/winternl/ns-winternl-_teb). I don't know what happens if someone uses the native api to inject a pure 64-bit thread into a wow64 process though - maybe we just could check all the threads? Is this information guaranteed to be included in a minidump?
* Delay the decision until we have found the runtime dll's - then we can just check the architecture of those. Seems like the most reliable way since we need to be able to find them to do anything at all anyways. However this requires some refactorization of the code.

@leculver What do you think?

leculver:
@poizan42 Sorry for the delay, I'm just back from vacation.

Overall I'm generally fine with making this kind of corner case work, if it's possible.  I tend to look at this as "we have made a best-effort to make this unsupported scenario work, but it may be impossible to make all functionality work properly."

As long as it doesn't affect normal operation I'd accept a (completed) pull request for this kind of change.  FYI, task manager also tends to produce...strangely non-functional crash dumps.  Even when the architecture mismatch is accounted for.  I've never fully tracked down what's missing that's expected.  Maybe that's been fixed since I tested it out a long time ago...

leculver:
@poizan42 I'm going to be forking the branch to build the next version of ClrMD before too long.  What's the status of this PR?  It will likely get quickly out of date in the next week or so.

poizan42:
The most useful part of it is already merged in #289. The bigger issue is in attempting to autodetect whether we have a 32 or 64-bit dump. Which I think might need some larger refactorizations to do reliably because it seems that the only sure way is to actually find the runtime dll first and then check the bitness of it. I might try to work on it sometime in the future, but feel free to close it for now.

leculver:
Ok sounds good!  As I mention in Issue https://github.com/microsoft/clrmd/issues/311, I'm about to branch the codebase for ClrMD 2.0 and there will be major changes to the master branch as I move that project along.  Thanks!

PR #251, opened 2019-03-27
Reference JetBrains.Annotations for nullability markup

isakiev:
Just curious, will it be acceptable to use JetBrains.Annotations for nullability markup inside ClrMd?

All attributes are conditional and will not leave traces in resulting ClrMd assembly by default.
Resulting ClrMd NuGet package will not have any dependencies as well.
https://www.nuget.org/packages/JetBrains.Annotations

ClrMd apis are not always obvious/consistent in terms of nullability.
E.g. you can't say straight away whether an abstract method returning `IEnumerable` can return null or not, there are samples of both behaviors.
With explicit nullability markup it will be possible to statically validate both api usages and api implementations right in the IDE and get corresponding warnings in advance.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
+++ b/src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
@@ -33,4 +33,7 @@
       <PackagePath></PackagePath>
     </Content>
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="JetBrains.Annotations" Version="2019.1.1" />
+  </ItemGroup>
 </Project>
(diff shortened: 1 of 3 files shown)


isakiev:
@leculver
In context of the recent null-check discussions it is interesting to know what are your thoughts about explicit nullability markup inside ClrMd?

weltkante:
Since nullability is coming to C# 8 anyways in a few months I'd wait until release and use the language integrated version instead of a 3rd party one.

leculver:
I don't think I want to use 3rd party extensions for this.  I'm not fully committed to using the ! and ? operators throughout the libraries on release, but if I do decide to go that route I will use the C# 8 features to do this.

PR #263, opened 2019-05-03
add live (Passive) data reader for Linux

xinyanmsft:
This change introduces a IDataReader implementation that targets a dotnet core process runs on Linux. LinuxLiveDataReader implements IDataReader that uses /proc/<pid>/maps and /proc/<pid>/mem files. 

Currently thread context portion of the data reader is not implemented.


xinyanmsft:
ping :-)

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -0,0 +1,517 @@
...
+            }
+            catch (Exception)
+            {
+                return false;
+            }
+        }
+
+        public ulong ReadPointerUnsafe(ulong address)
+        {
+            byte[] ptrBuffer = new byte[IntPtr.Size];
```
leculver: Please make this an instance field on the class.  That way we don't constantly allocate when reading pointers.

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -0,0 +1,517 @@
...
+            if (!ReadMemory(address, ptrBuffer, IntPtr.Size, out int read))
+            {
+                return 0;
+            }
+            return IntPtr.Size == 4 ? BitConverter.ToUInt32(ptrBuffer, 0) : BitConverter.ToUInt64(ptrBuffer, 0);
+        }
+
+        public uint ReadDwordUnsafe(ulong address)
+        {
+            byte[] ptrBuffer = new byte[4];
```
leculver: Same here, please make this an instance field on the class.

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -0,0 +1,517 @@
...
+            if (!ReadMemory(address, ptrBuffer, 4, out int read))
+            {
+                return 0;
+            }
+            return BitConverter.ToUInt32(ptrBuffer, 0);
+        }
+
+        public ulong GetThreadTeb(uint thread)
+        {
+            throw new NotImplementedException();
```
leculver: I would return 0 here instead of throwing.

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -0,0 +1,517 @@
...
+            ctx->Rdx = registerSet.Rdx;
+            ctx->Rsi = registerSet.Rsi;
+            ctx->Rdi = registerSet.Rdi;
+            ctx->Rip = registerSet.Rip;
+            ctx->Rsp = registerSet.Rsp;
+        }
+
+        private void LoadThreads()
+        {
+            if (!_threadIDs.Any())
```
leculver: I'd prefer _threadIDs.Count != 0.  No need to use linq unless we have to.

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -0,0 +1,517 @@
...
+        public bool IsMinidump { get { return false; } }
+
+        public void Close()
+        {
+            _memoryStream?.Dispose();
+            _memoryStream = null;
+        }
+
+        public void Flush()
+        {
```
leculver: This should be:

    _threadIDs.Clear();
    _memoryMapEntries = null;
    _memoryStream = LoadMemoryMap();

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
xinyanmsft: sure

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
xinyanmsft: Okay

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
xinyanmsft: yes, changed to if (_threadIDs.Count == 0) ...

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
xinyanmsft: I see. changed to:
            _threadIDs.Clear();
            _memoryStream?.Dispose();
            _memoryStream = null;
            _memoryMapEntries = this.LoadMemoryMap();

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
xinyanmsft: yes

xinyanmsft:
I've updated the code to your request. Thanks!

leculver:
Merged, thank you!

PR #273, opened 2019-07-03
LinuxFunctions: don't require development packages to be installed for 'libdl.so'

tmds:
On glibc based Linux distributions, 'libdl.so' is a symlink provided by development packages.

In case we are running on .NET Core 3.0+, we can use System.Runtime.InteropServices.NativeLibrary.
On other versions, we fall back to 'libdl.so.2' when we can't find 'libdl.so'.

cc @mikem8361 @leculver @jkotas @janvorli


src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
```diff
@@ -9,6 +9,59 @@ namespace Microsoft.Diagnostics.Runtime
...
     internal sealed class LinuxFunctions : PlatformFunctions
     {
+        private const string LibDlGlibc = "libdl.so.2";
+        private const string LibDl = "libdl.so";
+
+        private readonly Func<string, IntPtr> _loadLibrary;
+        private readonly Func<IntPtr, bool> _freeLibrary;
+        private readonly Func<IntPtr, string, IntPtr> _getExport;
+
+        public LinuxFunctions()
```
0xd4d: Use a static constructor?

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
```diff
@@ -9,6 +9,59 @@ namespace Microsoft.Diagnostics.Runtime
...
+        {
+            Type nativeLibraryType = Type.GetType("System.Runtime.InteropServices.NativeLibrary, System.Runtime.InteropServices", throwOnError: false);
+            if (nativeLibraryType != null)
+            {
+                // .NET Core 3.0+
+                _loadLibrary = (Func<string, IntPtr>)nativeLibraryType.GetMethod("Load", new Type[] { typeof(string) })?.CreateDelegate(typeof(Func<string, IntPtr>));
```
0xd4d: `_loadLibrary`, `freeLibrary`, `_getExport` can all be `null` if the `GetMethod()` calls fail.

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
leculver: These are fine as instance.  For most of the library I do not like carrying state.

leculver:
Overall this looks good to me if you handle the edge case when GetMethod fails.

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
tmds: Why would it fail? The `NativeLibrary` methods we're using are part of the public corefx API.

leculver: 1.  Null checks are cheap.
2.  The coreclr runtime could change in the future.
3.  What if someone tries to run it on a runtime which has a different implementation?  (Mono.)
4.  It's always better to defensively code if it doesn't cost anything.

0xd4d: > Why would it fail? The `NativeLibrary` methods we're using are part of the public corefx API.

Since you use `?.`, you expect the calls to sometimes fail. Public APIs can change and be removed in future versions, even if it doesn't happen very often.

tmds: I've added the null checks.

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
```diff
@@ -9,6 +9,61 @@ namespace Microsoft.Diagnostics.Runtime
...
+                Action<IntPtr> freeLibrary = (Action<IntPtr>)nativeLibraryType.GetMethod("Free", new Type[] { typeof(IntPtr) })?.CreateDelegate(typeof(Action<IntPtr>));
+                _freeLibrary = ptr => { freeLibrary(ptr); return true; };
```
0xd4d: `_freeLibrary` (field) should be `null` if `freeLibrary` (local) is `null` since it's checked for `null` below

mikem8361:
@leculver is this ready to merge?

leculver:
@mikem8361 Merged.

PR #274, opened 2019-07-09
Fix LinuxFunctions.GetProcAddress throwing an exception

mikem8361:
Because it uses System.Runtime.InteropServices.NativeLibrary.GetExport instead of TryGetExport
it throws an exception when the function can't be found instead of return null. This breaks
the DacLibrary init code (DacLibrary.cs line 86) when it is try to get the DAC_PAL_InitializeDLL
export and then the PAL_InitializeDLL export.


leculver:
@mikem8361 I'll ship a new NuGet package tomorrow morning.

leculver:
@mikem8361 Released here: https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/1.1.35902

PR #275, opened 2019-07-09
Avoid crashing on Linux when getting the threads callstack

chrisnas:
While improving parallel stack to attach to process (both on Windows and Linux - https://github.com/chrisnas/DebuggingExtensions/blob/master/src/ParallelStacks.Runtime/ParallelStack.cs#L86), I've faced exceptions on linux.
After debugging and trying to figure out how to use ptrace, I've ended up to this fix: it is not needed to call ptrace on each thread (especially with this type of code/error handling)


leculver:
@chrisnas Thank you for the PR, I have incorporated this change into this release:  https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/1.1.35902

PR #276, opened 2019-07-11
Fix linux live data reader

mikem8361:
The module file name needs to be the full path for SOS to work and to match the core dump data reader.


mikem8361:
No need for an official build yet.

PR #278, opened 2019-07-16
Handle app domains nullability and absence of shared app domain in the latest CoreCLR

isakiev:
According to https://github.com/dotnet/coreclr/pull/21031 
there's no shared app domain in the latest CoreCLR.

ClrMd fails with nullref in attempt to get SharedDomainAddress in DesktopGCHeap.LoadAllTypes() on such dumps.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrRuntime.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrRuntime.cs
@@ -55,10 +55,9 @@ namespace Microsoft.Diagnostics.Runtime
         public abstract int PointerSize { get; }
 
         /// <summary>
-        /// Enumerates the list of appdomains in the process.  Note the System appdomain and Shared
-        /// AppDomain are omitted.
+        /// Enumerates the list of appdomains in the process.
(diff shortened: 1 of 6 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ClrRuntime.cs
```diff
@@ -55,10 +55,9 @@ public abstract class ClrRuntime
         public abstract int PointerSize { get; }
 
         /// <summary>
-        /// Enumerates the list of appdomains in the process.  Note the System appdomain and Shared
-        /// AppDomain are omitted.
+        /// Enumerates the list of appdomains in the process.
         /// </summary>
-        public abstract IList<ClrAppDomain> AppDomains { get; }
+        public abstract IReadOnlyList<ClrAppDomain> AppDomains { get; }
```
leculver: This is a breaking change.

leculver:
I need to ship a build with these changes, so I'll manually roll back the IReadOnlyList change.  I'll have that updated in 1.2.

isakiev:
@leculver thank you!

PR #292, opened 2019-09-09
Various fixes

mikem8361:
Change thread CONTEXT size check to return error if less than instead of not equal.

Make COM interop wrapper accessible

Add setter to ModuleInfo.IsManaged and change the ELF dump reader
to set it if not an ELF module assuming it is a PE file.

Add auxv structs/enums. Add module list filtering for the "interpreter"
and any module name that begins with "\dev\".


mikem8361:
@leculver

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
```diff
@@ -65,6 +65,8 @@ public virtual bool IsManaged
                 InitData();
                 return _managed;
             }
+
+            set => _managed = value;
```
leculver: Can you mark this `internal` if you don't need it to be public?  I plan to refactor this class one day to fix a bunch of stuff and it's easier if there's less external dependencies.  If you actually need it to be `public` that's fine too.

src/Microsoft.Diagnostics.Runtime/src/Linux/ElfCoreFile.cs
```diff
@@ -73,6 +81,39 @@ private IEnumerable<ElfNote> GetNotes(ElfNoteType type)
             return ElfFile.Notes.Where(n => n.Type == type);
         }
 
+        private void LoadAuxvTable()
+        {
+            if (_auxvEntries != null)
+                return;
+
+            _auxvEntries = new Dictionary<ulong, ulong>();
+            ElfNote auxvNote = GetNotes(ElfNoteType.Aux).Single();
```
leculver: Is there any chance the dump could have no aux notes?  (Especially if the coredump wasn't generated by create dump.  Maybe a smaller dump with less options set?)

I usually default to trying to double check these issues and not throw InvalidOperationException that Single will.  For example, maybe use FirstOrDefault and handle the null return case.  (Not sure what the right thing to do is if you can't find an aux table, but if an exception needs to be thrown I would suggest `ClrDiagnosticsException`...or whatever I named the thing that clrmd throws.  That or a bad image related exception.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
mikem8361: Done.

src/Microsoft.Diagnostics.Runtime/src/Linux/ElfCoreFile.cs
mikem8361: Done.

mikem8361:
@leculver I made the CR changes you request.  Thanks.

mikem8361:
Could you kick off an official buold.


-------- Original message --------

leculver:
@mikem8361 Yep, I had to fix https://github.com/microsoft/clrmd/commit/ab08995887f104bea6e78c2279cc4a7c0b712aaf, but that's done now.  Building official version now, will have the nuget package out tonight or tomorrow.

leculver:
@mikem8361 https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/1.1.46104  Dunno when it'll be finished validating and indexed, but should be ready tomorrow.

PR #304, opened 2019-10-23
Support WSL 1

0xd4d:
Support WSL 1. It doesn't support `/proc/<pid>/mem` but it does support `process_vm_readv()`.

The new code path is only executed if `/proc/self/mem` doesn't exist, i.e., it's WSL 1.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Diagnostics.Runtime.Linux
     {
         private List<MemoryMapEntry> _memoryMapEntries;
         private FileStream _memoryStream;
+        private bool _initializedMemFile;
         private List<uint> _threadIDs = new List<uint>();
         private byte[] _ptrBuffer = new byte[IntPtr.Size];
         private byte[] _dwordBuffer = new byte[4];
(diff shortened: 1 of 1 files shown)


leculver:
Much appreciated!

leculver:
@mikem8361 Would a similar technique work for dotnet-dump?  It would be SUCH an improvement to be able to test a lot of the clrmd work (and dump debugging) on WSL 1.

0xd4d:
https://github.com/dotnet/diagnostics/issues/548

Some coreclr files ([createdump](https://github.com/dotnet/coreclr/tree/master/src/debug/createdump)) would need to be updated

```grep
debug/createdump/datatarget.cpp:    _snprintf_s(memPath, sizeof(memPath), sizeof(memPath), "/proc/%lu/mem", m_pid);
debug/createdump/datatarget.h:    int m_fd;                           // /proc/<pid>/mem handle
debug/createdump/main.cpp:        // The initialize the data target's ReadVirtual support (opens /proc/$pid/mem)
debug/di/shimremotedatatarget.cpp:    int m_fd;                           // /proc/<pid>/mem handle
debug/di/shimremotedatatarget.cpp:    _snprintf_s(memPath, sizeof(memPath), sizeof(memPath), "/proc/%lu/mem", m_processId);
pal/src/thread/process.cpp:        // Gives the child process permission to use /proc/<pid>/mem and ptrace
```

and https://github.com/dotnet/diagnostics/blob/master/src/Tools/dotnet-dump/Dumper.Linux.cs#L19

mikem8361:
It has been on my backlog for a while and it will probably work. I need to try it out.

PR #307, opened 2019-11-12
Update GCRootDemo

NextTurn:
> Project '..\\..\Microsoft.Diagnostics.Runtime.Tests\Microsoft.Diagnostics.Runtime.Tests.csproj' targets 'netcoreapp2.0'. It cannot be referenced by a project that targets '.NETFramework,Version=v4.5'.


leculver:
Much appreciated!

PR #309, opened 2019-11-18
Updates related to collectible AssemblyLoadContext

NextTurn:
Contributes to dotnet/coreclr#20166

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
@@ -235,24 +235,36 @@ namespace Microsoft.Diagnostics.Runtime
             else
                 Debug.Assert(type == GetObjectType(obj));
 
-            if (type == null || !type.ContainsPointers)
+            if (type == null || (!type.ContainsPointers && !type.IsCollectible))
                 return s_emptyObjectSet;
 
-            GCDesc gcdesc = type.GCDesc;
-            if (gcdesc == null)
-                return s_emptyObjectSet;
+            List<ClrObject> result = null;
 
-            ulong size = type.GetSize(obj);
-            if (carefully)
+            if (type.ContainsPointers)
             {
-                ClrSegment seg = GetSegmentByAddress(obj);
-                if (seg == null || obj + size > seg.End || (!seg.IsLarge && size > 85000))
+                GCDesc gcdesc = type.GCDesc;
+                if (gcdesc == null)
                     return s_emptyObjectSet;
+
+                ulong size = type.GetSize(obj);
+                if (carefully)
+                {
+                    ClrSegment seg = GetSegmentByAddress(obj);
+                    if (seg == null || obj + size > seg.End || (!seg.IsLarge && size > 85000))
+                        return s_emptyObjectSet;
+                }
+
+                result = new List<ClrObject>();
+                MemoryReader reader = GetMemoryReaderForAddress(obj);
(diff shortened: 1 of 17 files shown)


leculver:
This is looking pretty good so far.  What's needed before this comes out of being a draft?  I'm looking to wrap up pull requests before forking the branch for ClrMD v2 some time this week.

NextTurn:
This is ready for review now.

src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
```diff
@@ -235,24 +235,36 @@ protected internal override IEnumerable<ClrObject> EnumerateObjectReferences(ulo
...
+                }
+
+                result = new List<ClrObject>();
+                MemoryReader reader = GetMemoryReaderForAddress(obj);
+                gcdesc.WalkObject(obj, size, ptr => ReadPointer(reader, ptr), (reference, offset) => result.Add(new ClrObject(reference, GetObjectType(reference))));
+            }
+
+            if (type.IsCollectible)
+            {
+                result ??= new List<ClrObject>(1);
```
leculver: Can you please change all instances of this pattern to `result = result ?? ...`?  While this is a valid C# feature, it apparently breaks Arcade (the build system used to produce signed builds).  I haven't tracked down that problem yet, but if left as-is I'll have to go fix these on a second checkin.

src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
```diff
@@ -263,24 +275,37 @@ protected internal override IEnumerable<ClrObjectReference> EnumerateObjectRefer
             else
                 Debug.Assert(type == GetObjectType(obj));
 
-            if (type == null || !type.ContainsPointers)
+            if (type == null || (!type.ContainsPointers && !type.IsCollectible))
```
leculver: Do collectible types not have references?  I'm confused as to why this check is here but I'm not very familiar with the feature.

NextTurn:
> Is there a test program you have laying around that would the collectible type features I could use to test this with?

I've no idea how to use *csc.exe* to produce .NET Core executables. Shall I add a *.csproj* to the test targets?

src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
NextTurn: There is no difference in `ContainsPointers`: it still returns `false` for unmanaged types.

src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
NextTurn: Can you enable CI on PRs so contributors can check for any build or test failures?

src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
weltkante: > Do collectible types not have references?

Its negated, collectible types can have references (the loader allocator) even if the type doesn't contain pointers

src/Microsoft.Diagnostics.Runtime/src/Common/HeapBase.cs
NextTurn: Reverting, because Arcade has been updated in #316 to use .NET Core SDK 3.0 with C# 8.0 available, so I believe this should build successfully.

leculver: I will look into it after this current sprint of work.  I would like to have it on as well, but I'm not sure the process there.

PR #318, opened 2019-11-20
Update handle root names

NextTurn:


leculver:
Someone might have taken a dependency on the old names.  I'm going to save this PR for when I fork for v2 early next week.

PR #325, opened 2019-11-21
Initial code cleanups for ClrMD 2.0

leculver:
Cleaned up hundreds of messages from FXCop.

Removed ICorDebug related code:  The CLR diagnostics team should own something like this.  I hope to have an example of connecting these two APIs together in the codebase, but not something officially supported through the API.

Removed PDB reader:  ClrMD will no longer read PDBs.  Other Nuget packages should be used.

Removed LockInspection code:  I'm not planning to support this in a side library.

Removed old dump reading code: Before ClrMD 2.0 ships I plan to pick up a dependency on a better dump reading code, or creating a new dump reader.


GregNagelDA:
@leculver If you have time to answer this - I'm quite curious to know what you didn't like about your dump reader code? I've been playing with my own minidump reader for a while ([https://github.com/GregTheDev/MinidumpExplorer)](https://github.com/GregTheDev/MinidumpExplorer) and yours seems better than mine :)

GregTheDev:
@leculver If you have time to answer this - I'm quite curious to know what you didn't like about your dump reader code? I've been playing with my own minidump reader for a while ([https://github.com/GregTheDev/MinidumpExplorer)](https://github.com/GregTheDev/MinidumpExplorer) and yours seems better than mine :)

leculver:
It's fine-ish.  =)

The code currently memmaps the entire dump into memory, so a large dump will either fail to load or cause OOM issues on x86.  I'm mostly trying to reduce the overall code in ClrMD.  I'd rather can improve on something like the [SymStore repo's MiniDump reader](https://github.com/dotnet/symstore/blob/master/src/Microsoft.FileFormats/Minidump) instead of owning everything myself.

If I can't make it work with symstore I may end up adding back the code and merging some of address/layout reading with the linux core dump reader.

PR #326, opened 2019-11-22
Update PEImage to use Span<T>

leculver:
- Added Buffers, Memory, and Unsafe nuget packages.
- Rewrote PEImage to use Span<T>

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Helpers/ExtensionMethods.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Helpers/ExtensionMethods.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    static class ExtensionMethods
+    {
+        public static int Read(this Stream stream, Span<byte> span)
+        {
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(span.Length);
+            try
+            {
+                int numRead = stream.Read(buffer, 0, span.Length);
+                new Span<byte>(buffer, 0, numRead).CopyTo(span);
+                return numRead;
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(buffer);
+            }
+        }
+    }
+}
--- a/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
@@ -1,10 +1,11 @@
 ﻿using Microsoft.Diagnostics.Runtime.Interop;
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
-using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
 using System.Text;
 
 namespace Microsoft.Diagnostics.Runtime.Utilities
@@ -63,20 +64,20 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
             _virt = isVirtual;
             Stream = stream;
 
-            ushort? dosHeaderMagic = TryRead<ushort>(0);
+            ushort dosHeaderMagic = Read<ushort>(0);
             if (dosHeaderMagic != ExpectedDosHeaderMagic)
             {
                 IsValid = false;
             }
             else
             {
-                _peHeaderOffset = TryRead<int>(PESignatureOffsetLocation) ?? 0;
-                uint? peSignature = null;
+                _peHeaderOffset = Read<int>(PESignatureOffsetLocation);
 
+                uint peSignature = 0;
                 if (_peHeaderOffset != 0)
-                    peSignature = TryRead<uint>(_peHeaderOffset);
+                    peSignature = Read<uint>(_peHeaderOffset);
 
-                IsValid = peSignature.HasValue && peSignature.Value == ExpectedPESignature;
+                IsValid = peSignature == ExpectedPESignature;
             }
 
             _imageFileHeader = new Lazy<ImageFileHeader>(ReadImageFileHeader);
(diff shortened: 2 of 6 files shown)


leculver:
@stephentoub Would you mind giving me a review of this change around how I use Span<T>?  This is my first time using it and I wanted a mini-review of one part of ClrMD before I changed the entire library to use it.

Please note that ClrMD needs to run on both Desktop CLR and .Net Core.

Specifically there are three major functions I'd like direct feedback on (almost everything else is just updates to use these functions):

1.  `private bool TryRead<T>(int offset, out T t) where T : unmanaged`
2.  `internal T Read<T>(ref int offset) where T : unmanaged`
3.  `public unsafe T GetData<T>(int offset = 0) where T : unmanaged`

Essentially I'm trying to make these functions as efficient as possible on .Net Core while making them work on Desktop CLR, hence the `#if CORE_ONLY` block, in `TryRead`, which is attempting to only do stack allocation if we are on .Net Core (because the desktop implementation of `FileStream.Read(Span)` doesn't exist...we would hit my `Stream.Read(Span)` extension method added in this pull request).

stephentoub:
Was about to review this... no longer necessary?

leculver:
I'm still looking for general feedback.  The change was merged, but I'm still actively working on Span support in clrmd.  So yes, please give feedback on this PR if you don't mind, it informs the bigger work going on without having you have to review the entire change.  =)

src/Microsoft.Diagnostics.Runtime/src/Helpers/ExtensionMethods.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    static class ExtensionMethods
+    {
+        public static int Read(this Stream stream, Span<byte> span)
+        {
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(span.Length);
+            try
+            {
+                int numRead = stream.Read(buffer, 0, span.Length);
+                new Span<byte>(buffer, 0, numRead).CopyTo(span);
```
stephentoub: Nit: Span => ReadOnlySpan
(Won't make any functional difference... just highlights that you're only reading from it)

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -315,71 +286,95 @@ private string ReadString(int offset, int len)
...
+                for (int i = 0; i < len; i++)
                 {
-                    len = i;
-                    break;
+                    if (buffer[i] == 0)
+                    {
+                        len = i;
+                        break;
+                    }
                 }
```
stephentoub: Nit:
```C#
int nullTerm = Array.IndexOf(buffer, 0);
if (nullTerm != -1)
    len = nullTerm;
```

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -315,71 +286,95 @@ private string ReadString(int offset, int len)
...
         {
-            int size = Marshal.SizeOf(typeof(T));
-            byte[] buffer = GetBuffer(size);
+            t = default;
+            int size = Unsafe.SizeOf<T>();
 
-            SeekTo(offset);
-            int result = Stream.Read(buffer, 0, size);
-            _offset = offset + result;
+            // .Net Core only.  This isn't compiled into Desktop CLR because Stream.Read(Span<byte>) doesn't
```
stephentoub: Nit: .Net => .NET

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -315,71 +286,95 @@ private string ReadString(int offset, int len)
...
+            // exist, and so our stackalloc + copy will be more inefficent than just renting our own byte array.
+#if CORE_ONLY
+            if (size < Configuration.MaxStackAlloc)
+            {
+                byte* ptr = stackalloc byte[size];
```
stephentoub: If you now or in the future disable zero'ing of locals (today using the ILLink feature for it, or in the near future using the Roslyn feature for it), then it'll be more efficient to instead do:
```C#
stackalloc byte[Configuration.MaxStackAlloc]
```
It's a bit cheaper for the JIT to stackalloc a const size than it is a dynamic size.

stephentoub: How about:
```C#
Span<byte> span = stackalloc byte[size];
```
?

Then below instead of:
```C#
Unsafe.Copy(ref t, ptr);
```
you can do:
```C#
t = MemoryMarshal.Read<T>(span);
```
?

NextTurn: There is a slight difference between these two methods.

How about:

```
t = Unsafe.As<byte, T>(ref span[0]);
```

?

NextTurn: There is a slight difference between these two methods.

How about:

```csharp
t = Unsafe.As<byte, T>(ref span[0]);
```

?

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
NextTurn: It may need to be

```csharp
int nullTerm = Array.IndexOf(buffer, 0, 0, len);
```

if the `len` doesn't consider the null terminator.

https://github.com/leculver/clrmd/blob/24aa33e7ee970be03051fcbc6b567cd399341191/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs#L266-L268

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
stephentoub: > How about

Why? MemoryMarshal.Read uses Unsafe.ReadUnaligned on coreclr (and just Unsafe.As, otherwise)... what's your concern with that?

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
stephentoub: > if the len doesn't consider the null terminator

It should go from 0 to Len, but I don't know what you mean about considering a null terminator. We want to search until we either hit a null or reach len.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
NextTurn: `Unsafe.ReadUnaligned` has an additional `unaligned` instruction.

https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il#L85

https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il#L313

NextTurn: `Unsafe.ReadUnaligned` contains an additional `unaligned` instruction.

https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il#L85

https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il#L313

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
NextTurn: `ArrayPool.Rent` may return a larger array than requested (`len`).

NextTurn: > if the `len` doesn't consider the null terminator.

That is... the null terminator is not included in `len` so not read by `Stream.Read`.

`ArrayPool.Rent` may return a larger array than requested (`len`).

NextTurn: > if the `len` doesn't consider the null terminator.

That is... the null terminator is not included in `len` thus not read by `Stream.Read`.

`ArrayPool.Rent` may return a larger array than requested (`len`).

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
stephentoub: Right. Why are you suggesting the one that doesn't have it?

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
stephentoub: > may return a larger array than requested (len)

Yes. Ok, I think we're just talking past each other.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
NextTurn: I assumed that `localloc` shall allocate aligned addresses. Not sure if it still holds for the `byte` target type.

PR #330, opened 2019-11-22
Span!

NextTurn:
Not only spans.

In this PR:

Whenever possible:

- Using `ReadOnlySpan`
- Using `sizeof`
- Avoid `fixed`
- Implicitly converting from array to `Span`

---

Note that any Read method other than `Unsafe.As` is unaligned.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/MemoryReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/MemoryReader.cs
@@ -5,10 +5,11 @@
 using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 
 namespace Microsoft.Diagnostics.Runtime
 {
-    internal unsafe class MemoryReader
+    internal class MemoryReader
     {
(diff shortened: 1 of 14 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/MemoryReader.cs
```diff
@@ -61,11 +62,11 @@ internal bool TryReadPtr(ulong addr, out ulong value)
             if (_currPageStart <= addr && addr - _currPageStart < (uint)_currPageSize)
             {
                 ulong offset = addr - _currPageStart;
-                fixed (byte* b = &_data[offset])
-                    if (IntPtr.Size == 4)
-                        value = *((uint*)b);
-                    else
-                        value = *((ulong*)b);
+                ref byte b = ref _data[offset];
```
NextTurn:
```suggestion
                value = Unsafe.As<byte, UIntPtr>(ref _date[offset]).ToUInt64();
```

Maybe using `UIntPtr` here?

leculver:
LGTM, thanks!

PR #332, opened 2019-11-22
Fix issue where we over-fed string reads

leculver:
ArrayPool.Rent was returning a larger string than we actually needed, and the code pattern in question would cause stray \0's to end up in the generated string.


leculver:
@NextTurn FYI, minor fix for your last PR.

weltkante:
Yes, this an important thing to keep in mind during code reviews, many people get it wrong initially when porting over to Span. I've come across that kind of bug in various projects already, array pooling is allowed to return larger arrays than requested (and commonly does so if one happens to be available). If that wasn't center of attention yet it may be worth to do a full review of all pooled array access.

PR #337, opened 2019-11-25
Implement SuspendAndAttachToProcess on Linux

NextTurn:
This test sample is dangerous: if the parent process crashed due to an assertion failure, the child process won't be killed.

```cs
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

using System.Diagnostics;
using System.Linq;
using Xunit;

namespace Microsoft.Diagnostics.Runtime.Tests
{
    public class DataTargetTests
    {
        [Fact]
        public void SuspendAndAttachToProcess()
        {
            using Process process = Process.Start(new ProcessStartInfo
            {
                FileName = "...",
                Arguments = "...",
                RedirectStandardOutput = true,
            });

            _ = process.StandardOutput.ReadLine();

            using (DataTarget dataTarget = DataTarget.SuspendAndAttachToProcess(process.Id))
            {
                ClrRuntime runtime = dataTarget.ClrVersions.Single().CreateRuntime();
                uint mainThreadId = runtime.GetMainThread().OSThreadId;
                ProcessThread mainThread = process.Threads.Cast<ProcessThread>().Single(thread => thread.Id == mainThreadId);

                Assert.Equal(ThreadState.Wait, mainThread.ThreadState);
            }

            process.Kill();
        }
    }
}
```

```cs
using System;

class Program
{
    private static void Main(string[] args)
    {
        if (args.Length == 0)
            throw new Exception();

        Console.WriteLine();
        for (; ; ) { }
    }
}
```

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -239,15 +240,19 @@ namespace Microsoft.Diagnostics.Runtime.Linux
 
             ref AMD64Context ctx = ref Unsafe.As<byte, AMD64Context>(ref MemoryMarshal.GetReference(context));
             ctx.ContextFlags = contextFlags;
-            IntPtr registerSet = Marshal.AllocHGlobal(sizeof(RegSetX64));
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(sizeof(RegSetX64));
             try
             {
-                ptrace(PTRACE_GETREGS, (int)threadID, IntPtr.Zero, registerSet);
-                CopyContext(ref ctx, Unsafe.AsRef<RegSetX64>(registerSet.ToPointer()));
+                fixed (byte* data = buffer)
+                {
+                    ptrace(PTRACE_GETREGS, (int)threadID, IntPtr.Zero, new IntPtr(data));
+                }
+
+                CopyContext(ref ctx, Unsafe.As<byte, RegSetX64>(ref MemoryMarshal.GetReference(buffer.AsSpan())));
             }
             finally
             {
-                Marshal.FreeHGlobal(registerSet);
+                ArrayPool<byte>.Shared.Return(buffer);
             }
 
             return true;
--- a/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
@@ -1,10 +1,12 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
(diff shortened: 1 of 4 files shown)


weltkante:
Can't you put "process.Kill" in a finally-block to avoid the risk of leaking the process?

NextTurn:
That doesn't help. `Debug.Fail`/`Environment.FailFast` doesn't throw so the `finally` block won't be executed.

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -195,48 +237,19 @@ public unsafe bool GetThreadContext(uint threadID, uint contextFlags, Span<byte>
...
-            {
-                return false;
-            }
-            IntPtr ptrContext = Marshal.AllocHGlobal(sizeof(AMD64Context));
-            AMD64Context* ctx = (AMD64Context*)ptrContext;
-            ctx->ContextFlags = contextFlags;
-            IntPtr ptr = Marshal.AllocHGlobal(sizeof(RegSetX64));
+            ref AMD64Context ctx = ref Unsafe.As<byte, AMD64Context>(ref MemoryMarshal.GetReference(context));
+            ctx.ContextFlags = contextFlags;
+            IntPtr registerSet = Marshal.AllocHGlobal(sizeof(RegSetX64));
```
leculver: Can you swap this to ArrayPool.Rent?  (I know it's in the old code, but might as well update it here.

weltkante:
> That doesn't help. Debug.Fail/Environment.FailFast doesn't throw so the finally block won't be executed.

`Environment.FailFast` isn't used in the library so this isn't an issue, `Debug.Fail` can easily be redirected by replacing the logger in `System.Diagnostics.Trace.Listeners` collection (and IMHO definitely should be done in any test project so a `Debug.Fail` can be properly associated with the currently running test, the .NET Core default behavior is counterproductive for test projects).

The simplest solution would be to register a listener which throws an exception, though if you have cases where you catch all exceptions and still want to detect Debug.Fail you'll have to employ a more sophisticated scheme.

NextTurn:
`Debug.Fail` actually calls `DebugProvider.Fail`, but `DebugProvider` isn't part of public APIs.

There is an `Action` delegate designed for tests but it is internal as well.

```cs
        // internal and not readonly so that the tests can swap this out.
        internal static Action<string, string?, string?, string>? s_FailCore = null;
```

weltkante:
> `Debug.Fail` actually calls `DebugProvider.Fail`, but `DebugProvider` isn't part of public APIs.

[Debug.SetProvider](https://source.dot.net/#System.Private.CoreLib/shared/System/Diagnostics/Debug.cs,20) seems plenty public API to replace it with a throwing version.

But you don't need that, .NET Core still [supports Trace.Listeners collection](https://source.dot.net/#System.Diagnostics.TraceSource/System/Diagnostics/TraceInternal.cs,66), its used in .NET Core WinForms to [suppress Debug.Assert failures](https://github.com/dotnet/winforms/blob/9cb6ba16178a315f179c92a25b1837d238288fdc/src/Common/tests/NoAssertContext.cs#L30) for specific tests.

NextTurn:
Thanks for pointing this out! I can add tests for these later.

I'd like to mention that although some APIs appear to be public, they are not shipped as public ones. https://apisof.net/catalog/System.Diagnostics.Debug

PR #343, opened 2019-11-26
Reimplement dbgeng interfaces

leculver:
Push dbgeng interfaces out of ClrMD.  Now maintain a small implementation of dbgeng interfaces that I use.


Microsoft.Diagnostics.Runtime.sln
```diff
@@ -9,6 +9,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.Runti
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.Runtime.UnitTests", "src\Microsoft.Diagnostics.Runtime.UnitTests\Microsoft.Diagnostics.Runtime.UnitTests.csproj", "{214A7EC8-4D32-4202-8D8B-1761743EE527}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.Runtime.Util", "src\Microsoft.Diagnostics.Util\Microsoft.Diagnostics.Runtime.Util.csproj", "{BFC539C3-6E60-42DF-B68E-23315B337CD2}"
```
NextTurn:
```suggestion
Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.Runtime.Utilities", "src\Microsoft.Diagnostics.Utilities\Microsoft.Diagnostics.Runtime.Utilities.csproj", "{BFC539C3-6E60-42DF-B68E-23315B337CD2}"
```

💡 Consider avoiding abbreviations.

Is it by design that the directory name and project name are different?

leculver: It takes me about 3x as long to figure out what I want to name something than it does write the code.  The utilities library now is just a place to park code, I am planning to actually name it and organize it in the next couple of days.  I'll likely name the library (and .csproj) `Microsoft.Diagnostics.Utilities` and the debugger wrappers `[assemblyname].DbgEng` since that's the name of the debugger.

NextTurn:
@leculver, `Marshal.GetDelegateForFunctionPointer` is the slowest interop pattern in .NET.

https://github.com/NextTurn/interop-benchmark/blob/master/results/result.svg

Line Chart

leculver:
I'm colorblind so the chart is really difficult for me to interpret, but I assume delegate is the largest bar there.  Do you mean it's the slowest to call to create the delegate?  Or do you mean it's the slowest on every call?

I don't think we should worry about Marshal.GetDelegateForFunctionPointer unless 1) the calls show up in perf traces to indicate it's a problem, and 2) there is a reasonable and maintainable alternative that doesn't rely on us generating code.

I would like to avoid LCG, Reflection.Emit, or generated IL because it makes things very difficult to debug problems and failures.  That would only be a last-resort for me.  I also expect (but haven't measured) that even if there's some slowdown calling through delegates that ClrMD is bound by IO for crash dumps, and bound by memory copy time in live processes.  I'm happy to revisit that later if perf traces reveal that assumption is wrong.

NextTurn:
Here are the numbers:

https://github.com/NextTurn/interop-benchmark/blob/master/results/result.md

> Do you mean it's the slowest to call to create the delegate? Or do you mean it's the slowest on every call?

It's slow on every call. A naive understanding is that the delegate is designed for managed methods and it needs to go through an extra indirection to call the wrapped native function.

>1\) the calls show up in perf traces to indicate it's a problem

It's rather unclear to me how to collect any useful perf traces. VS profiling is still broken... This is more of a reminder that there is still room for improvements if we prove it's a problem.

image

> and 2) there is a reasonable and maintainable alternative that doesn't rely on us generating code.

I'm open to code generation at compile time on call sites only to replace the currently used `Delegate.Invoke`. It won't do harm to debuggability.

weltkante:
You [may want](https://github.com/dotnet/runtime/issues/32963#issuecomment-592609261) to look into the [function pointer proposal](https://github.com/dotnet/csharplang/blob/master/proposals/function-pointers.md#nativecallableattribute). It sounds like it can replace `Marshal.GetDelegateForFunctionPointer` usage with something more performant, so unless you have pressing performance concerns which need to be addressed now its probably better to wait for .NET 5.0 providing new alternatives instead of investing a lot of work into a custom solution.

PR #345, opened 2019-11-26
Improve LinuxLiveDataReader

NextTurn:
---

Removes `this` to conform to current styles. Personally I would prefer to prefix everything with `this`.


leculver:
Looks good to me.  I haven't tested this yet and since it's touching (a little) of the gcroot code I'm about to move I'm going to go ahead and merge it.  Since it's going into the 'testing' branch it'll get tested later as a whole when I'm wrapping up this sprint of changes.  Thanks!

PR #351, opened 2019-11-30
Don't read from /proc/<pid>/mem

### LINKED ISSUE #264: SIGABRT on Linux when their is a ptrace() failure
Repro by the following code.
This will fail since the process doesn't have `SYS_PTRACE`, but the problem is that any error that is thrown there will cause an access violation and kill the process.
I believe that the issue is something with the way the code jumps between managed / native codes there that isn't ready to handle exceptions, but I'm not sure

```
try
{
    var dt = DataTarget.AttachToProcess(Process.GetCurrentProcess().Id, 1000, AttachFlag.Passive);
    var runtime = dt.ClrVersions[0].CreateRuntime();
    var st = runtime.Threads[0].EnumerateStackTrace().First();
    Console.WriteLine(st.DisplayString);
}
catch (Exception e)
{
    Console.WriteLine(e);
    Console.WriteLine("Error, died");
}
```

NextTurn:
Closes #264 

`process_vm_readv` is faster and doesn't raise exceptions or signals.

Benchmark results

|  Method |      Mean |     Error |    StdDev |
|-------- |----------:|----------:|----------:|
| ProcMem | 13.869 ms | 0.3094 ms | 0.9122 ms |
|   Readv |  6.353 ms | 0.1366 ms | 0.4029 ms |

Benchmark sample

```cs
using System;
using System.Buffers;
using System.Collections.Generic;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using Microsoft.Diagnostics.Runtime;

public class Program
{
    static void Main() => _ = BenchmarkRunner.Run<Program>();

    static readonly DataTarget dataTarget = DataTarget.PassiveAttachToProcess(1);
    static readonly IEnumerable<ModuleInfo> modules = dataTarget.EnumerateModules();

    [Benchmark]
    public void ReadMemory()
    {
        foreach (ModuleInfo module in modules)
        {
            byte[] buffer = ArrayPool<byte>.Shared.Rent((int)module.FileSize);
            _ = dataTarget.DataReader.ReadMemory(module.ImageBase, buffer.AsSpan(0, (int)module.FileSize), out int _);
            ArrayPool<byte>.Shared.Return(buffer);
        }
    }
}
```

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
@@ -28,8 +28,6 @@ namespace Microsoft.Diagnostics.Runtime.Linux
     internal class LinuxLiveDataReader : IDataReader
     {
         private List<MemoryMapEntry> _memoryMapEntries;
-        private FileStream _memoryStream;
-        private bool _initializedMemFile;
         private readonly List<uint> _threadIDs = new List<uint>();
 
         private bool _suspended;
@@ -74,10 +72,6 @@ namespace Microsoft.Diagnostics.Runtime.Linux
             if (_disposed)
                 return;
 
-            _memoryStream?.Dispose();
-            _memoryStream = null;
-            _initializedMemFile = false;
-
             if (_suspended)
             {
                 int status = (int)ptrace(PTRACE_DETACH, (int)ProcessId, IntPtr.Zero, IntPtr.Zero);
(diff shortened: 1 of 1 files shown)


leculver:
Wow huge time savings.  This looks good but I'm not going to merge anything until after I wrap up the 2.0 refactor changes.

@mikem8361 Interesting findings here about the difference between /proc//men and process_vm_readv.

leculver:
I'm happy to change over to purely using `process_vm_readv`, looks like a good improvement.  Is this PR ready?

NextTurn:
This is ready now.

Issue #315, opened 2019-11-19
UnauthorizedAccessException on Linux PassiveMode


weltkante:
As discussed on dotnet/corefx#42621 I was trying out attaching to a Linux process to dump a stack trace (as the in-proc methods from Desktop Framework are not available), but I'm having issues with security

```
Unhandled Exception: System.UnauthorizedAccessException: Access to the path '/proc/24461/mem' is denied. ---> System.IO.IOException: Permission denied
   --- End of inner exception stack trace ---
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at Microsoft.Diagnostics.Runtime.Linux.LinuxLiveDataReader.OpenMemFile()
   at Microsoft.Diagnostics.Runtime.Linux.LinuxLiveDataReader.ReadMemory(UInt64 address, IntPtr buffer, Int32 bytesRequested, Int32& bytesRead)
   at Microsoft.Diagnostics.Runtime.DacInterface.DacDataTargetWrapper.ReadVirtual(IntPtr self, UInt64 address, IntPtr buffer, Int32 bytesRequested, Int32& bytesRead)
   at Microsoft.Diagnostics.Runtime.DacLibrary..ctor(DataTarget dataTarget, String dacDll)
   at Microsoft.Diagnostics.Runtime.DataTarget.ConstructRuntime(ClrInfo clrInfo, String dac)
   at Program.Main(String[] args) in /home/xxx/private/build/guard/Program.cs:line 23
```

OS is Ubuntu 18.04.3 LTS 64 bit and I'm running both the process that is attaching and the one that is attached under the same user (but different shells). I'm not a Linux expert and searching for those dotnet-centric errors is not leading to any results of what to do to get proc/mem access to my own processes.

I'm aware that I'll probably have to give up on PassiveMode and do dumps to avoid torn stack traces, but considering there is no issue explaining how to resolve this security problem I thought I'd create one to document how to get PassiveMode working in case other people run into it.

/cc @leculver

leculver:
[...]

PR #354, opened 2019-12-01
Document granting PTRACE access via prctl PR_SET_PTRACER option

weltkante:
Document resolution of #315 and include options for self-inspection

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -10,7 +10,13 @@ Using ClrMD to inspect its own process is not supported and not recommended.  Th
 
 The fundamental problem here is data consistency.  The CLR runtime uses locks and other mechanisms to ensure that it sees a consistent view of the world (like any software project, really) and the CLR debugging layer (mscordaccore.dll) ignores those locks by design.  This means that if you are inspecting an unsuspended process you can get all kinds of weird behavior.  Exceptions, infinite loops, etc.
 
-We only support inspecting suspended processes, which obviously won't work with You are always required to suspend a live process before inspecting it, which obviously doens't work for your own process.  (Note that passing AttachFlags.Invasive and AttachFlags.NonInvasive suspend the process on your behalf, only AttachFlags.Passive does not.)
+We only support inspecting suspended processes. You are always required to suspend a live process before inspecting it, which obviously doens't work for your own process.  (Note that passing AttachFlags.Invasive and AttachFlags.NonInvasive suspend the process on your behalf, only AttachFlags.Passive does not.)
+
+Options for inspecting yourself are:
+- `CreateSnapshotAndAttach` on Windows.
+- Use [`dotnet-dump`](https://github.com/dotnet/diagnostics/blob/master/documentation/dotnet-dump-instructions.md) and inspect the dump.
+- Fork and `SuspendAndAttachToProcess` then kill the fork on Linux.
+- Start a new inspecting process and `SuspendAndAttachToProcess`.
 
 
 ## Does this work with any architecture? (x86/x64?)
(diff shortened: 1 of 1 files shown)


doc/FAQ.md
```diff
@@ -35,3 +35,13 @@ example, you can wrap your calls to the API into a seperate process, then use
...
 anything to the API to do this automatically though.
 
+## I am receiving UnauthorizedAccessException when attaching to a Linux process
+
+You need `PTRACE` capabilities (see `CAP_SYS_PTRACE`) and various associated
+security checks passing in order to attach to a Linux process. One option to
+pass the checks is running as root. If however you are just trying to avoid the
+problems mentioned above regarding inspecting your own process then you may want
+to start the inspecting process from the process to be inspected. In this case
+you have the option to explicitely allow the child process access by doing a
```
NextTurn: The ptracer is not limited to child processes: you can specify an arbitrary one.

weltkante: I've reformulated without assuming its a child process.

doc/FAQ.md
```diff
@@ -35,3 +35,13 @@ example, you can wrap your calls to the API into a seperate process, then use
 interprocess communication to relay the information. I do not plan on adding
 anything to the API to do this automatically though.
 
+## I am receiving UnauthorizedAccessException when attaching to a Linux process
```
NextTurn:
```suggestion
## I am receiving `UnauthorizedAccessException` when attaching to a Linux process
```

doc/FAQ.md
```diff
@@ -35,3 +35,12 @@ example, you can wrap your calls to the API into a seperate process, then use
 interprocess communication to relay the information. I do not plan on adding
 anything to the API to do this automatically though.
 
+## I am receiving `UnauthorizedAccessException` when attaching to a Linux process
+
+You need `PTRACE` capabilities (see `CAP_SYS_PTRACE`) and various associated
```
NextTurn:
```suggestion
You need `ptrace` capabilities to attach to the target process. One option to pass the access checks is running the inspecting process as root. If however you are just trying to inspect your own process, you may explicitly grant access to the inspecting process by performing a P/Invoke call to `prctl` with the `PR_SET_PTRACER` option.
```

Starting the inspecting process as a child is a reasonable (and recommended) way to exchange process ids, but I feel it should go into tutorials/samples instead.

weltkante: Hmm, not too happy with the suggested rewording

> [...] If however you are just trying to inspect your own process [...]

I mentioned "inspecting your own process" as a reference to the first FAQ entry, reducing that and removing the advice to start a child process makes this part look a bit out of place (I wouldn't understand what it means in the context of the rest of the reworded text).

As far as examples go, I thought about it but couldn't see a straightforward way to turn my test code into a useful example, sorry. I'm not used to code multiple processes in .NET Core.

NextTurn: Oh sorry I forgot the first FAQ entry. I'll have a look into this again.

NextTurn:
```markdown
You need `ptrace` capabilities to attach to the target process. The followings are different ways to grant access to the inspecting process:

- Run the inspecting process as root.
- Run the target process from the inspecting process.
- If you are trying to [inspect your own process](#can-i-use-this-api-to-inspect-my-own-process), you may perform a P/Invoke call to `prctl` with the inspecting process ID and the `PR_SET_PTRACER` option.
```

After some consideration I think that advice (starting a new inspecting process ~as a child~) is more suitable for the first section, so it looks like:

```markdown
You may...

- `CreateSnapshotAndAttach` on Windows.
- Fork and `SuspendAndAttachToProcess` then kill the fork on Linux.
- Start a new inspecting process and `SuspendAndAttachToProcess`.
```

I'm concerned about the difference between a **child** process and a **new** process. As I mentioned above a **child** process is sometimes treated in a special way (automatically granting access to its parent, etc.). I'm not sure which you are referring to exactly.

weltkante: I don't mind splitting the information, I've pushed an update.

> As I mentioned above a child process is sometimes treated in a special way (automatically granting access to its parent, etc.). I'm not sure which you are referring to exactly.

I don't know either, I only tested child processes since anything else would be too much work to test for me right now (have to get the PID into a running process). I don't mind getting rid of the term itself.

>  Fork and `SuspendAndAttachToProcess` then kill the fork on Linux.

Is that actually supported though? I don't know how `fork` works internally but I'd assume other threads continue running? That could be a problem for the .NET runtime if it doesn't know that it was forked and keeps running. The discussion on [another issue](https://github.com/microsoft/clrmd/issues/303#issuecomment-556508640) indicates its unclear on whether this will work, so I'm not sure if it should be included in the FAQ in this form.

> Run the target process from the inspecting process.

I don't know if thats enough to grant you ptrace access though, I didn't test it but the manpages sound like you need to do some `PTRACE_TRACEME` call after the `fork` but before the `execve`.

I don't mind leaving either of these in, just thought it would make sense to fact-check them first.

NextTurn: > Fork and `SuspendAndAttachToProcess` then kill the fork on Linux.

This needs further evaluation. I was just giving an example.

> Run the target process from the inspecting process.

This is tested. I'll add tests for it later so don't worry.

PR #356, opened 2019-12-01
Clean up styles

NextTurn:


weltkante:
Just FYI there should be no technical reason to do that, only a stylistic decision. Empty strings are treated specially by the runtime for a long time, the static `string.Empty` is just a historic oddity from before this time and at this point doesn't provide any actual value.

NextTurn:
Yes, this is purely stylistic. There are analyzers checking this.

Issue #311, opened 2019-11-18
ClrMD 2.0 feedback


leculver:
I am planning to ship a ClrMD 2.0, which will allow me to make significant changes to the library.  I have put the [ClrMD v2 roadmap here](https://github.com/microsoft/clrmd/blob/master/doc/ReleaseNotes2.0.md) that explains what the overall plan is.

Right now the document is a bit sparse because I am looking for feedback from the community before getting started.  Please use this issue to give any feedback, questions, concerns, criticisms, etc of the plan of record.

There are more details there, but at its core moving your project to ClrMD 2.0 will require code changes since we will be making significant breaking changes, a lot of cleanup, and refactoring most of the codebase.  With that said, the same basic surface area (with some changes) will exist, so it should be a reasonable amount of work to port from 1.0 to 2.0 in your projects, if you choose to do so.

For those who don't wish to migrate, I still plan to fix critical issues in ClrMD 1.0 for the foreseeable future.  However, I plan to _only_ do new feature work in the ClrMD 2.0 codebase, but I would be happy to accept pull requests from the community who want to backport features from 2.0 -> 1.0.  In theory, I'm hoping this is the best of both worlds for everyone.

What does everyone think?  Are there any concerns or requests you have for ClrMD v2?

leculver:
FYI to some folks who have submitted recent issues or prs:  @mitikov @0xd4d @isakiev @drewnoakes @loop-evgeny.

drewnoakes:
For API changes, #163 comes to mind.

leculver:
@drewnoakes  Yes, that kind of change is absolutely the kind of stuff I'll be doing in ClrMD 2.0.  One of my tasks for this week is to go back through all of the open issues and mark them as 1.0 or 2.0 changes.  That one we'd go fix in ClrMD 2.0.

mitikov:
The roadmap looks doable, 
**Tests**: as long as core is built on abstractions, it shall be testable.
I've never seen/managed to get CPU-bound code with ClrMD; I/O was always the bottleneck for me; putting in different words - performance was good; I'd prefer to trade off/attempt to runaway from struct (ClrOBj) to interface/class to ease mocking.

That would allow to get proper **unit** test coverage so that logic inside library is covered.
At the same time the current approach with memory snapshot collection for integration tests might stay eligible to verify runtime/(hosting environment) assumptions used inside core.

The part I missed a lot is a possibility to get the snapshot/attach/(execute code) on demand.
[...]

PR #358, opened 2019-12-01
Refactor codebase for 2.0

leculver:
Major changes for #311.  Overview of changes (some not complete yet as of this draft PR):

- Simplify type hierarchy.
- Decouple the internals of ClrMD from itself.  The internals of the old ClrMD were tightly coupled.  Now each implementation takes an interface used to construct that type.
- ClrType is no longer unified across all AppDomains.  There is one unique ClrType per ClrModule it comes from.  This simplifies a lot of internal code (such as not having to ask for AppDomain when getting static field values).
- ClrType is collectible and comparible...meaning we could construct multiple ClrType objects for the same method table that will compare as equal, but are not reference equal.  This means that when we create types, we no longer have to hold onto them forever...we can discard them for memory reasons and recreate later.
- ClrModule is now per-AppDomain and not unified.  The original design caused a lot of headaches throughout ClrMD that we avoid at the cost of the user having to deal with multiple ClrModules for the same physical file on disk.  (This also brings ClrMD in line with what CLR is actually _actually_ doing, which is the main goal here.)
- Allow for users to provide their own type caching and string interning.
- ClrRuntime will now be disposable and addref/release of DAC library/interfaces should be much simpler.

The most relevant set of changes is [V45Runtime.cs](https://github.com/microsoft/clrmd/blob/2e02c264b9d6f2fe57815ad598b605762a267a2b/src/Microsoft.Diagnostics.Runtime/src/Desktop/V45Runtime.cs), which will be cleaned up, refactored, and split into multiple files before merging.

FYI @NextTurn since you have a lot of PRs open.  I hope to complete this PR by Dec 5th.


NextTurn:
No worries, I'm ready for any merge conflicts.

leculver:
I am checking this in now to establish a baseline for other pull requests.  The work here is definitely not complete but it compiles and tests pass.  (Though I did strip out some of the unit tests that were failing...I don't really understand xunit theories well enough to be able to fix them.)

I am continuing the work in the testing/2.0 branch and will not merge this into master until I feel the library is ready to ship a prerelease build on nuget.

FYI, anything that looks weird or wrong in the library should either have a `// TODO` marker in the code, or a line item in the [TODO list on the roadmap](https://github.com/microsoft/clrmd/blob/master/doc/v2_roadmap.md).  If there's something that looks wrong that doesn't have a todo marker somewhere then it's likely a bug or oversight by me.  This was a massive change so there's definitely stuff I forgot.

Feel free to comment on this PR after its merged, open an issue, or otherwise reach out to let me know if anyone finds something amiss.

PR #347, opened 2019-11-27
Enable incremental build on Linux

NextTurn:

PR #375, opened 2019-12-05
Enable incremental build on Linux

NextTurn:
Continued from #347 

It seems that the CI build is configured to treat warnings as errors. MSBuild at least generates a warning for a crashed process, so I've added some tricks to ignore the crashes.

You may need to resolve some unreachable code warnings before this can build well.


leculver:
Moving to a temporary side branch to use the "official" build system to make sure this builds properly in their weird environment.

PR #379, opened 2019-12-06
Annotate for nullable reference types

NextTurn:
Left about 130 warnings for you to consider, where null-forgiving is involved. :)

Feel free to push my branch for any additional changes.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
@@ -24,12 +24,12 @@ namespace Microsoft.Diagnostics.Runtime.Builders
         private readonly DacLibrary _library;
         private readonly ClrDataProcess _dac;
         private readonly SOSDac _sos;
-        private readonly SOSDac6 _sos6;
+        private readonly SOSDac6? _sos6;
         private readonly int _threads;
         private readonly ulong _finalizer;
         private readonly ulong _firstThread;
 
-        private volatile ClrType[]? _basicTypes;
+        private volatile ClrType?[]? _basicTypes;
         private readonly Dictionary<ulong, ClrAppDomain> _domains = new Dictionary<ulong, ClrAppDomain>();
         private readonly Dictionary<ulong, ClrModule> _modules = new Dictionary<ulong, ClrModule>();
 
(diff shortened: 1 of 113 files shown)


leculver:
Awesome work.  I'll work through these remaining warnings today.

src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
```diff
@@ -345,7 +349,7 @@ IReadOnlyList<ClrAppDomain> IRuntimeHelpers.GetAppDomains(ClrRuntime runtime, ou
             int i = 0;
             foreach (ulong domain in domainList)
             {
-                ClrAppDomain? ad = GetOrCreateAppDomain(builder, domain);
+                ClrAppDomain ad = GetOrCreateAppDomain(builder, domain);
                 if (ad != null)
```
NextTurn: No need to check for a non-nullable value.

src/Microsoft.Diagnostics.Runtime/src/Utilities/Symbols/DefaultSymbolLocator.Async.cs
```diff
@@ -125,18 +125,17 @@ private async Task<string> SearchSymbolServerForFile(string fileSimpleName, stri
...
             }
 
-            string result = await GetFirstNonNullResult(tasks).ConfigureAwait(false);
+            string? result = await GetFirstNonNullResult(tasks).ConfigureAwait(false);
             return result;
         }
 
-        private static async Task<T> GetFirstNonNullResult<T>(List<Task<T>> tasks)
-            where T : class
+        private static async Task<string> GetFirstNonNullResult(List<Task<string?>> tasks)
```
NextTurn:
```suggestion
        private static async Task<string?> GetFirstNonNullResult(List<Task<string?>> tasks)
```

There are codes checking against null for the call chain:

```
FindBinaryAsync
DownloadFileWorker
SearchSymbolServerForFile
GetFirstNonNullResult
```

https://github.com/microsoft/clrmd/blob/19725d98eae81948c7c9dbbfc85cdd65217cc636/src/Microsoft.Diagnostics.Runtime/src/Utilities/Symbols/DefaultSymbolLocator.Async.cs#L54

leculver: Thanks!

I am going to leave this for later.  Going through the symbol locator code is on my todo list so I'm hesitant to do more cleanups there until I know what the final result will look like.

PR #380, opened 2019-12-06
Enable incremental build on Linux (#375)

leculver:
From @NextTurn.

* Enable incremental build on Linux

* Suppress MSBuild warnings

* Use SDK host to execute

* Fix build warnings


leculver:
@NextTurn I got close to getting this compiling.  See the failing checks above.  Looks like the Windows build is failing because the CSC being used is the wrong architecture.  Do you have any ideas on fixing that one?  I haven't dug into what's the right thing to do here.  Maybe it's a simple as choosing the right csc.exe based on platform.

If you have time to resolve the last build error here that'd be great.  If not I'll probably get back around to this late next week.

Ignoring that for a moment, it looks like the underlying intent here is to run the full tests on checkin.  Unfortunately I think even if the last build error is fixed, we are going to run into trouble here since I don't think the Windows build machines that produce official builds (like the checks above) payload a copy of Debugging Tools for Windows.  I can write a mini-debugger that will generate the dumps for us.  That's probably nicer than requiring debugging tools for windows installed anyway.  I'll tackle that sometime later next week.

leculver:
Actually it looks like the debugger is launched just fine, just need to fix the last compile issue with the architecture of csc mismatch.  I can't look at that this week.

NextTurn:
> it looks like the underlying intent here is to run the full tests on checkin.

Ah, of course.

> That's probably nicer than requiring debugging tools for windows installed anyway.

Agree. Running tests won't require the debugging tools if we bring back [the original debugger](https://github.com/microsoft/clrmd/blob/57e5cf053f4d4d9c3056d377967881eb71a15a9d/src/Microsoft.Diagnostics.Runtime.Tests/Debugger.cs).

leculver:
Good point.  I added bringing back the old debuggers to my [todo list](https://github.com/microsoft/clrmd/blob/master/doc/v2_roadmap.md).

NextTurn:
@leculver You forgot to delete the temporary branch.

PR #396, opened 2019-12-09
Stop using Nullable<T>.Value

NextTurn:
See dotnet/coreclr# 27575 for backgrounds.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs
@@ -94,15 +94,15 @@ namespace Microsoft.Diagnostics.Runtime
         {
             get
             {
-                if (_minidump.HasValue)
-                    return !_minidump.Value;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs
```diff
@@ -114,36 +114,37 @@ public Architecture Architecture
...
-                    return _architecture.Value;
+                if (_architecture is Architecture architecture)
+                    return architecture;
 
                 IMAGE_FILE_MACHINE machineType = _control.GetEffectiveProcessorType();
                 switch (machineType)
                 {
                     case IMAGE_FILE_MACHINE.I386:
-                        _architecture = Architecture.X86;
+                        architecture = Architecture.X86;
```
leculver: Should be _architecture.

NextTurn: The local variable is assigned to the field after several lines.

Issue #398, opened 2019-12-09
Safe replacement for PEFile.GetFileVersionInfo?


arvindshmicrosoft:
Hello! Wondering if there are any thoughts or prior work on safely replacing the older PEFile.GetFileVersionInfo with an equivalent? That function was very helpful in abstracting the details on parsing the version info resource entries, and it would be great if we can have it back in some form.

leculver:
Ah thanks for the issue.  I forgot this wasn't part of the new `PEImage` class.

I added this to PEImage in ClrMD 2.0 as PR #399.  I don't plan to do any feature work in ClrMD 1.1 but you are welcome to submit a pull request if you want to backport the feature yourself.

Thanks!
-Lee

PR #399, opened 2019-12-09
Add FileVersionInfo to PEImage

leculver:


leculver:
Fixes https://github.com/microsoft/clrmd/issues/398.

PR #404, opened 2019-12-10
Clean up unsafe codes

NextTurn:


src/Microsoft.Diagnostics.Runtime/src/Linux/Structs/RegSetArm.cs
```diff
@@ -29,33 +30,31 @@ internal struct RegSetArm
...
-                ctx->R9 = r9;
-                ctx->R10 = r10;
-                ctx->R11 = fp;
-                ctx->R12 = ip;
-                ctx->Sp = sp;
-                ctx->Lr = lr;
-                ctx->Pc = pc;
-                ctx->Cpsr = cpsr;
-            }
+            ref ArmContext contextRef = ref Unsafe.As<byte, ArmContext>(ref MemoryMarshal.GetReference(context));
```
leculver: Does the changes to CopyContext _actually_ improve memory safety at all?  Using `Unsafe.As` here makes it so that we do not have to use an `unsafe` marker, but it doesn't actually prevent us from reading past the bounds of the Span any more than `if (context.Length < ArmContext.Size)` already does.

I think there's value in pointing out that the code in CopyContext fundamentally _is_ unsafe.  I might change my mind here, so let me think about this for a bit.

NextTurn: Both the `unsafe` keyword and the `Unsafe` class are unsafe in the end. I should have named it "clean up `fixed` statements".

leculver: Indeed.

I chatted with my coworker and we both agree your change is for the better here.  At minimum it would theoretically let the JIT unpin memory, though in practice I don't think this region would ever be jitted as interruptible.

NextTurn: >  At minimum it would theoretically let the JIT unpin memory

That's exactly why I prefer references to pointers (although I often can't tell any difference in benchmarks).

PR #405, opened 2019-12-10
Throw ArgumentException if the process is not running

### LINKED ISSUE #281: Inconsistent and unhelpful error when process does not exist
When calling `DataTarget.AttachToProcess` with a PID that does not exist the exception thrown is significantly different from Windows to Linux and the message is not very helpful.

Linux:

```
System.IO.DirectoryNotFoundException: Could not find a part of the path '/proc/15351/maps'.
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at Microsoft.Diagnostics.Runtime.Linux.LinuxLiveDataReader.LoadMemoryMap()
   at Microsoft.Diagnostics.Runtime.Linux.LinuxLiveDataReader..ctor(UInt32 processId)
   at Microsoft.Diagnostics.Runtime.DataTarget.AttachToProcess(Int32 pid, UInt32 msecTimeout, AttachFlag attachFlag)
```

Windows:

```
Microsoft.Diagnostics.Runtime.ClrDiagnosticsException: Could not attach to process. Error 0.
   at Microsoft.Diagnostics.Runtime.LiveDataReader..ctor(Int32 pid, Boolean createSnapshot) in /_/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Live/LiveDataReader.cs:line 51
   at Microsoft.Diagnostics.Runtime.DataTarget.AttachToProcess(Int32 pid, UInt32 msecTimeout, AttachFlag attachFlag) in /_/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTar
get.cs:line 140
```

It would be nice to have something like a ProcessNotFoundException for this, specifying the process ID. (The DirectoryNotFoundException on Linux could be an inner exception to this - it's still kind of nice to see that detail.)

NextTurn:
Fixes #281

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Live/LiveDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Live/LiveDataReader.cs
@@ -9,6 +9,7 @@ using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
+using Microsoft.Diagnostics.Runtime.Utilities;
 
 namespace Microsoft.Diagnostics.Runtime
 {
@@ -29,6 +30,8 @@ namespace Microsoft.Diagnostics.Runtime
             if (createSnapshot)
             {
                 _originalPid = pid;
+
+                // Throws
                 Process process = Process.GetProcessById(pid);
                 int hr = PssCaptureSnapshot(process.Handle, PSS_CAPTURE_FLAGS.PSS_CAPTURE_VA_CLONE, IntPtr.Size == 8 ? 0x0010001F : 0x0001003F, out _snapshotHandle);
                 if (hr != 0)
@@ -49,6 +52,9 @@ namespace Microsoft.Diagnostics.Runtime
 
             if (_process == IntPtr.Zero)
             {
+                if (!WindowsFunctions.IsProcessRunning(_pid))
+                    throw new ArgumentException("The process is not running");
+
                 int hr = Marshal.GetLastWin32Error();
                 throw new ClrDiagnosticsException($"Could not attach to process. Error {hr}.", ClrDiagnosticsExceptionKind.Unknown, hr);
             }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataReaders/Live/LiveDataReader.cs
```diff
@@ -26,10 +26,12 @@ internal unsafe class LiveDataReader : IDataReader
 
         public LiveDataReader(int pid, bool createSnapshot)
         {
+            // Throws
+            using Process process = Process.GetProcessById(pid);
```
leculver: I don't think this is a good solution here.

This works 99.9% of the time, but it introduces a race condition if the process dies between the creation of this object and when we call PssCaptureSnapshot.  As a minor point it also requires us to go through whatever machinery that `Process.GetProcessById` requires and to collect the object.  That issue has to be fixed by properly checking HRESULTs/error codes and throwing a more appropriate exception.

NextTurn:
FYI on Windows we'll get an `ERROR_INVALID_PARAMETER` in `PassiveAttachToProcess` and -1 in `SuspendAndAttachToProcess`.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Live/LiveDataReader.cs
NextTurn: Did you mean `ArgumentException` isn't appropriate?

leculver: Sorry that was unclear.  Argument exception is good, I was only worried about using GetProcessById here.

PR #349, opened 2019-11-29
Implement GetFileVersion/GetVersionInfo on Linux

NextTurn:
Fixes #148 
Fixes #306 

`GetFileVersion` works for any *.so* file.

## `CoreDumpReader.GetVersionInfo` in mini dump

```
/usr/share/dotnet/shared/Microsoft.NETCore.App/3.0.0/System.Native.so
v0.0.0.00
/usr/share/dotnet/shared/Microsoft.NETCore.App/3.0.0/libclrjit.so
v0.0.0.00
/usr/share/dotnet/shared/Microsoft.NETCore.App/3.0.0/libcoreclr.so
v4.700.19.46205
/usr/share/dotnet/shared/Microsoft.NETCore.App/3.0.0/libhostpolicy.so
v3.0.19.46706
/usr/share/dotnet/host/fxr/3.0.0/libhostfxr.so
v3.0.19.46706
```

## `LinuxLiveDataReader.GetVersionInfo`

```
/usr/share/dotnet/shared/Microsoft.NETCore.App/3.0.0/System.Native.so
v4.700.19.46214
/usr/share/dotnet/shared/Microsoft.NETCore.App/3.0.0/libclrjit.so
v4.700.19.46205
/usr/share/dotnet/shared/Microsoft.NETCore.App/3.0.0/libcoreclr.so
v4.700.19.46205
/usr/share/dotnet/shared/Microsoft.NETCore.App/3.0.0/libhostpolicy.so
v3.0.19.46706
/usr/share/dotnet/host/fxr/3.0.0/libhostfxr.so
v3.0.19.46706
```

PR #410, opened 2019-12-11
Various fixes

NextTurn:


NextTurn:
```
Total tests: 107
     Passed: 63
     Failed: 23
    Skipped: 21
 Total time: 29.0333 Seconds
```

leculver:
Looks great so far.  Thanks for tackling this.

NextTurn:
Only `HeapTests.ServerSegmentTests` and `PdbTests.PdbEqualityTest` fail now. Can't help with these two.

NextTurn:
Only `PdbTests.PdbEqualityTest` fails now.

NextTurn:
Enabling PR status checks should be as simple as installing [Azure Pipelines](https://github.com/marketplace/azure-pipelines) and granting it access to the repo. We'll soon be ready for full CI testing.

src/Microsoft.Diagnostics.Runtime.Tests/src/TestTargets.cs
```diff
@@ -90,12 +90,18 @@ public TestTarget(string name)
 
         private static DataTarget LoadDump(string path)
         {
-            DataTarget dt = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? DataTarget.LoadCrashDump(path) : DataTarget.LoadCoreDump(path);
-            dt.BinaryLocator = new SymbolServerLocator("");
-            return dt;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
```
leculver: Maybe I should remove "LoadCoreDump/LoadCrashDump" and just publish "LoadDump" and have the right reader selected based on either the operating system or the header of the file itself.  (The latter gets complicated because DbgEngDataReader allows for .zip and .cab files too, which is useful for internal Microsoft developers.)

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
```diff
@@ -150,39 +151,35 @@ public bool Read(ulong address, Span<byte> buffer, out int bytesRead)
...
 
-        public unsafe bool Read<T>(ulong addr, out T value) where T : unmanaged
+        public unsafe bool Read<T>(ulong address, out T value) where T : unmanaged
         {
             Span<byte> buffer = stackalloc byte[sizeof(T)];
-            if (!Read(addr, buffer, out _))
+            if (Read(address, buffer, out int size) && size == sizeof(T))
             {
-                value = Unsafe.As<byte, T>(ref buffer[0]);
+                value = Unsafe.As<byte, T>(ref MemoryMarshal.GetReference(buffer));
```
leculver: Ah right.  Thanks for these cleanups.  I'm still getting used to Unsafe.As and MemoryMarshal after being stuck on an older version of Desktop .Net for years.

src/Microsoft.Diagnostics.Runtime.Tests/src/TestTargets.cs
leculver: Any thoughts?

NextTurn: In fact, some functionalities are still available for `LoadCoreDump` on Windows, which don't request information from DAC.

leculver: Good point.  I'll leave this as-is for now.

NextTurn: `LoadCoreDump` behaved well when I was working on #349 by experimenting with a coredump on Windows, but it won't work now because it throws after #393. Maybe we can throw that exception later when DAC is required.

PR #412, opened 2019-12-12
Implement FileVersionInfo for PEImage

arvindshmicrosoft:
Addresses #398 in 1.1


src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
```diff
@@ -30,4 +30,10 @@
       <PackagePath></PackagePath>
     </Content>
   </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Buffers" Version="4.5.0" />
```
leculver: Unfortunately ClrMD 1.1 has a strict requirement that it doesn't carry dependencies.  These libraries have to be removed.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/FileVersionInfo.cs
```diff
@@ -2,47 +2,50 @@
...
+        public string? FileVersion { get; }
 
         /// <summary>
         /// Comments to supplement the file version
         /// </summary>
-        public string Comments { get; }
+        public string? Comments { get; }
 
-        internal FileVersionInfo(byte* data, int dataLen)
+        internal FileVersionInfo(Span<byte> data)
```
leculver: Since there's no Span support without System.Memory, you could just implement this with a byte* like the old version.

NextTurn:
Do you mean ClrMD 2.0 can carry dependencies?

src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
arvindshmicrosoft: Thanks for pointing this out - I have addressed the feedback.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/FileVersionInfo.cs
arvindshmicrosoft: Done!

leculver:
This version looks good!  I'll ship a release sometime next week unless you need it today.

PR #418, opened 2019-12-12
Test pull request for CI builds

leculver:
Test pull request to see if we correctly enabled CI builds for pulls from other forks.


leculver:
@NextTurn Looks like CI builds are now on for the repo.  I still need to create a debugger to use other than cdb, so I'll probably work on that next.  (Right now I'm working on sign extension issues.  I'm trying to file issues for work now that most of it is under control and assign the issue to myself as I start work on it).

Issue #402, opened 2019-12-10
Unable to analyze 4GB dumps of 32-bit processes


swift-kim:
It seems ClrMD is not able to process 32-bit memory dumps larger than 2GB due to inconsistent address conversion between coreclr and ClrMD.

As @leculver and @mikem8361 already discussed in #280, using `ulong` types for address values may lead to unexpected behavior on particular platforms.

For your information, I am developing for Tizen wearable devices which run 32-bit (armel) .NET processes on 64-bit Linux kernels. Unlike normal 32-bit processes on other platforms, each user process has 4GB of address space.

As far as I can see in the DAC implementation of coreclr, most of exposed address values are in `CLRDATA_ADDRESS` types, so we can safely assume them as sign-extended according to [this comment](https://github.com/dotnet/runtime/blob/20ab2fa963934b5823e0b8afbba858169dd27169/src/coreclr/src/debug/daccess/dacimpl.h#L35-L111). When the values are passed to ClrMD however, they are (implicitly) converted into `ulong` type (instead of `long`) which seems not correct for negative `CLRDATA_ADDRESS`'es.

For example, when the DAC calls into ClrMD using `DataTargetAdapter::ReadVirtual()` and the value of `address` is larger than 0x7FFFFFFF (let's say 0xFFCCBBAA), the `ulong` value passed to `DacDataTargetWrapper` appears to be super larger than expected (0xFFFFFFFFFFCCBBAA).

https://github.com/dotnet/runtime/blob/master/src/coreclr/src/debug/daccess/datatargetadapter.cpp#L186-L196

Is this conversion intentional (should I assume sign-extension for `ulong`)? Or am I missing something? 

In my experiment, my sample code (just printing a managed stakctrace) simply worked by modifying the implementation of `TO_CDADDR` macro in the coreclr runtime. I don't think this would be a right choice however. Changing all occurrences of `ulong` in the DAC interfaces of ClrMD to `long` also looks bad since we will need a lot of work to be done and additional maths in the code.

leculver:
Thanks for bringing this to my attention.  I'm in the process of making ClrMD 2.0 and have the opportunity to make breaking changes as needed, but it's not entirely clear what should be done to fix it.  I'll try to make a test this week and see what work needs to be done, but it's a very difficult scenario to test.

> Is this conversion intentional (should I assume sign-extension for ulong)? Or am I missing something? 

[...]

PR #419, opened 2019-12-12
Fix sign extension issues

leculver:
CLRDATA_ADDRESS is defined as a signed 64 bit integer in the CLR codebase.  When inspecting a 32bit process, this means that we cannot simply use ulong to represent these values when communicating with the dac because addresses with the highest bit set will be sign extended by the dac.

This change adds a ClrDataAddress struct to represent a possibly sign-extended data value.  All structs that the dac produces have been updated to match their definition in coreclr/src/inc/dacprivate.h.

This struct will do the right thing to un-sign extend the address when inspecting a 32bit process.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
@@ -126,13 +126,13 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             return uint.MaxValue;
         }
 
-        public ulong GetThreadFromThinlockId(uint id)
+        public ClrDataAddress GetThreadFromThinlockId(uint id)
         {
             InitDelegate(ref _getThreadFromThinlockId, VTable->GetThreadFromThinlockID);
-            if (_getThreadFromThinlockId(Self, id, out ulong thread) == S_OK)
+            if (_getThreadFromThinlockId(Self, id, out ClrDataAddress thread) == S_OK)
                 return thread;
 
-            return 0;
+            return default;
         }
 
         public string? GetMethodDescName(ulong md)
@@ -282,22 +282,22 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             return hr == S_OK;
         }
 
-        public ulong GetMethodDescPtrFromFrame(ulong frame)
+        public ClrDataAddress GetMethodDescPtrFromFrame(ulong frame)
         {
             InitDelegate(ref _getMethodDescPtrFromFrame, VTable->GetMethodDescPtrFromFrame);
-            if (_getMethodDescPtrFromFrame(Self, frame, out ulong data) == S_OK)
+            if (_getMethodDescPtrFromFrame(Self, frame, out ClrDataAddress data) == S_OK)
                 return data;
 
-            return 0;
+            return default;
         }
 
-        public ulong GetMethodDescPtrFromIP(ulong frame)
+        public ClrDataAddress GetMethodDescPtrFromIP(ulong frame)
         {
             InitDelegate(ref _getMethodDescPtrFromIP, VTable->GetMethodDescPtrFromIP);
-            if (_getMethodDescPtrFromIP(Self, frame, out ulong data) == S_OK)
+            if (_getMethodDescPtrFromIP(Self, frame, out ClrDataAddress data) == S_OK)
                 return data;
 
-            return 0;
+            return default;
         }
 
         public string GetFrameName(ulong vtable)
@@ -368,12 +368,12 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             return _getCommonMethodTables(Self, out commonMTs) == S_OK;
         }
 
-        public ulong[] GetAssemblyList(ulong appDomain)
+        public ClrDataAddress[] GetAssemblyList(ulong appDomain)
         {
             return GetAssemblyList(appDomain, 0);
         }
 
-        public ulong[] GetAssemblyList(ulong appDomain, int count)
+        public ClrDataAddress[] GetAssemblyList(ulong appDomain, int count)
         {
             return GetModuleOrAssembly(appDomain, count, ref _getAssemblyList, VTable->GetAssemblyList);
         }
@@ -502,13 +502,13 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             }
         }
 
-        public ulong GetMethodTableByEEClass(ulong eeclass)
+        public ClrDataAddress GetMethodTableByEEClass(ulong eeclass)
         {
             InitDelegate(ref _getMTForEEClass, VTable->GetMethodTableForEEClass);
-            if (_getMTForEEClass(Self, eeclass, out ulong data) == S_OK)
+            if (_getMTForEEClass(Self, eeclass, out ClrDataAddress data) == S_OK)
                 return data;
 
-            return 0;
+            return default;
         }
 
         public bool GetModuleData(ulong module, out ModuleData data)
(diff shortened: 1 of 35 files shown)


leculver:
Potential fix for #402.

swift-kim:
Unfortunately I get this error when creating a runtime from a coredump.
```
Unhandled exception. System.IO.InvalidDataException: Could not create well known type 'Free' from MethodTable 2b084c0.
   at Microsoft.Diagnostics.Runtime.Builders.RuntimeBuilder.CreateSystemType(ClrHeap heap, UInt64 mt, String kind) in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\Builders\RuntimeBuilder.cs:line 792
   at Microsoft.Diagnostics.Runtime.Implementation.ClrmdHeap..ctor(ClrRuntime runtime, IHeapBuilder heapBuilder) in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\Implementation\ClrmdHeap.cs:line 166
   at Microsoft.Diagnostics.Runtime.Builders.RuntimeBuilder..ctor(ClrInfo clr, DacLibrary library, SOSDac sos) in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\Builders\RuntimeBuilder.cs:line 91
   at Microsoft.Diagnostics.Runtime.ClrInfo.ConstructRuntime(String dac) in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\DataTargets\ClrInfo.cs:line 111
   at Microsoft.Diagnostics.Runtime.ClrInfo.CreateRuntime() in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\DataTargets\ClrInfo.cs:line 97
   at Program.Main(String[] args) in E:\dummy\ClrMDSample\Program.cs:line 13
Aborted (core dumped)
```

With the following line commented out and #421 applied, my [sample](https://github.com/swift-kim/ClrMDSample/blob/master/Program.cs) program runs without error.
https://github.com/microsoft/clrmd/blob/09279f4fc01702009ed9a8e1ad681a57360d3199/src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdHeap.cs#L166

The only problem is that I can see a mixed usage of sign-extended and 0-extended addresses, and some special frame names (such as `[InlinedCallFrame]` and `[GCFrame]`) are missing from the stack trace as follows.
```
Thread 7C71:
FFFFFFFFFFE554BC     F72AF038
FFFFFFFFFFE554BC     EEB0C5F9
    FFE554B8     EEB0C5F9
    FFE55568     EEB0C18F Tizen.Applications.CoreBackend.UICoreBackend.Run(System.String[])
    FFE55590     ED8C3673 Tizen.Applications.CoreApplication.Run(System.String[])
    FFE555C8     EEB0BCEF Tizen.Applications.CoreUIApplication.Run(System.String[])
    FFE555E8     EB6AD7EB Alarm.Program.Main(System.String[])
FFFFFFFFFFE557F4     F328ABD7
FFFFFFFFFFE55B88     F328ABD7
```

I will try with a full unit test when I have time.

leculver:
Ah I missed a few places I needed to use ClrDataAddress, which I think is the problem you are seeing with frame names/unsigned values.  I pushed another commit and will merge this pull request, but I'm leaving the issue open until I have a chance to debug this on an actual device next week.  Mike gave me access to some offline.

The issue creating Free is very strange.  I'll have to check that next week.  In the meantime, could you check and tell me if this code prints "Succeeded" or something else?

    using ClrRuntime runtime = dataTarget.ClrVersions.Single().CreateRuntime();
    using var dac = runtime.DacLibrary.SOSDacInterface;
    if (!dac.GetCommonMethodTables(out CommonMethodTables mts))
        Console.WriteLine("Failed to get common method tables");
    else if (!dac.GetMethodTableData(mts.FreeMethodTable, out MethodTableData data))
        Console.WriteLine("Failed to get Free MethodTableData");
    else
        Console.WriteLine("Succeeded");

Thanks for your help here!

leculver:
I'm happy with this change, so I'm going to merge it.  Still leaving open issue #402 until I've had time to test this on an actual arm64 device next week.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
```diff
@@ -726,7 +726,7 @@ public ulong GetMethodDescFromToken(ulong module, uint token)
         private delegate int DacGetIntPtr(IntPtr self, out IntPtr data);
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
-        private delegate int DacGetUlongWithArg(IntPtr self, ulong arg, out ulong data);
+        private delegate int DacGetUlongWithArg(IntPtr self, ulong arg, out ClrDataAddress data);
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         private delegate int DacGetUlongWithArgs(IntPtr self, ulong arg, uint id, out ulong data);
```
swift-kim: Shouldn't this `data` be `ClrDataAddress` as well?

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
```diff
@@ -726,7 +726,7 @@ public ulong GetMethodDescFromToken(ulong module, uint token)
         private delegate int DacGetIntPtr(IntPtr self, out IntPtr data);
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
-        private delegate int DacGetUlongWithArg(IntPtr self, ulong arg, out ulong data);
+        private delegate int DacGetUlongWithArg(IntPtr self, ulong arg, out ClrDataAddress data);
```
swift-kim: Are you fine with the delegate name(s) which now seems to be a bit misleading?

Also, I wonder if it is your intent to leave args passed to the Dac as `ulong`. (I see no problem in practice.)

swift-kim:
```csharp
using ClrRuntime runtime = dataTarget.ClrVersions.Single().CreateRuntime();
```
Replacing the above line (which causes `ImmutableArray<ClrInfo> does not contain a definition for 'Single'`) with
```csharp
using ClrRuntime runtime = dataTarget.ClrVersions[0].CreateRuntime();
```
the result is
```
Failed to get Free MethodTableData
```
Maybe the error is specific to my machine. I just don't know what the Free methodtable is for and the only thing I've figured out with lldb is pEEClass is not NULL in this context.

https://github.com/dotnet/runtime/blob/4359ebfc9943608b34f411366b1e544ac45702b7/src/coreclr/src/debug/daccess/request.cpp#L132-L133
```cpp
BOOL DacValidateMethodTable(MethodTable *pMT, BOOL &bIsFree)
{
        ...
        EEClass *pEEClass = pMT->GetClass();
        if (pEEClass==NULL)
        {
            // Okay to have a NULL EEClass if this is a free methodtable
            ...
        }
        else
        {
            // Standard fast check
            if (!pMT->ValidateWithPossibleAV())
                goto BadMethodTable; // <- We are here.
```

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
leculver: Yes it should.

leculver: There's a few other places this should be changed and we should rename the functions.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
leculver: The args we pass **to** the dac shouldn't matter.  On the dac side these are going to get casted down to 32bit integers/pointers so we shouldn't need to sign extend anything ourselves.  Please let me know if you find that's not the case because it would be very surprising...

leculver:
Sorry I didn't reply back here, I saw the message and it was on my todo list.

> Replacing the above line (which causes ImmutableArray<ClrInfo> does not contain a definition for 'Single') with

Did you add `using System.Linq;` to the top of the file?

> `Failed to get Free MethodTableData`
> Maybe the error is specific to my machine. I just don't know what the Free methodtable is for and the only thing I've figured out with lldb is pEEClass is not NULL in this context.

I haven't had a chance to try on device, but it's surprising to me.  I can't repro it on an x86/x64 device.  It's possible you are running into a real bug in the dac.  We can work around this by creating our own Free type without calling into the dac for it, but I'd like to get to the bottom of the problem to report it to the CLR diagnostics team.  I'm planning to debug this on-device today, assuming I can get one to work.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
leculver: https://github.com/microsoft/clrmd/pull/448

leculver:
Let's move further discussion back to https://github.com/microsoft/clrmd/issues/402 so I can keep up with the thread. =)

PR #422, opened 2019-12-13
Use PEReader to get metadata info from PEImage

swift-kim:
Although it's not such a critical issue, there are some cases that `DacDataTargetWrapper.GetMetadata()` fails to get metadata information from IL images, because the `mdRva` value given by the Dac is 0 and it's not available from `PEImage.OptionalHeader.ComDescriptorDirectory.VirtualAddress` either.

Instead, we may use `PEReader.GetMetadata()` to get the information from any images. The idea is from `SOS.MetadataHelper.GetMetadataLocator()` in [SOS.MetadataHelper](https://github.com/dotnet/diagnostics/blob/master/src/SOS/SOS.NETCore/MetadataHelper.cs). We may also add support for pdbs if required.

I know that this is a kind of issue stated by https://github.com/microsoft/clrmd/issues/311#issuecomment-565125956, so please just take a look for motivation. There's no need to merge this.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DacInterface/DacDataTargetWrapper.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DacInterface/DacDataTargetWrapper.cs
@@ -6,6 +6,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Reflection.PortableExecutable;
 using System.Runtime.InteropServices;
 using System.Threading;
 using Microsoft.Diagnostics.Runtime.DbgEng;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/DacInterface/DacDataTargetWrapper.cs
```diff
@@ -254,30 +255,29 @@ public int Request(IntPtr self, uint reqCode, uint inBufferSize, IntPtr inBuffer
...
             PEImage? peimage = _dataTarget.LoadPEImage(filePath);
             if (peimage is null || peimage.OptionalHeader is null)
                 return E_FAIL;
 
             DebugOnly.Assert(peimage.IsValid);
 
-            uint rva = mdRva;
-            uint size = bufferSize;
-            if (rva == 0)
+            Stream peStream = peimage.Stream;
```
leculver: This change made me realize that none of the PEImage code is threadsafe...  I need to fix that, but it's not something you have to address in this change.

leculver:
The change looks good, but I'm holding off on merging this until I decide whether to use System.Reflection.Metadata or not.  (I plan to use it as long as it's not missing anything I need, but I want to actually do the investigation before merging this.)

If I decide not to use that library I'll figure out how it's getting the metadata location and reimplement that here.

Thanks!

NextTurn:
@leculver The only thing missing from the `PEReader` in S.R.M is the `ResourceEntry`.

leculver:
Ah.  We do make use of the resource reading code.  The main thing is figuring out how to read resources if I swap over to that library.  I haven't really taken a look at it yet.

NextTurn:
Reading resources is almost the same as before.

I have a [WIP](https://github.com/NextTurn/clrmd/tree/peimage) branch in my fork if you have time for a glance.

leculver:
That WIP does seem pretty reasonable so far.

swift-kim:
@leculver Great. I'll also investigate further why it's not getting metadata correctly if you decide not to use S.R.M. Thanks.

NextTurn:
I'm interested in why these changes make any difference. Their underlying implementations look much the same. It'll be great if you could figure out why or when `PEReader` succeeds and `PEImage` fails. Thanks!

swift-kim:
The fundamental issue has been fixed by the new PR, so I think we can discard this change for better thread safety and respecting the value `mdRva` from the Dac.

Issue #163, opened 2018-12-08
Use of IList<T>


drewnoakes:
Several types use `IList<T>` to expose collections of data. `ClrType` is a good example.

There are a few issues with this:

- Enumerating them requires allocating an enumerator
- They're mutable
- Calls are virtual

The mutability thing is not a problem when everyone is well behaved, but the allocation/virtual points are only avoidable by casting to `List<T>` or `T[]`, which is suboptimal and risky.

Would you consider a dependency upon `System.Collections.Immutable` and using `ImmutableArray<T>` in place of `IList<T>`? When used well, it provides the performance and compactness of arrays with immutability and essentially zero overhead. It would be a breaking API change however.

leculver:
When was ImmutableArray added?  Currently ClrMD is supported as far back as CLR 4.5 (and needs to be carried without dependencies on Desktop CLR).  Was ImmutableArray around in 4.5?

I do wish I hadn't used IList<T> so long ago, but I worry it's too big of a breaking change to modify these return types.

drewnoakes:
I'm pretty sure it's always been distributed as a NuGet package. The latest release supports `net45`.

https://www.nuget.org/packages/System.Collections.Immutable

https://github.com/dotnet/corefx/blob/master/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.cs

`ImmutableArray<T>` does implement `IList<T>` which would ease a transition, however it's a struct which means boxing and, more significantly, no `null` value (you use `IsDefault` or `IsDefaultOrEmpty` properties). For APIs that can never return `null` this might not be such a big deal.

I get the importance of backwards compatability.

If it's something you wanted to try out, I would be willing to help.

leculver:
That runs into the second part of the problem, ClrMD on desktop CLR is supposed to run without any dependencies (not having to carry other DLLs).  There's still a small window of time we could probably make breaking changes for 1.0 of clrmd (especially since I need to do a large round of bugfixing).  Let me think about this.

drewnoakes:
If it helps, the actual implementation of `ImmutableArray` is very simple and could be easily replicated within ClrMD directly to avoid any dependencies. There's no need for the other immutable collection types if you just want the immutable array container. Again, I'm happy to spike this out if you feel it might be worth it. No worries either way.

leculver:
[...]

PR #423, opened 2019-12-13
Switch to ImmutableArray

NextTurn:
Contributes to #163 

Creating `ImmutableArray<T>`s has copying overheads if the exact size can't be known in advance.

Remaining public APIs that have interface type results:

- `IModuleHelpers.GetSortedTypeDefMap`
- `IModuleHelpers.GetSortedTypeRefMap`

The result size can sometimes be large.

- `IHeapHelpers.CreateSegments`

Needs evaluation.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/CacheOptionsTest.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/CacheOptionsTest.cs
@@ -1,7 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
 using System.Linq;
-using System.Text;
 using Xunit;
 
 namespace Microsoft.Diagnostics.Runtime.Tests
@@ -132,8 +133,8 @@ namespace Microsoft.Diagnostics.Runtime.Tests
             ClrType type = module.GetTypeByName("Foo");
             Assert.NotEqual(0ul, type.MethodTable);  // Sanity test
 
-            Assert.Same(type.Fields, type.Fields);
-            Assert.Same(type.StaticFields, type.StaticFields);
+            Assert.Equal(type.Fields, type.Fields);
+            Assert.Equal(type.StaticFields, type.StaticFields);
 
             ClrField field = type.GetFieldByName("o");
             ClrField field2 = type.Fields.Single(f => f.Name == "o");
@@ -154,7 +155,7 @@ namespace Microsoft.Diagnostics.Runtime.Tests
             ClrType type = module.GetTypeByName("Foo");
             Assert.NotEqual(0ul, type.MethodTable);  // Sanity test
 
-            Assert.NotSame(type.Fields, type.Fields);
+            Assert.NotEqual(type.Fields, type.Fields);
 
             ClrField field = type.GetFieldByName("o");
             ClrField field2 = type.Fields.Single(f => f.Name == "o");
--- a/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Buffers;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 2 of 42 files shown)


src/Microsoft.Diagnostics.Runtime/src/ImmutableArrayExtension.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    internal static class ImmutableArrayExtension
+    {
+        /// <summary>
+        /// <b>WARNING:</b> Typing this method may cause Visual Studio IntelliCode 2.2 to crash.
+        /// </summary>
+        internal static unsafe Span<T> DangerousGetSpan<T>(this ImmutableArray<T>.Builder builder) where T : unmanaged
+        {
+            // MemoryMarshal.CreateSpan isn't available on .NET Standard 2.0
+            fixed (T* pointer = &builder.ItemRef(0))
```
leculver: I don't think this is correct.  We are fixing a pointer to an immutable array then returning it as a Span.  If the GC decides to relocate this poitner   The memory backed by ImmutableArray<T>.Builder.ItemRef(0) could be relocated by the GC at any time after returning, leading to heap corruption if we try to write to it after that.  (Even if the runtime _can_ track this pointer, I still think I'd like to avoid this pattern.  It looks wrong from code review, and I'd be hard pressed to prove its correctness.)

Since we only use this method in one place around creating the buildid, I'd rather just be inefficient at creating the buildid...  Stackalloc the byte[] to get the buildid, then copy it over to an ImmutableArray, it's ok if we have to be a little inefficient when it makes sense to.

src/Microsoft.Diagnostics.Runtime.Tests/src/CacheOptionsTest.cs
```diff
@@ -132,8 +133,8 @@ public void FieldCachingTest()
             ClrType type = module.GetTypeByName("Foo");
             Assert.NotEqual(0ul, type.MethodTable);  // Sanity test
 
-            Assert.Same(type.Fields, type.Fields);
-            Assert.Same(type.StaticFields, type.StaticFields);
+            Assert.Equal(type.Fields, type.Fields);
```
leculver: This isn't quite asserting the same thing.  The intent here was to assert that we only created the underlying list once and didn't recreate it every call.

Is that not possible to assert with ImmutableArray or have we actually changed the semantics, and now we recreate the immutable array every call?  If it's not possible to assert, can you assert that each entry in Fields/StaticFields are .Same?  That is close to what I was trying to do in this test.

src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
```diff
@@ -109,15 +110,19 @@ public void Dispose()
...
+            out IReadOnlyList<FinalizerQueueSegment> fqRoots,
+            out IReadOnlyList<FinalizerQueueSegment> fqObjects)
         {
             List<ClrSegment> segs = new List<ClrSegment>();
             List<AllocationContext> allocContexts = new List<AllocationContext>();
             List<FinalizerQueueSegment> finalizerRoots = new List<FinalizerQueueSegment>();
             List<FinalizerQueueSegment> finalizerObjects = new List<FinalizerQueueSegment>();
 
-            if (allocContexts.Count == 0)
+            if (true)
```
leculver: Ah, the `allocContexts.Count == 0` is bad leftover code from when I tried to cache that value earlier.  Can you just remove the `if (true)`?

leculver: and the brackets

src/Microsoft.Diagnostics.Runtime.Tests/src/CacheOptionsTest.cs
NextTurn: `ImmutableArray`s equality comparison checks the reference equality of the underlying arrays.

leculver: Ah great!

src/Microsoft.Diagnostics.Runtime/src/ImmutableArrayExtension.cs
NextTurn: Well, that unmanaged pointer passed to the span will be stored as a managed pointer, and since any managed pointer is tracked by GC it can by updated whenever the array is relocated. And the span will always be valid because the array doesn't come from array pool and therefore won't be modified for reuse. So this method is quite safe...

If you don't like this idea, what about a hack that directly converts an array into an immutable array?

If you don't like both... I'll take copying instead.

leculver: A coworker pointed out that since I ship two copies of the library anyway, we might as well change the netstandard 2.0 one over to .Net Core 2.1.  Then we could just use MemoryMarshal.CreateSpan and avoid this problem altogether.  Let me make a PR for that real quick and then you can just use CreateSpan.

(Also I'm fine with hacks in general if they get around stuff like this, so I'd be fine with one that converts an array to an immutable one, but I think we can avoid both by updating the library to target .Net Core 2.1 instead of netstandard.)

leculver: https://github.com/microsoft/clrmd/pull/426.

I'm going to let the CI builds finish to make sure this doesn't break anything unexpected in the weird CI build environment.

leculver: Sigh.  After all of that I forgot MemoryMarshal.CreateSpan isn't in 4.7.1, so it's a moot point.

Go for the hack that converts an array into an immutable array.

leculver: It's odd that System.Memory.dll doesn't have MemoryMarshal.CreateSpan.  I'm not sure why that's the case, but I doubt it's an oversight.  It might just be that it can't be supported on desktop CLR due to the implementation.

NextTurn: It it in .NET Standard 2.1 that .NET Framework won't support. Nothing to do with the implementation...

leculver: 66a93ff seems fairly reasonable to me.  Let me know when this is ready for review/merge.

NextTurn: This is ready for merge. Further updates (if any) can be made in separate PRs.

leculver: Done.

NextTurn: FYI `MemoryMarshal.CreateSpan` needs runtime support, but the `System.Memory` package targets .NET Standard and is not exclusive for CLRs.

From dotnet/corefx# 32943:

> MemoryMarshal.CreateSpan is the new method name, and is available in .NET Core 2.1. It's not available in the portable implementation because it's not implementable safely without runtime support (it would have similar problems to what you cite regarding passing around pointers).

drewnoakes:
Very nice!

PR #427, opened 2019-12-14
Fix regression in ReadPdbs

arvindshmicrosoft:
2.0 had a regression wherein DefaultPdb would return null though the
binary had valid debug info embedded. The correct offset is now being
used to check for the CV7 signature, thereby fixing the regression.


arvindshmicrosoft:
@leculver I would like to add a test case for PEImage.DefaultPdb which in turn ends up at ReadPdbs() so that we can detect any future issues on this code path. I add a test and push an update to this PR, please hold off any final actions till then.

leculver:
Sorry, I'm about to ship an updated 2.0 beta build today or tomorrow.  I would like to have this fix.  Even though you said wait for a test to merge it, I am going ahead and merging it now to get this fix into that build.

PR #429, opened 2019-12-14
Clean up spans and references

NextTurn:


NextTurn:
From ECMA-335 III.3.47 `localloc`

> *address* is aligned so that any built-in data type can be stored there using the `stind` instructions and loaded using the `ldind` instructions.

src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
```diff
@@ -146,13 +146,13 @@ public void Dispose()
                     {
                         // As long as we got at least one heap we'll count that as success
                         result = true;
-                        ProcessHeap(segBuilder, clrHeap, in heap, allocContexts, segs, finalizerRoots, finalizerObjects);
+                        ProcessHeap(segBuilder, clrHeap, heap, allocContexts, segs, finalizerRoots, finalizerObjects);
```
leculver: ClrHeap is a relatively large struct, and passing it by read-only reference was intentional.  Before I review the rest of the changes, why are you proposing changing these to by-value?  My cutoff here is roughly "if a struct is larger than 3 pointers, consider passing a readonly by-ref instead of by-value".

NextTurn: Ah, it's still passed by reference. The `in` modifier is unnecessary at the call site. We don't usually use that...

src/Microsoft.Diagnostics.Runtime/src/Implementation/ValueReader.cs
```diff
@@ -181,19 +181,28 @@ internal static class ValueReader
...
+
             try
             {
-                if (!reader.Read(dataAddress, new Span<byte>(buffer, 0, length * 2), out int count))
+                if (!reader.Read(dataAddress, buffer, out int count))
                     return null;
 
-                return Encoding.Unicode.GetString(buffer, 0, count);
+#if NETCOREAPP
+                return Encoding.Unicode.GetString(buffer.Slice(0, count));
```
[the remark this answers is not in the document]
NextTurn: Added to to-do list.

Issue #415, opened 2019-12-12
Consider using System.Reflection.Metadata


leculver:
Instead of using IMetaDataImport and our own PEFile implementation, consider using System.Reflection.Metadata in its place, if possible.

leculver:
Additionally, PEImage needs to be threadsafe.

NextTurn:
What are the benefits of using S.R.M in place of `IMetaDataImport`?

leculver:
I'm not sure if there are any.  It's more of something I wanted to investigate if there were benefits or if we should maintain the status quo.

PR #430, opened 2019-12-14
Initial works on reimplementing PEImage using S.R.PE.PEReader

NextTurn:
Contributes to #415 

I'm considering drafting an API proposal with regard to `IMAGE_RESOURCE_DIRECTORY`. Any thoughts?


NextTurn:
`ModuleBuilder.Size` returning zero is causing `PdbTests.PdbEqualityTest` to fail on Linux (unrelated to this PR) and `PEReader` to fail after this PR.

https://github.com/microsoft/clrmd/blob/dfe156cc75ec468f2a6ee077120b6ab1a2efc66a/src/Microsoft.Diagnostics.Runtime/src/Builders/ModuleBuilder.cs#L31

https://github.com/microsoft/clrmd/blob/dfe156cc75ec468f2a6ee077120b6ab1a2efc66a/src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdModule.cs#L73

NextTurn:
The failures are caused by conservative behaviors of some readers. `DbgEngDataReader` and `LiveDataReader` always try to read memory, sometimes succeeding even if the address is not on the memory maps. `CoreDumpReader` and `LinuxLiveDataReader` check the maps for every read operation. `PEReader` need the exact image size to function, which isn't available if the image base address is not on the maps.

NextTurn:
`PEReader` needs the exact image size to function, which isn't available if the image base address is not on the module maps, or note maps on Linux. Using segment maps may fix this issue on Linux.

leculver:
Maybe we could use [DacpGetModuleData](https://github.com/dotnet/runtime/blob/master/src/coreclr/src/inc/dacprivate.h#L980) from `IXCLRDataModule` to get the full module size?  I can take a look at wrapping that struct/request/interface later today.

NextTurn:
`DacpGetModuleData` isn't available on the desktop CLR. It was added in dotnet/coreclr# 6783.

leculver:
This is (mostly) a Linux problem right?  Desktop CLR isn't used on Linux.  It's ok if .Net Core has better/more complete data if we can't get the information on Desktop CLR.

NextTurn:
It's an issue on both platforms. Also reported by #438 on Windows.

leculver:
Ah, right.  Still though, I'll fix it any way I can.  If we can only make it better on .Net Core then that's what I'll do.  Is there any option for fixing this on Desktop?  We can read the size out of the PE header (if we can find it) on Desktop CLR and report that, if it's close enough to what S.R.* needs.

NextTurn:
If we can read the PE header, we already have half of our own PE reader. Why do we need `PEReader` then?

leculver:
Yep, that's my thoughts exactly.  This is a tricky problem with ClrMD that requires some thought.  =)

There's a a lot of angles and issues lurking here, which is why I haven't tackled the problem yet and was leaving it for later to really think about what we need to do.  Though I will say having a very small internal PE "reader" to get filesize isn't that big of a deal.  It's just reading a few offsets to get the size instead of carrying full implementations of PE image headers.  It's not _terrible_ design to have ClrMD implement just enough to get the size to feed into S.R.PE.PEReader.

Even if we went that route and everything worked perfectly there's still the problem of thread safety.  If ClrMD supports thread safety and hands out PEReader instances then we have no mechanism to coordinate synchronization between users and DacDataTarget.ReadVirtual.

I'll review this change but hold off on merging for now.  I'll get around to figuring out what to do today.

NextTurn:
> Is there any option for fixing this on Desktop?

`MetadataSize + MetadataStart - ILBase` is an alternative size for headers and metadata, but not large enough for `IMAGE_DEBUG_DIRECTORY`s.

Using `int.MaxValue` as the stream size fallback can bypass all size checks in `PEReader` and makes it work. We don't check the size in our implementation anyway.

https://github.com/microsoft/clrmd/blob/375c64ce94e1b512c491b905117b36a97baa98d2/src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdModule.cs#L73

https://github.com/dotnet/runtime/blob/4359ebfc9943608b34f411366b1e544ac45702b7/src/libraries/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/StreamExtensions.cs

NextTurn:
It seems super weird to me that we have to check `IsValid` on every access. We can throw for an invalid image, or returns `null` in a factory method if the image is not a PE file, so that every `PEImage` instance is valid.

leculver:
Is this PR ready for review?

Without having fully looked at it yet, it looks like PEImage is still there, but you were able to delete a ton of code because those structures and implementations live in S.R.PE?

> It seems super weird to me that we have to check `IsValid` on every access. We can throw for an invalid image, or returns `null` in a factory method if the image is not a PE file, so that every `PEImage` instance is valid.

I'd like to avoid throwing for an invalid image just because that gets really tedious in some minidump cases that crop up often internally at Microsoft.  I'd be happy with a change that creates a factory method, like a static `PEImage.Create` that returns null on failure.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
```diff
@@ -11,12 +11,14 @@ namespace Microsoft.Diagnostics.Runtime
     /// <summary>
     /// Provides information about loaded modules in a DataTarget.
     /// </summary>
-    public class ModuleInfo
+    public class ModuleInfo : IDisposable
```
leculver: If possible, I'd like to arrive at a design that doesn't require ModuleInfo to be disposable.  It would be acceptable if we stopped caching the instance here and returned a new PEImage every call to OpenPEImage (or whatever I called that function).  In fact, that might solve a lot of other problems as well internally with having to worry about PEImage thread safety.

I would suggest not tackling that problem in this PR.  Instead leave ModuleInfo as non-IDisposable and file an issue for me to clean this up when the PR is complete.

NextTurn:
This should be ready for review now, I think.

> but you were able to delete a ton of code because those structures and implementations live in S.R.PE?

There are classes for various headers in S.R.PE with more descriptive names. There aren't corresponding structs because they don't use them in parsing, but that's the implementation detail.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
leculver: Perfect, filed this as https://github.com/microsoft/clrmd/issues/465.

leculver:
LGTM

NextTurn:
I should have test this earlier... But most tests are failing now...

leculver:
No worries, we'll just clean it up.  This is why I need to enable CI tests.  =)

leculver:
I'm going to roll this back to get the CI changes through.

leculver:
Would you mind opening a PR for this change again?  Or should I revert #466?

NextTurn:
> Would you mind opening a PR for this change again?

Yes, soon. I've figured out the cause of the failures.

Issue #408, opened 2019-12-11
[Question] MetaDataImport, Generic Type Parameters


holc-tkomp:
Hoping you could put some light on this.

### Questions

1. What happened to `IMetaDataImport2`? Will it ever be implemented?
2. Is there a way to retrieve generic type parameters as an array?

### Some context

This came up on this when I tried to retrieve more detailed informations about generic type using available functionalities and `MetaDataImport` looked like the way to go.

Given that I already have enumerated objects and their `Type` property contains correct `ClrType` objects, I now have access to `ClrObject.Type.MetadataToken`, which should contain `TypeDef token`. With it I can utilize `MetaDataImport`, to retrieve the name without generic params:
_Note: following code was run on version 1.0.5._
```c#
ClrModule _module; // initialized somewhere
ClrType _type; // initialized somewhere
// Get the COM wrapper and token
MetaDataImport _metadata = (MetaDataImport)_module.MetadataImport;
int _token = _type.MetadataToken; // Why it's not some unsigned int btw.?
_metadata.GetTypeDefProperties(_token, out var _name, out _, out _);

// Assuming C# type was "class Generic<T1,T2>"
// _name will contain "MyAssembly.Generic`2"
```
Probably there are more efficient ways to retrieve this, but let it be for now.

From what I see `MetaDataImport` class is a partial implementation of [`IMetaDataImport`](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/metadata/imetadataimport-interface) interface.

Searching on how could I retrieve generic type parameters with it I stepped on [`IMetaDataImport2`](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/metadata/imetadataimport2-interface) interface, which gloriously unveils `EnumGenericParams` and `GetGenericParamProps` methods which looks like exactly what I wanted.

Unfortunately when I looked into source code:

- in `1.0.5` there's a .cs file for this interface, but it does not appear to be finished, there's no class (COM wrapper) using it and comments hint that there's work to be done to make it work;
- in `2.0` (current master branch) this file is gone and it does not appear to be implemented elsewhere,

so it appears like there's no way of accessing those methods at the moment 😞.

leculver:
This is just an oversight on my part.  ClrMD 2.0 is still being built, but I completely forgot about the second interface.

[...]

PR #431, opened 2019-12-15
Add ClrGenericParameter

NextTurn:
Contributes to #408


NextTurn:
`QueryInterface` returns `E_NOINTERFACE` for `IID_IMetaDataImport2` on .NET Framework. Using `IID_IMetaDataImport` as a workaround, since `IMetaDataImport2` is available since .NET Framework 2.0.

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdType.cs
```diff
@@ -124,19 +124,30 @@ private GCDesc GetOrCreateGCDesc()
...
 
+        public override IEnumerable<ClrGenericParameter> EnumerateGenericParameters()
+        {
+            MetaDataImport2? import = Module?.MetadataImport2;
+            if (import is null)
+                yield break;
+
+            foreach (int token in import.EnumerateGenericParams((int)MetadataToken))
+                if (import.GetGenericParamProps(token, out int index, out GenericParameterAttributes attributes, out string? name))
+                    yield return new ClrGenericParameter(index, attributes, name);
```
leculver: We should add the token itself to ClrGenericParameter.

src/Microsoft.Diagnostics.Runtime.Tests/Microsoft.Diagnostics.Runtime.Tests.csproj
```diff
@@ -4,9 +4,6 @@
     <TargetFrameworks>netcoreapp3.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 
-    <!-- This _IS_ a Unit Test Project, but because of build_test_assets.cmd, we don't want to run as part of CI right now -->
```
leculver: This should be in its own pull request.  Currently the tests fail when this block is removed, even on Linux.  I haven't had a chance to dig through it yet.  I agree we should remove this block, but it should be in its own PR with associated fixes.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrGenericParameter.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public readonly struct ClrGenericParameter
+    {
+        public int Index { get; }
+
+        public GenericParameterAttributes Attributes { get; }
+
+        public string Name { get; }
+
+        internal ClrGenericParameter(int index, GenericParameterAttributes attributes, string name)
```
leculver: We should add the metadata token for this generic parameter in case the user wants to grab more data out of IMetaDataReader in the future.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/MetaDataImport2.cs
```diff
@@ -0,0 +1,123 @@
...
+            name = null;
+            InitDelegate(ref _getGenericParamProps, VTable.GetGenericParamProps);
+
+            int hr = _getGenericParamProps(
+                Self, token, out index, out attributes, out int owner, out _, null, 0, out int needed);
+
+            if (hr < 0)
+                return false;
+
+            string nameResult = new string('\0', needed - 1);
```
leculver: I like this pattern.  I'll try to use this in the future for other SOS/Dac calls if I get around to it.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/MetaDataImport2.cs
```diff
@@ -0,0 +1,123 @@
...
+            try
+            {
+                IntPtr handle = IntPtr.Zero;
+                int hr;
+
+                while ((hr = _enumGenericParams(Self, ref handle, token, tokens, tokens.Length, out int count)) >= 0 && count > 0)
+                    for (int i = 0; i < count; i++)
+                        yield return tokens[i];
+
+                CloseEnum(handle);
```
leculver: I'd suggest moving this to the finally block as `if (handle != IntPtr.Zero) CloseEnum(handle);`.

src/Microsoft.Diagnostics.Runtime.Tests/Microsoft.Diagnostics.Runtime.Tests.csproj
NextTurn: This isn't part of the PR, but the PR will be blocked until the tests pass.

NextTurn:
I need to call `QueryInterface` on `IMetaDataImport` with `IID_IMetaDataImport2` to get `IMetaDataImport2`, which isn't possible with the current `CallableCOMWrapper` implementation. Shall I just use `IID_IMetaDataImport` based on the fact that the two interfaces are the same (since .NET Framework 2.0)?

leculver:
Sure.  Or change CallableCOMWrapper to expose what it should, whichever you think is better or more efficient for you.  CallableCOMWrapper is intentionally marked public and should support what you are trying to do, but don't feel you need to fix it.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/MetaDataImport2.cs
NextTurn: Then we have duplicated zero checks in both the caller and the callee.

src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
```diff
@@ -1057,7 +1057,7 @@ private void CreateFieldsForMethodTableWorker(ClrType type, out ImmutableArray<C
             statics = staticsBuilder.MoveOrCopyToImmutable();
         }
 
-        public MetaDataImport? GetMetaDataImport(ClrModule module) => _sos.GetMetadataImport(module.Address);
+        public MetaDataImport? GetMetadataImport(ClrModule module) => _sos.GetMetadataImport(module.Address);
```
leculver: I don't mind following the guidelines you point to as long as it's 100% consistent.  I don't want the class name and the method name to not match.  Note that the reason it's MetaDataImport in ClrMD is because that's how [IMetaDataImport](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/metadata/imetadataimport-interface) is spelled.

If if you change `MetaData` to `Metadata` I'm fine with that (I like it better even) but also please change the `MetaDataImport` class with this PR too.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/MetaDataImport2.cs
leculver: That's ok, having the CloseEnum in the finally block is worth an extra branch.

src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
NextTurn: > Note that the reason it's MetaDataImport in ClrMD is because that's how IMetaDataImport is spelled.

I think the consistency with other types and properties (for example, `MetadataToken`) is more important than to follow the interface's old-style real name. I've renamed that type to adopt the new style.

PR #433, opened 2019-12-15
Remove AttachToProcess tests

NextTurn:


NextTurn:
I don't want these two tests now...

src/TestTargets/Spin/Spin.cs
```diff
@@ -9,6 +9,7 @@ private static void Main(string[] args)
             throw new Exception();
 
         Console.WriteLine();
-        Thread.SpinWait(int.MaxValue);
+
+        while (true) ;
```
leculver: I would prefer a `while (true) Thread.Sleep(1000);` so we don't peg the CPU at max while waiting on the debugger to attach.

src/Microsoft.Diagnostics.Runtime.Tests/Microsoft.Diagnostics.Runtime.Tests.csproj
```diff
@@ -5,7 +5,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 
     <!-- This _IS_ a Unit Test Project, but because of build_test_assets.cmd, we don't want to run as part of CI right now -->
-    <IsUnitTestProject Condition="'$(ContinuousIntegrationBuild)' == 'true'">false</IsUnitTestProject>
+    <IsUnitTestProject Condition="'$(OS)' == 'Windows_NT'">false</IsUnitTestProject>
```
leculver: This is causing the tests to fail CI, so I'd like to move this change to Tests.csproj to its own Pull Request to try to resolve the CI failures before checkin.

(Alternatively if you get CI to pass on this one I don't mind keeping it here, but I'm not sure what's broken yet or what the fix is.)

NextTurn: The problem is that I can't get CI to pass right now. The tests succeeded on all my machines, so I can only debug this by printing more information in tests.

Also I think these can be removed as I can't see any chance they could actually fail on normal machines.

leculver: Ah in that case, feel free to just remove them...  Do you want me to merge this PR to fix it or do you want to submit commit that removes them?

Yeah I'm going to have the same problem with CI, but I've left it for another day.

NextTurn: I can try to resolve the failures sometime later if you don't mind my flood debugging commits.

NextTurn:
The target processes received `SIGQUIT` signals.

src/Microsoft.Diagnostics.Runtime.Tests/Microsoft.Diagnostics.Runtime.Tests.csproj
leculver: Nope, I don't mind at all.  I've appreciated your work throughout all of this effort.  =)

Issue #438, opened 2019-12-16
Modules Size is always 0


PhDuck:
I wanted to use the modules information to figure out the size of loaded assemblies, since I have application that loads thousands of assemblies at runtime.

Unfortunately though it seems that the Size field on ClrModule is always 0. I have tried both memory dumps and attaching to the process.

Version info:
Microsoft.Diagnostics.Runtime 1.1.57604
.NET Core 2.0, 3.0, and 3.1
Windows 10

The following code prints:
0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0

```
class Program
{
    static void Main(string[] args)
    {
        var procc = Process.GetCurrentProcess();
        using (var dtt = DataTarget.AttachToProcess(procc.Id, 1000, AttachFlag.Passive))
        {
            ClrRuntime runtime = dtt.ClrVersions.First().CreateRuntime();
            Console.WriteLine(String.Join(',', runtime.Modules.Select(e => e.Size)));
        } 
    } 
}
```

leculver:
Is this an x86 or x64 process?

PhDuck:
x64

NextTurn:
Possibly related to https://github.com/microsoft/clrmd/pull/430#issuecomment-566161896.

leculver:
@PhDuck What are you using the size for?

As I'm looking closer at fixing this there's a few different interpretations of what 'Size' could mean (the original design here is ambiguous which is bad).  For example, it could mean the linker image size, the size as loaded into the virtual address space, the size on disk of the PE image...

If I had a better understanding of what you are using it for that might inform what we should really be producing as the size here.

leculver:
This is now fixed in the current versions of 1.1 and 2.0 on NuGet today.  It now works as well as it has in the past...we are still trying to figure out what this looks like going forward for 2.0.

PhDuck:
Hey @leculver my usage is to measure the memory consumption of having the assembly loaded into memory.

In our product we load in thousands of assemblies, so I wanted to see the potential gain of unloading them, from the perspective of the assembly. 

Afterwards I will have to figure out the impact on the loader heap of loading these C# assemblies and creating classes from them.

dcorbin-wintellect:
Note: I am currently trying to accomplish something extremely similar, and having the same issue with the size always being zero....

PR #442, opened 2019-12-17
Add ClrDataModule

NextTurn:
`DacpModuleData` was renamed to `ModuleData`. How to name `DacpGetModuleData`?


src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
```diff
@@ -344,13 +344,25 @@ public bool GetRCWData(ulong rcw, out RCWData data)
...
+        }
+
         public MetaDataImport? GetMetadataImport(ulong module)
         {
             if (module == 0)
                 return null;
 
-            InitDelegate(ref _getMetaData, VTable->GetMetaDataImport);
-            if (_getMetaData(Self, module, out IntPtr iunk) != S_OK)
+            InitDelegate(ref _getModule, VTable->GetModule);
```
leculver: This isn't intentional right?

src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataModule.cs
```diff
@@ -0,0 +1,97 @@
...
+                    return true;
+
+            data = default;
+            return false;
+        }
+
+        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
+        private delegate int RequestDelegate(IntPtr self, uint reqCode, int inBufferSize, void* inBuffer, int outBufferSize, void* outBuffer);
+
+        internal struct UntitledModuleData
```
leculver: Maybe `ExtendedModuleData`?

src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataModule.cs
```diff
@@ -0,0 +1,97 @@
...
+        private RequestDelegate? _request;
+
+        public ClrDataModule(DacLibrary library, IntPtr pUnknown)
+            : base(library?.OwningLibrary, IID_IXCLRDataModule, pUnknown)
+        {
+        }
+
+        private ref readonly IClrDataModuleVTable VTable => ref Unsafe.AsRef<IClrDataModuleVTable>(_vtable);
+
+        internal bool GetModuleData(out UntitledModuleData data)
```
leculver: You should mark this (and the struct) public.  We do expose these things publicly in case someone is using ClrMD to build their own implementation.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
NextTurn: From what I saw in [sospriv.idl](https://github.com/dotnet/runtime/blob/137aa8460fd18788b00d53b22ab127bbba5f9132/src/coreclr/src/inc/sospriv.idl#L197) or [sospriv.h](https://github.com/dotnet/runtime/blob/137aa8460fd18788b00d53b22ab127bbba5f9132/src/coreclr/src/pal/prebuilt/inc/sospriv.h#L747) the method name is `GetModule`.

src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
```diff
@@ -1057,6 +1050,8 @@ private void CreateFieldsForMethodTableWorker(ClrType type, out ImmutableArray<C
             statics = staticsBuilder.MoveOrCopyToImmutable();
         }
 
+        public ClrDataModule? GetClrDataModule(ClrModule module) => _sos.GetClrDataModule(module.Address);
```
leculver: I am not sure whether we should expose this through ClrModule or not.  For example, ClrThread doesn't expose `ClrStackWalk` which is similar to this.  I would suggest using ClrDataModule/UntitledModuleData to fill in the properties of ClrModule more accurately if it exists, but not directly expose it through ClrModule's API surface.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs
leculver: You have `GetClrDataModule` and `GetMetadataImport` both using the same `VTable->GetModule`, which can't be right?

leculver: Or do both go through the same VTable slot, and the QI determines the interface?

leculver: Ah nevermind, I get it now.  =)

leculver:
I think this draft is ready to merge as-is.  Even though we haven't fully fixed the issue on Desktop CLR, this looks like a solid point to check in, unless you want to do more here before it's merged?

NextTurn:
I'd like to hold this off for a while. I've opened #447 to partially fix the zero size issue.

As mentioned in https://github.com/microsoft/clrmd/issues/438#issuecomment-566797610, we have several module sizes either on disk or in memory that are not clearly distinguished. My question is that are the file size and the memory size always equal so that we can use them interchangeably? If not, when and why will they be different?

CoreCLR's PEFile handles both sizes in the same way:

```cpp
    COUNT_T GetSize() const;          // size of file on disk, as opposed to GetVirtualSize()
```

https://github.com/dotnet/runtime/blob/1ed0a5661b5b9f2d9938478bad97140bbb272dc6/src/coreclr/src/inc/pedecoder.inl#L196

```cpp
    COUNT_T GetVirtualSize() const; // OptionalHeaders.SizeOfImage - size of mapped/expanded image in memory
```

https://github.com/dotnet/runtime/blob/1ed0a5661b5b9f2d9938478bad97140bbb272dc6/src/coreclr/src/inc/pedecoder.inl#L330

leculver:
The real problem with exposing Size at all is that it was never well defined whether we mean size on disk or virtual size.  The original meaning was intended to be VirtualSize, but it's unclear if we ever produced the correct value most of the time in older versions.

Also keep in mind that on x86/Windows, the .Net Runtime doesn't call LoadLibrary on managed modules (this is true of Desktop, not sure if it was changed in .Net Core), meaning there won't be a ModuleInfo created for those modules...  So we're still setting `Size` to 0 on x86 even after PR #447.

A couple options:

1.  One option here is to simply remove it and make the user go through ModuleInfo.GetPEImage (and make sure PEImage has the ability to get or calculate those sizes).  That'll still be broken on x86 where there won't be a ModuleInfo for those modules.
2.  Another option is to rename `Size` to be `VirtualSize` and/or `SizeOfImage` and just produce those properties if we can by loading PEImage and calculating the result on demand.

NextTurn:
That explains why I only saw the main *.exe* in crash dumps. On .NET Core we have `ModuleInfo` for every managed module loaded, on both platforms.

```
ModLoad: 00000000`7b5d0000 00000000`7be27000   C:\Program Files (x86)\dotnet\shared\Microsoft.NETCore.App\3.1.0\System.Private.CoreLib.dll
ModLoad: 00000000`07a00000 00000000`07a08000   clrmd\src\TestTargets\Types\bin\Debug\netcoreapp3.0\Types.dll
ModLoad: 00000000`07a00000 00000000`07a08000   clrmd\src\TestTargets\Types\bin\Debug\netcoreapp3.0\Types.dll
ModLoad: 00000000`7c050000 00000000`7c05e000   C:\Program Files (x86)\dotnet\shared\Microsoft.NETCore.App\3.1.0\System.Runtime.dll
ModLoad: 00000000`07860000 00000000`07868000   clrmd\src\TestTargets\Types\bin\Debug\netcoreapp3.0\SharedLibrary.dll
ModLoad: 00000000`07860000 00000000`07868000   clrmd\src\TestTargets\Types\bin\Debug\netcoreapp3.0\SharedLibrary.dll
ModLoad: 00000000`7c160000 00000000`7c1a8000   C:\Program Files (x86)\dotnet\shared\Microsoft.NETCore.App\3.1.0\System.Collections.dll
```

NextTurn:
Since #447 has fixed the issue on .NET Core we don't even need this now.

leculver:
> Since #447 has fixed the issue on .NET Core we don't even need this now.

I'd still like to merge the changes when you feel they are ready.  This isn't used by ClrMD but this wrapping may be useful to others or for testing CLR diagnostics.  Also if I end up needing it or ClrDataModule in the future it would be nice not to have to start from scratch.

NextTurn:
OK. I'll see if I can fix the PDB issue here.

NextTurn:
The PDB issue has been fixed in https://github.com/microsoft/clrmd/pull/442/commits/8adc07110edb15b6cfc21351e012d277be15f82b. Dynamic modules and modules written in IL don't have PDBs so you can't check against null for them (#444).

Consider choosing a consistent name regarding file layout:

- IsVirtual - ClrMD
- !IsFileLayout - CoreCLR
- IsLoadedImage - S.R.M

NextTurn:
> are the file size and the memory size always equal so that we can use them interchangeably?

Unfortunately these sizes can be completely different. For example in a *Types* coredump, the image size of *Types.dll* in the ELF note is 0x2000, `DacpGetModuleData.LoadedPESize` is 0x1600, while the full `SizeOfImage` is 0x8000. In that case the `SizeOfImage` may not be a valid value to be used (#452).

So I'm still going to remove that size dictionary.

NextTurn:
> are the file size and the memory size always equal so that we can use them interchangeably?

Unfortunately these sizes can be completely different. For example in a *Types* coredump, the image size of *Types.dll* in the ELF note is 0x2000, `DacpGetModuleData.LoadedPESize` is 0x1600, while the full `SizeOfImage` is 0x8000. In that case the `SizeOfImage` may not be a valid value to be used (#452).

> there won't be a ModuleInfo created for those modules...

At least there are `ModuleInfo`s for *mscorlib.dll* and the executable on desktop.

leculver:
>  Consider choosing a consistent name regarding file layout:

I'd prefer to more closely align with CoreCLR and uses `!IsFileLayout`.

> Do we have so many exceptions?

We can remove that catch block that you are looking at.  I cannot remember why I put it there, but reading through the code it looks wrong to have a blanket catch (or any try/catch) in that location.

leculver:
> Unfortunately these sizes can be completely different.

I think the best thing here is to make this the "size as loaded into memory", and rename it from `Size` to be `VirtualSize` (meaning size in the virtual address space).  My reasoning is that it lets you easily know if an arbitrary address in the process is within the module, such as `module.ImageBase <= address && address < module.ImageBase + module.VirtualSize`.  We shouldn't pick SizeOfImage because it can be obtained by creating a PEImage (`ModuleInfo.GetPEImage`) and reading the value from it.

NextTurn:
We have a few places where the names keep changing along the data flow with the meaning unchanged. These changes don't make much sense.

`ClrmdModule.Imagebase <- ModuleBuilder.ILImageBase <- ModuleData.ILBase`

NextTurn:
`VirtualSize` is semantically correct but it diverges from the CoreCLR's. In CoreCLR the `VirtualSize` is exactly the `SizeOfImage` (mapped/expanded) (https://github.com/microsoft/clrmd/pull/442#issuecomment-567108707).

Also note that in my example above *Types.dll* is file layout (not mapped/expanded, **`isVirtual` is `false`**).

leculver:
> VirtualSize is semantically correct but it diverges from the CoreCLR's

That's frustrating, but I'd still rather use `VirtualSize` and be semantically correct over matching CoreCLR.  

> We have a few places where the names keep changing along the data flow with the meaning unchanged. These changes don't make much sense.

That's just a leftover from my endless refactoring, feel free to pick whichever one makes the most sense and rename the rest.

NextTurn:
> That's frustrating, but I'd still rather use VirtualSize and be semantically correct over matching CoreCLR.

Sorry I just realized that `VirtualSize` is not semantically correct... How can we have a `VirtualSize` when `isVirtual` is false? Then is it virtual in the end?

From my point of view `LoadedFileSize` (or more misleading `FileSize`) may be a descriptive and correct name to describe that **possibly truncated file size loaded into memory**.

leculver:
> How can we have a VirtualSize when isVirtual is false? Then is it virtual in the end?

Good point.  `LoadedFileSize` sounds fine to me, even if it's possibly truncated.  We'll just document what the size is.

NextTurn:
Let's proceed to merge this and leave the naming issues in a later PR.

We may also need a more appropriate name in place of `isVirtual` in `PEImage` parameters, now that whether the file is loaded (`IsLoadedImage`) or in the virtual address space (`IsVirtual`) doesn't determine its layout.

More choices here:

- `!IsFileLayout` - DAC `ClrDataModule`
- `IsMappedLayout`/`IsMemoryLayout`/...
- `!IsFlat` - CoreCLR `PEDecoder`
- `IsMapped` - CoreCLR `PEDecoder`

PR #445, opened 2019-12-18
Fix metadata reading

swift-kim:
Alternative fix for #422.

Yeah, CorHeader (#192) is already there but never used.


leculver:
Much appreciated!

Issue #436, opened 2019-12-16
Non-deterministic failures in TestFinalizerQueueObjects


NextTurn:
https://dev.azure.com/dnceng/public/_build/results?buildId=458580&view=ms.vss-test-web.build-test-results-tab

https://dev.azure.com/dnceng/public/_build/results?buildId=458803&view=ms.vss-test-web.build-test-results-tab

NextTurn:
I believe this is an issue with the test target, not the codes.

leculver:
GCRoot tests also looks to have non-deterministic failures.  That one looks like a real product issue.

Issue #417, opened 2019-12-12
Rewrite documentation and tutorials


leculver:
Update/rewrite tutorials before ClrMD 2.0 RTM.

PR #456, opened 2019-12-19
Standardize documentation

NextTurn:
Possibly the last PR before you come back from vacation. Have fun!

I'll keep pushing overtime but every single commit is ready for review.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/Helpers.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/Helpers.cs
@@ -84,11 +84,11 @@ namespace Microsoft.Diagnostics.Runtime.Tests
             return runtime.AppDomains.Single(ad => ad.Name == domainName);
         }
 
-        public static ClrModule GetModule(this ClrRuntime runtime, string filename)
+        public static ClrModule GetModule(this ClrRuntime runtime, string fileName)
         {
             return (from module in runtime.AppDomains.SelectMany(ad => ad.Modules)
                     let file = Path.GetFileName(module.FileName)
-                    where file.Equals(filename, StringComparison.OrdinalIgnoreCase)
+                    where file.Equals(fileName, StringComparison.OrdinalIgnoreCase)
                     select module).FirstOrDefault();
         }
 
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrField.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrField.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Diagnostics.Runtime
     public abstract class ClrField
     {
         /// <summary>
-        /// Gets the ClrType containing this field.
+        /// Gets the <see cref="ClrType"/> containing this field.
         /// </summary>
         public abstract ClrType Parent { get; }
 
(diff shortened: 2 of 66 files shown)


leculver:
I'll actually be home during my vacation with my parents in town, so I'll still be here reviewing and merging changes as I need a break from family...  I just won't be doing any programming.  =)

Will take a look at this shortly.

leculver:
The overall change looks good, nothing seems controversial.  The code coverage bot looks great, I'd love to install it, but I'm very bad at setting that kind of thing up.  (I'm also not sure whether or not our tests are being fully run by CI right now.)

NextTurn:
> but I'm very bad at setting that kind of thing up.

It could be as simple as setting up the Azure Pipelines app.

> I'm also not sure whether or not our tests are being fully run by CI right now.

I think testing can be enabled once #436 is resolved.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrField.cs
```diff
@@ -43,9 +43,9 @@ public abstract class ClrField
         public virtual bool IsPrimitive => ElementType.IsPrimitive();
 
         /// <summary>
-        /// Gets a value indicating whether this field is a ValueClass (struct).
+        /// Gets a value indicating whether this field is a value type.
         /// </summary>
-        /// <returns>True if this field is a ValueClass (struct), false otherwise.</returns>
+        /// <returns>True if this field is a value type, false otherwise.</returns>
         public virtual bool IsValueClass => ElementType.IsValueClass();
```
leculver: We should probably rename IsValueClass (and all related instances) to ValueType since that's what the class is called? 
 [ValueType](https://docs.microsoft.com/en-us/dotnet/api/system.valuetype?view=netframework-4.8).

NextTurn: That's my thought. I guess this was named `ValueClass` because there is an `IsValueClass` API in some `ICorDebug*` interface?

leculver: Yeah it's from old diagnostics code and the term made its way into ClrMD ages ago.  I'll file an issue to remember to do it.

src/Microsoft.Diagnostics.Runtime.Tests/src/Helpers.cs
```diff
@@ -84,11 +84,11 @@ public static ClrAppDomain GetDomainByName(this ClrRuntime runtime, string domai
             return runtime.AppDomains.Single(ad => ad.Name == domainName);
         }
 
-        public static ClrModule GetModule(this ClrRuntime runtime, string filename)
+        public static ClrModule GetModule(this ClrRuntime runtime, string fileName)
```
leculver: As a personal preference, I hate `fileName`, but I suppose I will live with it because it is more consistent.  =)

NextTurn:
> It would be good to have instructions or similar checked in somewhere on re-running what you did to make this cleanup.

Having too many checks/warnings tends to hurt developer productivity. I can add the analyzers when you start working on #417.

leculver:
I don't mean automatically.  Just something that is documented so I understand how to run it again manually every 6 months or so.  =)

PR #457, opened 2019-12-19
Implement IDataReader.SearchMemory

leculver:
- Searching through memory to look for a sequence of bytes is a useful thing to have.
- Simplify GetVersionInfo on Linux by using SearchMemory, and by manually parsing the ascii version string.
- Simplify reading of Ascii characters in ReadNullTerminatedAscii.  Since the text we are looking for will have the matching code points between ascii and unicode we can just Append them directly to StringBuilder and not go through Encoding.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Extensions/MemorySearcher.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Extensions/MemorySearcher.cs
@@ -1,0 +1,83 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Buffers;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    /// <summary>
+    /// A public extension methods to support searching an IMemoryReader for a given span.
+    /// </summary>
+    public static class MemorySearcher
+    {
+        /// <summary>
+        /// Searches memory from startAddress to endAddress, looking for the memory specified by `searchFor`.  Note
+        /// that this is NOT meant to be used to search the entire address space.  This method will attempt to read
+        /// all memory from startAddress to endAddress, so providing very large ranges of memory will make this take
+        /// a long time.
+        /// </summary>
+        /// <param name="reader">The memory reader to search through.</param>
+        /// <param name="startAddress">The address to start searching memory.</param>
+        /// <param name="endAddress">The address to stop searching memory.</param>
+        /// <param name="searchFor">The memory to search for.</param>
+        /// <returns>The address of the value if found, 0 if not found.</returns>
+        public static ulong SearchMemory(this IMemoryReader reader, ulong startAddress, ulong endAddress, ReadOnlySpan<byte> searchFor)
+        {
+            if (reader is null)
+                throw new ArgumentNullException(nameof(reader));
+
+            // While not strictly disallowed, providing a sanity check to make sure the user isn't
+            // searching from 0 to ulong.MaxValue is a good idea.
+            if (startAddress == 0 || endAddress == ulong.MaxValue)
+                throw new ArgumentException($"The start and end addresses must be within allocated memory.");
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Runtime/src/Extensions/MemorySearcher.cs
```diff
@@ -0,0 +1,83 @@
...
+                {
+                    int bytesRemaining = (int)(endAddress - startAddress);
+                    if (bytesRemaining < searchFor.Length)
+                        break;
+
+                    Span<byte> buffer = new Span<byte>(array, 0, Math.Min(bytesRemaining, array.Length));
+                    if (!reader.Read(startAddress, buffer, out int read) || read < searchFor.Length)
+                        break;
+
+                    buffer = buffer.Slice(0, read);
```
NextTurn: [MemoryExtensions.IndexOf](https://docs.microsoft.com/dotnet/api/system.memoryextensions.indexof#System_MemoryExtensions_IndexOf__1_System_ReadOnlySpan___0__System_ReadOnlySpan___0__) may be used here.

NextTurn:
> we can just Append them directly to StringBuilder and not go through Encoding.

Similar updates to the other `ReadNullTerminatedAscii` overload and possibly other `Encoding.ASCII` uses?

src/Microsoft.Diagnostics.Runtime/src/Extensions/MemorySearcher.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// that this is NOT meant to be used to search the entire address space.  This method will attempt to read
+        /// all memory from startAddress to endAddress, so providing very large ranges of memory will make this take
+        /// a long time.
+        /// </summary>
+        /// <param name="reader">The memory reader to search through.</param>
+        /// <param name="startAddress">The address to start searching memory.</param>
+        /// <param name="endAddress">The address to stop searching memory.</param>
+        /// <param name="searchFor">The memory to search for.</param>
+        /// <returns>The address of the value if found, 0 if not found.</returns>
+        public static ulong SearchMemory(this IMemoryReader reader, ulong startAddress, ulong endAddress, ReadOnlySpan<byte> searchFor)
```
NextTurn: 💡 It's common to accept a size/length parameter to indicate the end.

leculver:
The second overload for ReadNullTerminatedAscii seemed fine, since we read in a full byte[] and then convert it to a string in one pass.  I don't think it buys much to avoid Encoding there and use StringBuilder, it's just more code to do roughly the same thing.  I mostly wanted to get rid of the byte-by-byte reads and the per-byte Encoding calls in the implementation I modified.

I did (somehow) miss that same pattern in `GetFileVersion` though.  I'll make another pr.

leculver:
Ah I remember why I didn't change that.  Anything that makes use of `IAddressSpace` makes me very sad.  That's mostly because the nested nature of those address spaces make it _very_ difficult to debug problems when there's something wrong in `CoreDumpReader`, and because if we want to keep IAddressSpace then it should derive from IMemoryReader instead of defining its own read.

I'm going to leave the last implementation alone until I figure out whether I'm going to rewrite the Linux core dump reader in 2.0, mostly just to eliminate IAddressSpace.  I'll mark it with a `todo` comment, and implement your size/length suggestion.

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
```diff
@@ -96,6 +95,7 @@ public LinuxFunctions()
 
         internal static void GetVersionInfo(IDataReader dataReader, ulong baseAddress, ElfFile loadedFile, out VersionInfo version)
         {
+            Console.WriteLine($"GetVersionInfo2");
```
NextTurn: Is this intentional?

leculver: Console.Write anywhere in the main library is never intentional.  I probably shouldn't be writing code with 3 hours sleep...

Fixed in https://github.com/microsoft/clrmd/pull/458.

PR #458, opened 2019-12-19
Memory searcher cleanups

leculver:
- Add todo marker where we need to improve memory searching in the future.
- Remove Console.WriteLines.
- Change SearchMemory to use a length.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/MemoryReaderTests.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/MemoryReaderTests.cs
@@ -43,11 +43,11 @@ namespace Microsoft.Diagnostics.Runtime.Tests
                         if (!dt.DataReader.Read(expectedOffset, slice, out int read) || read != slice.Length)
                             continue;
 
-                        ulong addressFound = dt.DataReader.SearchMemory(segment.Start, segment.End, buffer.Slice(0, i + sizeof(ulong)));
+                        ulong addressFound = dt.DataReader.SearchMemory(segment.Start, (int)segment.Length, buffer.Slice(0, i + sizeof(ulong)));
 
                         // There could still accidently be a pattern that matches this somewhere
                         while (addressFound != 0 && addressFound < expectedOffset)
-                            addressFound = dt.DataReader.SearchMemory(addressFound + 1, segment.End, buffer.Slice(0, i + sizeof(ulong)));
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Runtime.Tests/src/MemoryReaderTests.cs
```diff
@@ -43,11 +43,11 @@ public void SearchMemoryTest()
                         if (!dt.DataReader.Read(expectedOffset, slice, out int read) || read != slice.Length)
                             continue;
 
-                        ulong addressFound = dt.DataReader.SearchMemory(segment.Start, segment.End, buffer.Slice(0, i + sizeof(ulong)));
+                        ulong addressFound = dt.DataReader.SearchMemory(segment.Start, (int)segment.Length, buffer.Slice(0, i + sizeof(ulong)));
 
                         // There could still accidently be a pattern that matches this somewhere
                         while (addressFound != 0 && addressFound < expectedOffset)
-                            addressFound = dt.DataReader.SearchMemory(addressFound + 1, segment.End, buffer.Slice(0, i + sizeof(ulong)));
+                            addressFound = dt.DataReader.SearchMemory(addressFound + 1, (int)segment.End, buffer.Slice(0, i + sizeof(ulong)));
```
NextTurn: Arithmetic is needed here.

Well, from the uses of that method it does seem that accepting an end address is more natural... I have no idea why most APIs in the core library prefer a length though.

leculver: It's surprising that the test passed.  I must have gotten lucky with that underflow.

leculver: Taking a length makes a lot of sense when working with strings, arrays, and lists.  I'm happy to follow that convention here.  I just need more coffee to stop failing at basic coding.  =)

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
```diff
@@ -231,6 +232,8 @@ internal override unsafe bool GetFileVersion(string dll, out int major, out int
 
                 address += s_versionLength;
 
+                // TODO:  This should be cleaned up to not read byte by byte in the future.  Leaving it here
```
NextTurn: Oh that's a `StreamAddressSpace` wrapping a `FileStream`. Just use the `FileStream` with `Seek` and `Read` instead and eliminate the byte-to-byte reads.

PR #464, opened 2019-12-20
Fix MemorySearcher

NextTurn:
#462 

Not all memory pages are present in dump files. The coredump reader now fails to find versions of native modules because the first RW LOAD segment of the module isn't placed in the coredump and therefore the memory searcher stops on the first read operation.

See the original implementation:

https://github.com/microsoft/clrmd/blob/429070b994d605e46aca8483e83e3061ac67259d/src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs#L120-L125

Alternatively we may tell the searcher where to skip by exposing the memory page information from the readers, which can be obtained from ELF program headers or `/proc/<pid>/maps` on Linux or `VirtualQuery` on Windows, if this will bring large perf improvements. (ETW is currently broken on my machine so I'm hesitant about perf related changes without profiling)

Do you have any idea on how to determine the buffer size? The benchmark results don't show that using larger buffer size makes the search faster.

| BufferSize |     Mean |    Error |   StdDev |
|----------- |---------:|---------:|---------:|
|        0x2 | 37.37 ms | 1.692 ms | 3.007 ms |
|        0x4 | 36.26 ms | 1.212 ms | 2.058 ms |
|        0x6 | 18.16 ms | 0.507 ms | 0.901 ms |
|        0x8 | 18.53 ms | 0.431 ms | 0.720 ms |
|       0x10 | 17.84 ms | 0.515 ms | 0.915 ms |
|       0x20 | 17.21 ms | 0.408 ms | 0.693 ms |
|       0x40 | 16.58 ms | 0.479 ms | 0.814 ms |
|       0x60 | 15.54 ms | 0.377 ms | 0.669 ms |
|       0x80 | 15.77 ms | 0.259 ms | 0.461 ms |
|      0x100 | 16.03 ms | 0.470 ms | 0.836 ms |
|      0x200 | 16.55 ms | 0.407 ms | 0.692 ms |
|      0x300 | 17.97 ms | 0.579 ms | 1.029 ms |
|      0x400 | 18.10 ms | 0.522 ms | 0.914 ms |
|      0x500 | 18.61 ms | 0.598 ms | 1.062 ms |
|      0x600 | 18.64 ms | 0.532 ms | 0.946 ms |
|      0x700 | 19.67 ms | 0.506 ms | 0.899 ms |
|      0x800 | 21.19 ms | 1.136 ms | 2.020 ms |
|      0x900 | 20.08 ms | 0.666 ms | 1.167 ms |
|     0x1000 | 21.59 ms | 0.717 ms | 1.274 ms |

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Extensions/MemorySearcher.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Extensions/MemorySearcher.cs
@@ -49,13 +49,16 @@ namespace Microsoft.Diagnostics.Runtime
                     if (bytesRemaining < searchFor.Length)
                         break;
 
-                    Span<byte> buffer = new Span<byte>(array, 0, Math.Min(bytesRemaining, array.Length));
+                    Span<byte> buffer = array.AsSpan(0, Math.Min(bytesRemaining, array.Length));
                     if (!reader.Read(startAddress, buffer, out int read) || read < searchFor.Length)
-                        break;
+                    {
+                        startAddress += (uint)searchFor.Length;
+                        continue;
+                    }
 
                     buffer = buffer.Slice(0, read);
                     int index = buffer.IndexOf(searchFor);
-                    if (index != -1)
+                    if (index >= 0)
                         return startAddress + (uint)index;
 
                     // To keep the code simple, we'll re-read the last bit of the buffer each time instead of
@@ -63,7 +66,7 @@ namespace Microsoft.Diagnostics.Runtime
                     // along the way.
                     int increment = buffer.Length - searchFor.Length;
                     if (increment <= 0)
-                        break;
+                        increment = searchFor.Length;
 
                     startAddress += (uint)increment;
                 }


leculver:
Thanks for taking a look at this, I was going to work on it this morning.  Happy to see it's already fixed.  =)

This long post is mostly just me organizing my thoughts.

> Do you have any idea on how to determine the buffer size? The benchmark results don't show that using larger buffer size makes the search faster.

(Just to check, in these measurements you did use new[] and not ArrayPool's Rent, correct?  ArrayPool will round up your buffer size.  I assume so but just wanted to check.)

I have (mostly) avoided trying to do too much performance work until most of the 2.0 rewrite is completed.  Things like tuning buffer size for reads is something I would leave for later...only things which would drastically impact architecture is something I'd want to worry about now.

As to your specific question, historically we have observed that performance for sequential memory reads (such as for ClrHeap.EnumerateObjects) is GREATLY improved by using a larger (1024) buffer compared to incrementally reading pointer by pointer.  This observation came from using dbgeng as the data reader with SOS.  I'm still working from that assumption, but it's definitely something we should (eventually) test to see if it's still true.

If we find that a 0x400 buffer size is roughly equivalent to a 0x40 buffer size, I think the larger buffer size will win out in a multithreaded scenario because we'll reduce the number of times we have to take the lock...  However, that's something I'd want to test and tune, not something I'd assume is true.

> Not all memory pages are present in dump files.

It's interesting that you bring this up now, as I was just thinking about it yesterday and how to fix this for the library.  It boils down to the fact that ClrMD actually has a poorly thought out memory reading strategy:

1.  IMemoryReader/IDataReader is supposed to provide memory reads to ClrMD.  This memory isn't complete, as the contents of modules may not be in the dump.
2.  DacDataTargetWrapper accounts for this and reads from modules.
3.  Anything else in ClrMD that uses IMemoryReader/IDataReader directly and _doesn't_ account for the missing data...but it should.

I'll likely build a wrapper around IMemoryReader (or IDataReader) as part of fixing https://github.com/microsoft/clrmd/issues/465 which addresses this more generally.

NextTurn:
> (Just to check, in these measurements you did use new[] and not ArrayPool's Rent, correct? ArrayPool will round up your buffer size. I assume so but just wanted to check.)

No, I did use `ArrayPool.Rent`. At that moment I didn't remember how `ArrayPool` rounds up the buffer size, otherwise I wouldn't use so many parameters.

PR #468, opened 2019-12-20
Reimplement PEImage using S.R.PE.PEReader

NextTurn:
The problem is that we need the SOS DAC API to determine the file layout every time we create a `PEImage` from memory.

- `ElfLoadedImage.OpenAsPEImage` <- `CoreDumpReader.CreateModuleInfo`
- `ModuleInfo.GetPEImage`

Idea: The current CoreCLR implementation maps flat files with read-only permissions (`PAGE_READONLY`), and expanded files with permissions defined in PE sections (`SEC_IMAGE`). If there is at least one writable or executable (typically a .text section) page in a module's memory ranges, we can assume that it's an expanded image. See my last commit.


NextTurn:
I need the idea to be approved to move forward.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
```diff
@@ -49,11 +51,22 @@ public class ModuleInfo
 
             try
             {
-                return _image = new PEImage(new ReadVirtualStream(_dataReader, (long)ImageBase, FileSize), isVirtual: true);
+                _image = new PEImage(new ReadVirtualStream(_dataReader, (long)ImageBase, FileSize), isVirtual: true);
+                _ = _image.IsManaged;
```
leculver: What is the purpose of this line?

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdModule.cs
```diff
@@ -69,8 +69,8 @@ public ClrmdModule(ClrAppDomain parent, IModuleHelpers helpers, ulong addr)
             {
                 if (_pdb is null)
                 {
-                    using ReadVirtualStream stream = new ReadVirtualStream(_helpers.DataReader, (long)ImageBase, (long)(Size > 0 ? Size : 0x1000));
-                    PEImage pefile = new PEImage(stream, !IsFileLayout);
+                    using ReadVirtualStream stream = new ReadVirtualStream(_helpers.DataReader, (long)ImageBase, (long)(Size > 0 ? Size : int.MaxValue));
+                    using PEImage pefile = new PEImage(stream, !IsFileLayout);
```
leculver: Unrelated to your change, this should probably use `DataTarget.LoadPEImage`.  Can you make that change here?  Be sure not to dispose it.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -4,45 +4,35 @@
...
     {
         private const ushort ExpectedDosHeaderMagic = 0x5A4D;   // MZ
         private const int PESignatureOffsetLocation = 0x3C;
         private const uint ExpectedPESignature = 0x00004550;    // PE00
-        private const int ImageDataDirectoryCount = 15;
-        private const int ComDataDirectory = 14;
-        private const int DebugDataDirectory = 6;
 
-        private readonly bool _virt;
+        private readonly PEReader _reader = null!;
```
leculver: Maybe this should be a public property so others can directly use it if they want?

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdModule.cs
leculver: Also `lock (pefile)` as its used here.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
NextTurn: The third commit is a demonstration to show that this is a layout issue. By calling `IsManaged`, the `PEReader` will read all headers and calculate the metadata location, and throw if the layout is incorrect.

leculver: Ahh ok.

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdModule.cs
NextTurn: > this should probably use `DataTarget.LoadPEImage`

What? `DataTarget.LoadPEImage` only load files on disk.

NextTurn:
I didn't see any module with file layout on Windows. Can we assume layouts on Windows are always mapped/loaded?

NextTurn:
I didn't see any module with file layout on Windows. Can we assume layouts on Windows are always expanded?

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
NextTurn: Done.

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdModule.cs
leculver: Ah right.  I should probably unify those two paths (or at least refresh my understanding of why these are different), but that's not something for now.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
```diff
@@ -11,11 +11,14 @@ namespace Microsoft.Diagnostics.Runtime
...
     /// Provides information about loaded modules in a <see cref="DataTarget"/>.
     /// </summary>
+#pragma warning disable CA1001 // Types that own disposable fields should be disposable
     public class ModuleInfo
+#pragma warning restore CA1001 // Types that own disposable fields should be disposable
     {
         private readonly IDataReader _dataReader;
         private bool? _isManaged;
         private VersionInfo? _version;
+        internal bool _isVirtual;
```
leculver: If this is writable it should be a public property.  By making it an internal variable then someone couldn't (reasonably) write their own DataReader similar to LinuxLiveDataReader.

As an alternative, it looks like the only place we set this value is LinuxLiveDataReader?  Without having deeply read through this change, maybe it's possible to recreate ModuleInfo in LinuxLiveDataReader instead of updating the value?  I'm not sure if that's feasible, just brainstorming.

NextTurn:
> I'm not sure offhand.

I can simply open an issue in dotnet/runtime to ask about this...

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
NextTurn: > it looks like the only place we set this value is LinuxLiveDataReader?

In Linux part, yes. If we go this route on Windows then `LiveDataReader` and `DbgEngDataReader` may have similar behaviors.

> maybe it's possible to recreate ModuleInfo in LinuxLiveDataReader instead of updating the value?

I assume recording these values in a collection and creating `ModuleInfo`s in the end could have lower overheads than recreating.

NextTurn: > it looks like the only place we set this value is LinuxLiveDataReader?

In Linux part, yes. If we go this route on Windows then `LiveDataReader` and `DbgEngDataReader` may have similar behaviors.

> maybe it's possible to recreate ModuleInfo in LinuxLiveDataReader instead of updating the value?

I assume recording these values in a collection and creating `ModuleInfo`s in the end could have lower overheads than recreating.

leculver: > I assume recording these values in a collection and creating `ModuleInfo`s in the end could have lower overheads than recreating.

That seems like the best approach.

NextTurn: Then we'll arrive at a design with a private read-only field. In both cases, the users only see a parameter. However, by making the field internally writable, we have additional benefits of reducing recreating or other overheads in our internal implementation.

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -334,8 +338,7 @@ private int GetReadableBytesCount(ulong address, int bytesRequested)
             for (int i = startIndex; i < _memoryMapEntries.Count; i++)
             {
                 var entry = _memoryMapEntries[i];
-                if (!entry.IsReadable()         // the current region is not readable
-                    || entry.BeginAddr > endAddress
+                if (!entry.IsReadable                    || entry.BeginAddr > endAddress
```
NextTurn: Opened dotnet/roslyn#40758.

NextTurn:
Possibly layout-related PR: dotnet/coreclr# 10772

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
NextTurn: Can you comment on this? The internal field looks similar to that internal method of `ElfLoadedImage` (although it's not a public type). By making it internal we do make sure it's initialized only so `GetPEImage` is guaranteed to return the correct result. The users could achieve the same by recreating or other one-time creation approaches.

https://github.com/microsoft/clrmd/blob/daafcb3fa7426526237a5fecccee042a0d243996/src/Microsoft.Diagnostics.Runtime/src/Linux/ElfLoadedImage.cs#L50

I have no objection to making it public, just want to think more about it.

NextTurn:
The core dumps don't seem right. We are getting negative addresses.

leculver:
@NextTurn Can you update this with recent changes to the master branch?  I'd like to take a closer look at this and wrap it up.

I didn't want to force push an update to your branch.

leculver:
Also, please feel free to just fix the compile issues and not the other code review feedback.  I plan to merge this change as-is (with build issues fixed) and take a look at how I need to handle ModuleInfo.

Thanks for your patience here!

NextTurn:
> to just fix the compile issues and not the other code review feedback

Done.

leculver:
Thanks!  I will tackle this tomorrow morning.

leculver:
Or maybe tonight if all goes well...

PR #471, opened 2019-12-21
Collect code coverage on CI

NextTurn:
Adds basic code coverage collection on CI.

The Arcade SDK doesn't play well with coverage tools. This PR starts Coverlet to prepare assemblies before the tests run and stops it to calculate results after tests succeed.

There are three outstanding issues till now.

1. The paths in embedded PDBs of assemblies built with `DeterministicSourcePaths`(which is enabled by `ContinuousIntegrationBuild`) flag don't point to local source, so Coverlet can't tell which assembly to process. tonerdo/coverlet#363

2. Azure Pipelines [coverage board](https://dev.azure.com/dnceng/public/_build/results?buildId=465265&view=codecoverage-tab) doesn't merge reports from different jobs. Codecov works fine on this.

3. There isn't the try-finally block in MSBuild, so if any test failed... :)


NextTurn:
Demo report: https://dev.azure.com/dnceng/public/_build/results?buildId=464617&view=codecoverage-tab

NextTurn:
No coverage data is collected when `ContinuousIntegrationBuild` and `Build` are both set.

NextTurn:
No coverage data is collected for assemblies built with `ContinuousIntegrationBuild` flag.

NextTurn:
Finally ready for review.

leculver:
Any idea on the warnings produced?  https://dev.azure.com/dnceng/public/_build/results?buildId=465409&view=results

Do you plan to fix those after checkin?

NextTurn:
The "multiple file or directory matches were found" warnings are not problems. You can see that the ReportGenerator will merge multiple files and generate a single report per job. I plan to remove that task after integrating some other form of reports, such as Codecov.

The "failure in sending the provision message" warnings are not caused by this PR.

leculver:
I don't feel like I'm qualified to code review this, so let's merge it and see what happens. =)

I'm still on vacation until Monday.

PR #478, opened 2019-12-25
Clean up COM interfaces

NextTurn:


leculver:
I'd like to review this in depth so I may not get to it until I'm fully back.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/ImageResourceDirectoryEntry.cs
```diff
@@ -21,44 +22,42 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
     /// set to indicate this.  Otherwise the high bit is clear and the offset
     /// field points to a resource data entry.
     /// </summary>
-    internal struct IMAGE_RESOURCE_DIRECTORY_ENTRY
+    [StructLayout(LayoutKind.Sequential)]
```
leculver: Has this struct change been tested?  Is the layout before/after the `StructLayout` and field rearrange the same?

leculver: Ah these are the only fields, the rest are properties.

NextTurn: Struct layouts are sequential by default.

See Remarks in https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.structlayoutattribute

Issue #481, opened 2019-12-27
ClrObject with arrays


leculver:
All array related functions need to be moved out of `ClrObject` into a struct like `ClrArray`.  `ClrObject` needs new function `ClrArray AsArray()` that's only callable when `ClrObject.IsArray` returns true.

The new struct should cache the `.Length` property's result so that `ClrArray.Length` in a for loop is not a perf problem.

Additionally there needs to be a function to read the entire contents of the array at once, for example:  `T[] ReadContents<T>(int lengthLimit = -1) where T: unmanaged` and `string[] ReadStringContents(int lengthLimit = -1)`.

PR #482, opened 2019-12-28
Introduce ClrArray and API for reading multiple values from arrays

PhDuck:
This PR addresses issue #481 ClrObject with arrays.

Originally stemming from an optimization to decrease the amount of calls into dbgeng when loading out values of an array. Based on own testing this seems to perform ~4-5 times better. 

Originally code like this is needed: 
```
T[] arr = new T[item.Length];
for (int i = 0; i < arr.Length; i++)  
{  
    arr[i] = (T)item.Type.GetArrayElementValue(item.Address, i);  
}
```
Which can now be replaced with the following.

```
ClrArray arrItem = item.AsArray();
T[] arr = arrItem.GetContent(arrItem.Address, arrItem.Length);
```

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
@@ -71,6 +71,8 @@ namespace Microsoft.Diagnostics.Runtime
         /// <returns><c>true</c> if the <see cref="Address" /> of the parameter is same as <see cref="Address" /> in this instance; <c>false</c> otherwise.</returns>
         public override bool Equals(object? other)
         {
+            if (other is null)
+                return false;
             if (other is ulong ul)
                 return ul == Address;
             if (other is ClrArray clrArray)
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
@@ -69,7 +69,17 @@ namespace Microsoft.Diagnostics.Runtime
         /// </summary>
         /// <param name="other">The <see cref="ClrArray" /> to compare to this instance.</param>
         /// <returns><c>true</c> if the <see cref="Address" /> of the parameter is same as <see cref="Address" /> in this instance; <c>false</c> otherwise.</returns>
-        public override bool Equals(object? other) => other is ClrArray clrArray && this.Address == clrArray.Address;
+        public override bool Equals(object? other)
+        {
+            if (other is ulong ul)
+                return ul == Address;
+            if (other is ClrArray clrArray)
+                return Address == clrArray.Address;
+            if (other is ClrObject clrObject)
+                return Address == clrObject.Address;
+
+            return false;
+        }
 
         /// <summary>
         /// Determines if this instance and another specific <see cref="ClrArray" /> have the same value.
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
@@ -1,0 +1,114 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    /// <summary>
+    /// Represents an array in the target process.
+    /// </summary>
+    public struct ClrArray : IEquatable<ClrArray>, IEquatable<ClrObject>
+    {
+        /// <summary>
+        /// The address of the object.
+        /// </summary>
+        public ulong Address { get; }
+
+        /// <summary>
+        /// The type of the object.
+        /// </summary>
+        public ClrType Type { get; }
+
+        private int length;
+
+        /// <summary>
+        /// Returns the count of elements in this array.
+        /// </summary>
+        public int Length
+        {
+            get
+            {
+                if (this.length == -1)
+                {
+                    this.length = this.Type.ClrObjectHelpers.DataReader.Read<int>(this.Address + (ulong)IntPtr.Size);
+                }
+
+                return this.length;
+            }
+        }
+
+        public ClrArray(ulong address, ClrType type)
+        {
+            this.Address = address;
+            this.Type = type;
+            // default uninitialized value for size. Will be lazy loaded.
+            this.length = -1;
+        }
+
+        /// <summary>
+        /// Gets <paramref name="count"/> value elements from the array.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="count"></param>
+        /// <returns></returns>
+        public T[]? GetContent<T>(int count) where T : unmanaged
+        {
+            if (count > this.Length)
+            {
+                throw new ArgumentException("Cannot read more elements than elements in array.");
+            }
+
+            return this.Type.GetArrayElementsValues<T>(this.Address, count);
+        }
+
+        /// <summary>
+        /// Determines if this instance and another specific <see cref="ClrArray" /> have the same value.
+        /// <para>Instances are considered equal when they have same <see cref="Address" />.</para>
+        /// </summary>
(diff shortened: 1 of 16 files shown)


PhDuck:
@NextTurn 
> Copying in `ReadArray<T>` methods needs to be avoided.

Sounds like a good idea, what "copy" are your referring to?

NextTurn:
> Sounds like a good idea, what "copy" are your referring to?

From `buffer` to `values` result. I'd rather allocate an array and write into it directly.

PhDuck:
@NextTurn 
> > Sounds like a good idea, what "copy" are your referring to?
> 
> From `buffer` to `values` result. I'd rather allocate an array and write into it directly.

How would you do that? Doesn't the underlying function we call into (ReadProcessMemory) only support returning bytes? Looking at the other functions Read<T> does exactly the same as I do.

```
public unsafe bool Read<T>(ulong address, out T value) where T : unmanaged
{
    Span<byte> buffer = stackalloc byte[sizeof(T)];
    if (Read(address, buffer, out int size) && size == sizeof(T))
    {
        value = Unsafe.As<byte, T>(ref MemoryMarshal.GetReference(buffer));
        return true;
    }

    value = default;
    return false;
}
```

NextTurn:
> Doesn't the underlying function we call into (ReadProcessMemory) only support returning bytes?

The `Read` functions accept `Span<byte>` parameters. You could use `MemoryMarshal.Cast` to cast spans of other types to byte spans.

> Looking at the other functions Read does exactly the same as I do.

It's a little different. The size of the whole array may be sometimes large, so `stackalloc` is not preferred here.

PhDuck:
>The Read methods accept Span<byte> parameters. You could use MemoryMarshal.Cast to cast spans of other types to byte spans.

I didn't even know you could do that; how would you feel about this @NextTurn:

```
public unsafe bool ReadArray<T>(ulong addr, T[] values) where T : unmanaged
{
    Span<byte> buffer = MemoryMarshal.Cast<T, Byte>(values);

    if (Read(addr, buffer, out int size) && size == buffer.Length)
    {
        return true;
    }

    return false;
}
```

With this we can also get rid of the T[]**?**. Then I will allocate the T[] in ValueReader like this.
```
internal static T[] GetValuesFromAddress<T>(IDataReader reader, ulong addr, int count) where T : unmanaged
{
    var values = new T[count];

    reader.ReadArray(addr, values);

    return values;
}
```

NextTurn:
Looks good. You may still want to check the return value in `GetValuesFromAddress`.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/IMemoryReader.cs
```diff
@@ -41,6 +41,16 @@ public interface IMemoryReader
...
 
+        /// <summary>
+        /// Reads <c>values.Length</c> unmanaged values from the given address into <paramref name="values"/>.
+        /// </summary>
+        /// <typeparam name="T">The type to read.  This may be a struct or primitive type as long as it does
+        /// not managed pointeres.</typeparam>
+        /// <param name="address">The address to read from.</param>
+        /// <param name="values">An array the read values will be written into.</param>
+        /// <returns><see langword="true"/> if <c>values.Length</c> were read; else <see langword="false"/>.</returns>
+        bool ReadArray<T>(ulong address, T[] values) where T : unmanaged;
```
leculver: I don't want to add this to IMemoryReader.  We can accomplish this with just reading `Span<byte>` of the right size.

I try to keep `IMemoryReader` and `IDataReader` as simple as possible (within reason) so that we don't have to implement a lot across a lot of implementations.

Instead it would make sense to just do the `MemoryMarshal.Cast` on the caller site wherever its needed.

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdArrayType.cs
```diff
@@ -86,5 +86,30 @@ public override ulong GetArrayElementAddress(ulong objRef, int index)
 
             return ValueReader.GetValueAtAddress(Heap, DataReader, cet, addr);
         }
+
+        public override T[]? GetArrayElementsValues<T>(ulong objRef, int count)
+        {
+            ulong addr = GetArrayElementAddress(objRef, 0);
```
leculver: Check if `addr == 0` and return null here as an early out.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -0,0 +1,114 @@
...
+        private int length;
+
+        /// <summary>
+        /// Returns the count of elements in this array.
+        /// </summary>
+        public int Length
+        {
+            get
+            {
+                if (this.length == -1)
```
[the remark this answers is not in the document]
leculver: And the code everywhere its been added.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// The address of the object.
+        /// </summary>
+        public ulong Address { get; }
+
+        /// <summary>
+        /// The type of the object.
+        /// </summary>
+        public ClrType Type { get; }
+
+        private int length;
```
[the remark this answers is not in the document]
PhDuck: Sure thing.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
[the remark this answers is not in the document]
PhDuck: Sure thing.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -0,0 +1,114 @@
...
+            return this.Type.GetArrayElementsValues<T>(this.Address, count);
+        }
+
+        /// <summary>
+        /// Determines if this instance and another specific <see cref="ClrArray" /> have the same value.
+        /// <para>Instances are considered equal when they have same <see cref="Address" />.</para>
+        /// </summary>
+        /// <param name="other">The <see cref="ClrArray" /> to compare to this instance.</param>
+        /// <returns><c>true</c> if the <see cref="Address" /> of the parameter is same as <see cref="Address" /> in this instance; <c>false</c> otherwise.</returns>
+        public override bool Equals(object? other) => other is ClrArray clrArray && this.Address == clrArray.Address;
```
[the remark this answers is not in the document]
PhDuck: Sounds good

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdArrayType.cs
PhDuck: Yeah, sounds reasonable.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/IMemoryReader.cs
PhDuck: Good idea, when I first wrote it, I didn't actually really know the IMemoryReader interface, so I was just writing the code down there.
Good observation that it can be accomplished without adding new methods.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        /// <summary>
+        /// Determines if this instance and another specific <see cref="ClrArray" /> have the same value.
+        /// <para>Instances are considered equal when they have same <see cref="Address" />.</para>
+        /// </summary>
+        /// <param name="other">The <see cref="ClrArray" /> to compare to this instance.</param>
+        /// <returns><c>true</c> if the <see cref="Address" /> of the parameter is same as <see cref="Address" /> in this instance; <c>false</c> otherwise.</returns>
+        public override bool Equals(object? other)
+        {
```
NextTurn:
```suggestion
        {
            if (other is null)
                return false;
```

An early out for null may be better.

PhDuck: Yeah, sure.

PR #491, opened 2020-01-01
Add ClrArray.Rank

### LINKED ISSUE #488: Support for multi dimensional arrays
ClrMD should handle multi-dimensional arrays, such as `string[,,]`.  At minimum we should expose a property like `ClrArray.Rank` to tell you whether it's an szarray or not.  Multi-dim arrays are not a common feature, so I'm not sure how much effort I want to spend on them.  Especially if there's not a lot of dac support for it.

NextTurn:
Contributes to #488


leculver:
LGTM, clever solution.

PR #500, opened 2020-01-04
Install StyleCop.Analyzers

NextTurn:
You probably won't like many rules enabled by default.


NextTurn:
Some rules aren't working with new C# language features. They need to be fixed in StyleCop repo first.

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdHeap.cs
```diff
@@ -173,11 +171,12 @@ private IReadOnlyList<ClrSegment> Initialize()
         {
             lock (_sync)
             {
-                _helpers.CreateSegments(this,
-                                        out IReadOnlyList<ClrSegment> segments,
-                                        out IReadOnlyList<AllocationContext> allocContext,
-                                        out IReadOnlyList<FinalizerQueueSegment> fqRoots,
-                                        out IReadOnlyList<FinalizerQueueSegment> fqObjects);
+                _helpers.CreateSegments(
```
leculver: I dislike this refactor, is there a way to turn this one off?

leculver:
The mechanical changes overall look good so far, minus one question.

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdHeap.cs
NextTurn: Aligning by opening parentheses will result in inconsistencies...

src/Microsoft.Diagnostics.Runtime.Utilities/Microsoft.Diagnostics.Runtime.Utilities.csproj
```diff
@@ -1,25 +1,11 @@
...
-    <Description>.</Description>
-    <PackageReleaseNotes></PackageReleaseNotes>
-    <PackageTags></PackageTags>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <CheckForOverflowUnderflow>false</CheckForOverflowUnderflow>
-    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)\..\..\eng\Microsoft.Diagnostics.Runtime.snk</AssemblyOriginatorKeyFile>
-    <RepositoryUrl>https://github.com/Microsoft/clrmd</RepositoryUrl>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net471|AnyCPU'">
-    <DefineConstants>TRACE;DEBUG;NET45</DefineConstants>
```
leculver: Does this remove the DEBUG constant or is that implied somehow?

src/Microsoft.Diagnostics.Runtime.Tests/Microsoft.Diagnostics.Runtime.Tests.csproj
```diff
@@ -1,8 +1,7 @@
-﻿<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.0</TargetFrameworks>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <IsPackable>false</IsPackable>
```
leculver: Why is this set to false?

I don't have any objection if there's reason, just wanted to understand here.

Directory.Build.props
```diff
@@ -7,13 +7,10 @@
   <PropertyGroup>
     <LangVersion>Latest</LangVersion>
     <Features>strict</Features>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-    <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
```
leculver: This was intentional.  We do not want overflow/underflow to interfere with the library but I did want actual overflow issues to surface in test runs or when debugging the library.  Was this changed on purpose?

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdHeap.cs
leculver: I suppose that's fine, but it's going to be a hard habit to break after 20 years of doing it the other way.  =)

src/Microsoft.Diagnostics.Runtime.Utilities/Microsoft.Diagnostics.Runtime.Utilities.csproj
NextTurn: It's implied by the `Debug` configuration.

src/Microsoft.Diagnostics.Runtime.Tests/Microsoft.Diagnostics.Runtime.Tests.csproj
NextTurn: I saw this is set to false for UnitTests.csproj so I synced them.

Directory.Build.props
NextTurn: I'll revert this.

NextTurn:
Functionality changes are unintentional. If there is anything beyond formatting (like the overflow/underflow one) please do leave a comment.

Directory.Build.props
NextTurn: The current configurations override this property to always false in non-test libraries.

NextTurn: Enabling overflow checks broke a bunch of tests 😅. That's a real functionality change which should be saved for later.

src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdHeap.cs
NextTurn: Also, is it a habit to spell out .NET as .Net? It should be all capitalized.

[.NET Glossary](https://github.com/dotnet/docs/blob/master/docs/standard/glossary.md#net)

[Microsoft Style Guide](https://github.com/MicrosoftDocs/microsoft-style-guide/blob/master/styleguide/a-z-word-list-term-collections/n/net.md)

Directory.Build.props
leculver: haha that's pretty interesting...  I should look into that at some point.

PR #505, opened 2020-01-06
Improve ClrArray, add multi-dimensional APIs

### LINKED ISSUE #488: Support for multi dimensional arrays

NextTurn:
Closes #488

Please review the API shapes before I start working on them.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
@@ -127,6 +127,22 @@ namespace Microsoft.Diagnostics.Runtime
         /// <returns>An <see cref="int"/> hash code for this instance.</returns>
         public override int GetHashCode() => Address.GetHashCode();
 
+        public int GetLowerBound(int dimension) => throw new NotImplementedException(nameof(GetLowerBound) + nameof(dimension));
+
+        public int GetUpperBound(int dimension) => throw new NotImplementedException(nameof(GetUpperBound) + nameof(dimension));
+
+        // -or- GetValue
+        public T ReadValue<T>(int index) where T : unmanaged => throw new NotImplementedException(nameof(ReadValue) + nameof(index));
+
+        // -or- GetValue
+        public T ReadValue<T>(params int[] indices) where T : unmanaged => throw new NotImplementedException(nameof(ReadValue) + nameof(indices));
+
+        // -or- ReadObjectValue
+        public ClrObject ReadValueAsObject(int index) => throw new NotImplementedException(nameof(ReadValueAsObject) + nameof(index));
+
+        // -or- ReadObjectValue
+        public ClrObject ReadValueAsObject(params int[] indices) => throw new NotImplementedException(nameof(ReadValueAsObject) + nameof(indices));
+
         /// <summary>
         /// Determines whether two specified <see cref="ClrArray"/> have the same value.
         /// </summary>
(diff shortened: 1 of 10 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -111,6 +111,22 @@ public ClrArray(ulong address, ClrType type)
         /// <returns>An <see cref="int"/> hash code for this instance.</returns>
         public override int GetHashCode() => Address.GetHashCode();
 
+        public int GetLowerBound(int dimension) => throw new NotImplementedException(nameof(this.GetLowerBound));
+
+        public int GetUpperBound(int dimension) => throw new NotImplementedException(nameof(this.GetUpperBound));
+
+        // -or- GetValue
+        public T ReadValue<T>(int index) where T : unmanaged => throw new NotImplementedException(nameof(this.ReadValue));
```
leculver: I'm not sure if `Get` or `Read` is more clear.  For now I think using `GetValue` is right to be consistent with `ClrObject`, but if we decide to change it to Read later we should update `GetField` and similar as well.

leculver:
This surface area seems good to me.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -127,6 +117,139 @@ public int GetLength(int dimension)
...
+            }
+
+            int valueByteOffset = dataByteOffset + valueOffset * Type.ComponentSize;
+            return Type.ClrObjectHelpers.DataReader.Read<T>(Address + (ulong)valueByteOffset);
+        }
+
+        public ClrObject GetObjectValue(int index)
+        {
+            ulong address = GetValue<UIntPtr>(index).ToUInt64();
+            if (!Type.IsObjectReference)
```
leculver: nit:  Maybe this should go above the read?

leculver: Same for GetObjectValue.  It seems like we should throw an InvalidOperationException if the type isn't an object reference unconditionally.  It's possible that GetValue might throw a different exception depending on the underlying IMemoryReader implementation, so we should throw the InvalidOperationException as a priority.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -127,6 +117,139 @@ public int GetLength(int dimension)
...
+                throw new InvalidOperationException();
+
+            return Type.Heap.GetObject(address);
+        }
+
+        public ClrObject GetObjectValue(params int[] indices)
+        {
+            ulong address = GetValue<UIntPtr>(indices).ToUInt64();
+            if (!Type.IsObjectReference)
+                throw new InvalidOperationException();
```
leculver: `throw new InvalidOperationException();` maybe this should be `throw new InvalidOperationException($"{Type} does not contain object references.");` or something similarly more descriptive.

NextTurn:
> Still a draft?

I have some additional changes.

NextTurn:
Do we want to check `sizeof(T) == Type.ComponentSize`?

leculver:
> Do we want to check `sizeof(T) == Type.ComponentSize`?

Yes, since we are checking against the length/bounds of the array.  It prevents a useful case where you want to read it all as a byte array or something, but we could (eventually) provide other mechanisms to get that.

PR #507, opened 2020-01-09
Fix overflow

NextTurn:


src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
```diff
@@ -25,12 +25,12 @@ public class ModuleInfo
         /// <summary>
         /// Gets the file size of the image.
         /// </summary>
-        public uint FileSize { get; }
+        public int FileSize { get; }
```
leculver: I've been really unsure whether to make these `int` or `uint` in 2.0.  Technically these fields _are_ defined as `uint` in their , but we really only use them as a key and never do any math with the fields, so I think this is fine, especially if it fixes checked/unchecked problems.

NextTurn: They're defined as `int` in core libraries as well.

https://github.com/dotnet/runtime/blob/2c54bd95ab18962fa6a2c51ef9871d8b6f57d91b/src/libraries/System.Reflection.Metadata/src/System/Reflection/PortableExecutable/CoffHeader.cs#L19-L22

https://github.com/dotnet/runtime/blob/2c54bd95ab18962fa6a2c51ef9871d8b6f57d91b/src/libraries/System.Reflection.Metadata/src/System/Reflection/PortableExecutable/PEHeader.cs#L115-L119

leculver: That works for me then.  No need to do anything different.

PR #508, opened 2020-01-10
Add Encoding.GetString extension

NextTurn:


NextTurn:
Unrelated: How about adding a `netcorecpp3.0` TFM and replacing these reflection calls?

https://github.com/microsoft/clrmd/blob/f3c50ff41e49aeb9160cf0d8c17ca8dcf46200ff/src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs#L31

leculver:
> Unrelated: How about adding a netcorecpp3.0 TFM and replacing these reflection calls?

Yeah we could do that.

PR #510, opened 2020-01-12
Add weak WinRT handle kind

NextTurn:
To reproduce: follow the steps referenced by dotnet/coreclr#27860 and call `ClrmdRuntime.EnumerateHandles()` when the process crashes.

https://github.com/dotnet/runtime/blob/afc21e51af60a4d2cb11a1081fb38b1e3a8df0a2/src/coreclr/src/gc/gcinterface.h#L424-L436

https://github.com/dotnet/runtime/blob/afc21e51af60a4d2cb11a1081fb38b1e3a8df0a2/src/coreclr/src/debug/daccess/request.cpp#L3255-L3257

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHandleKind.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHandleKind.cs
@@ -47,6 +47,8 @@ namespace Microsoft.Diagnostics.Runtime
         /// <summary>
         /// Strong handle used internally for book keeping.
         /// </summary>
-        SizedRef = 8
+        SizedRef = 8,
+
+        WeakWinRT = 9,
     }
 }
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHandleKind.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHandleKind.cs
@@ -49,6 +49,9 @@ namespace Microsoft.Diagnostics.Runtime
         /// </summary>
         SizedRef = 8,
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ClrHandleKind.cs
```diff
@@ -47,6 +47,8 @@ public enum ClrHandleKind
         /// <summary>
         /// Strong handle used internally for book keeping.
         /// </summary>
-        SizedRef = 8
+        SizedRef = 8,
+
+        WeakWinRT = 9,
```
leculver: Need a comment on this briefly describing what it is.

NextTurn:
> Since they are weak there's not much ClrMD does with them.

There is extra information in the "secondary" field, but I don't think we need to add a type for it now.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrHandleKind.cs
NextTurn: A WeakWinRT handle is a weak WinRT handle, that's all... How could you expect the users to understand the following:

> WinRT weak reference handles hold two different types of weak handles to any RCW with an underlying COM object that implements IWeakReferenceSource.  The object reference itself is a short weak handle to the RCW...

?

leculver: Just having `/// Weak WinRT handle.` is enough here.  I was only objecting to not having a comment on the enum at all because it can cause a build warning with some settings.  We don't have to fully explain what it is here.

PR #511, opened 2020-01-12
Update handle names

NextTurn:
- What does these `@` characters mean?

https://github.com/microsoft/clrmd/blob/f3c50ff41e49aeb9160cf0d8c17ca8dcf46200ff/src/Microsoft.Diagnostics.Runtime/src/Common/ClrHandle.cs#L49

- Don't you feel these names (#318) are more readable?

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/GCRootTests.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/GCRootTests.cs
@@ -190,7 +190,7 @@ namespace Microsoft.Diagnostics.Runtime.Tests
 
             foreach (GCRootPath rootPath in rootPaths)
             {
-                if (rootPath.Root.RootKind == ClrRootKind.PinningHandle)
+                if (rootPath.Root.RootKind == ClrRootKind.PinnedHandle)
                     hasStatic = true;
                 else if (rootPath.Root.RootKind == ClrRootKind.Stack)
                     hasThread = true;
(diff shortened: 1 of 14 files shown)


leculver:
> * What does these `@` characters mean?

It's mostly just a way to distinguish that the `@address` is a slot which points to an object, and that the `@address` itself is not an address of an object itself.  Having two addresses next to each other as in `$"{address:x} {clrObject}"` was very confusing when I was debugging some things.

> * Don't you feel these names (#318) are more readable?

Yes those are definitely better.  Removing those names was not intentional, it happened when I was refactoring the entire runtime and didn't notice I removed that.  I'd be happy to bring those back either myself or feel free to wrap them into this PR.

(FYI, I am snowed in where I live and only working part time for this week.  Responses will be a little slow, but if I miss something feel free to `@leculver` to ping me.  I should be back to work as normal on 1/21.)

NextTurn:
>  Having two addresses next to each other as in `$"{address:x} {clrObject}"` was very confusing when I was debugging some things.

I'd suggest adding an arrow between them.

Generally if something is exclusive for debugging you could add `DebuggerDisplay` (#437) or `DebuggerTypeProxy` attributes.

> but if I miss something feel free to `@leculver` to ping me.

I believe you're watching this repo so you'll be notified of any messages even if I don't mention (@) you?

NextTurn:
Is "finalization handle" really a "handle"?

leculver:
> I'd suggest adding an arrow between them.

An arrow is acceptable.

> Is "finalization handle" really a "handle"?

That term came from old CLR a while back.  They are conceptually a handle, but no they aren't a real handle.  The term can be changed/removed here.

NextTurn:
> The term can be changed/removed here.

What to replace? "Finalization root"/"Finalizer root"/"Finalizer queue/thread root"?

NextTurn:
> The term can be changed/removed here.

What to replace with? "Finalization root"/"Finalizer root"/"Finalizer queue/thread root"?

leculver:
Probably "Finalization root"

NextTurn:
Ha, I see that class is named `ClrFinalizerRoot`.

NextTurn:
Done. Either this or #510 shall be merged first so I can add name for the new handle kind.

PR #512, opened 2020-01-12
Don't throw NullReferenceException

NextTurn:
[Handling and throwing exceptions in .NET](https://docs.microsoft.com/dotnet/standard/exceptions/index)

Those exceptions should only be thrown by the runtime.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrObject.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrObject.cs
@@ -158,11 +158,12 @@ namespace Microsoft.Diagnostics.Runtime
         public static implicit operator ulong(ClrObject clrObject) => clrObject.Address;
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ClrObject.cs
```diff
@@ -284,7 +290,7 @@ private ulong GetFieldAddress(string fieldName, ClrElementType element, out ClrT
             ClrType type = GetTypeOrThrow();
 
             if (IsNull)
-                throw new NullReferenceException();
+                throw new InvalidOperationException();
```
leculver: This should also have an "Object is null" or similar exception message since we are throwing a more generic exception.  Maybe introduce a constant string for the class which is used in all places we would have thrown a NullReferenceException?

leculver: I think I want to clean these up myself, there's more than just this one.

PR #514, opened 2020-01-13
Improve Linux readers

NextTurn:
I'm not satisfied with the implementation, but it's a little better.

What's your plan to rewrite the readers with `MemoryMappedFile`?


leculver:
> What's your plan to rewrite the readers with `MemoryMappedFile`?

It's not something I want to implement right now.  I have a vague plan around using memorymappedfile to keep chunks of the GC heap open in memory to read so that parallel operations can operate faster and with less locking.  After some review with a coworker, I concluded this is mostly just an optimization for me and there's far more important things that I want to tackle first.

PR #515, opened 2020-01-14
Add .NET Core 3.1 TFM

NextTurn:
Only one warning left.

Do we want to ensure non-null return here?

https://github.com/microsoft/clrmd/blob/fb1fa753c54e35e3468612d8b6f27843c70619f3/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs#L192-L196

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.101"
+    "dotnet": "3.1.100"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "5.0.0-beta.20055.1"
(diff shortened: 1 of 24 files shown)


leculver:
(The power and internet is going on and off at my house, so I may not be able to respond further today, depending on the weather situation.)

> Do we want to ensure non-null return here?

Interesting, it's surprising the compiler missed that before?  That function is intended to return `ClrModule?` also the method needs to lock _modules.  Can you replace that function with this:

        private ClrModule? GetModule(ulong addr)
        {
            lock (_modules)
            {
                _modules.TryGetValue(addr, out ClrModule module);
                return module;
            }
        }

Then update everything that calls it to accept `ClrModule?` (including the constructor for `ClrmdArrayType`).  That should resolve the problem and fix the concurrency bug of not locking _modules when reading from it.

NextTurn:
Even in good weather my network was down for a whole day.

> Interesting, it's surprising the compiler missed that before?

.NET Core 2.1 doesn't have the nullable attributes (including `NotNullWhen` which is causing problems now).

NextTurn:
I've submitted dotnet/roslyn-analyzers#3208 to track the magic CA2000 false positive issue. Let's suppress this warning for now.

PR #517, opened 2020-01-16
Add .NET Standard 2.0 to the list of TFMs

adamsitnik:
Yesterday I've tried to update BenchmarkDotNet to use the latest ClrMD and it turned out that the .NET Standard 2.0 was removed.

@leculver could you please restore it? It's a must-have thing for us (BenchmarkDotNet). We are still supporting .NET 4.6.1 (I wish I could stop, but we have too many users who for some reasons can't move to 4.7.2)

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
+++ b/src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
@@ -26,10 +26,6 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <!-- <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.113">
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-      <PrivateAssets>all</PrivateAssets>
-    </PackageReference> -->
     <PackageReference Include="System.Buffers" Version="4.5.0" />
     <PackageReference Include="System.Collections.Immutable" Version="1.7.0" />
     <PackageReference Include="System.Memory" Version="4.5.3" />
--- a/src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
+++ b/src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net471</TargetFrameworks>
     <NoWarn>1591;1701</NoWarn>
     <Nullable>enable</Nullable>


leculver:
I am on vacation and will take a much harder look at the request when I'm back on Tuesday.  Also, please keep in mind the current version of ClrMD is pre-release and may have some API breaking changes before it comes out of beta.

The overall refactor of ClrMD for 2.0 is focused on setting up the product for the next 5 years.  I've made a conscious choice to rebaseline the library at 4.7.1 because once it ships it'll be a huge breaking change to move the library forward from ns2.0 (or 4.6.1) in the future so I'll basically be stuck on 4.6.1 forever for 2.0 if I change this back to ns2.0.  I'd like to take a very close look at the API surface differences in the BCL before adding back ns2.0 to make sure I'll never need them in the future.  Since the library currently builds for 4.6.1 it's not unreasonable, but I'd like to take a much closer look before making a decision.

Also please note that I fully intend to continue supporting ClrMD 1.1 for the next few years, [see the release notes](https://github.com/microsoft/clrmd/blob/master/doc/ReleaseNotes2.0.md#quick-notes).  Even if we can't accommodate the request to move backwards, your product will still continue to work on 1.1 and receive bugfixes.

leculver:
Sorry for the (very long) delay.  I've been putting off making a decision on this for a while to see if we are actually going to make use of some updated libraries in desktop as the project gets more mature.  I didn't want to dial the version back until things were in a more complete state, but it turns out we aren't using anything out of the ordinary.  The stuff I thought was added between those two versions actually came from a separate NuGet package.

I'm closing this PR and instead fixing it this way:  https://github.com/microsoft/clrmd/pull/612.

adamsitnik:
> I'm closing this PR and instead fixing it this way

The problem is that when a library targets .NET Core and .NET it can not be referenced by a .NET Standard library. The official guidance is that targeting .NET Standard should be the default for libraries while targeting specific runtime is the default for apps.

IMHO as long as you plan to support full .NET Framework, adding `.NET Standard 2.0` to the list of supported TFMs should not be a problem because you still can't use .NET Core specific APIs in the codebase and even if you provide some workarounds for missing APIs the Full .NET Framework solution is going to work for .NET Standard as well. So all the `#if` defs could be defined as following:

```cs
#if NETCOREAPP
// .NET Core
#else
// .NET and .NET Standard
#endif
```

@leculver could you please reconsinder adding the .NET Standard 2.0 support? It would allow BenchmarkDotNet to always use the latest and greatest version of `ClrMD`, which we use for some of our features like disassembler to work.

leculver:
Blah.  Right.  I forgot about that.  What a mess.

Though it sucks if someone pulls in the netstandard2.0 version instead of the .Net Core version by accident (or transitive closure of including a library) because perf on the netstandard version will be much worse.

Why does BenchmarkDotNet target netstandard if it uses ClrMD?  Do you target Mono/Unity at all for benchmarks, just without using ClrMD?

leculver:
Let's chat via teams if you are around, it will be more efficient than github messages.

weltkante:
would it make sense to add a msbuild target to the nuget package which checks if the netstandard package gets pulled into a .NET Core target framework build and emit a warning in that combination? presumably the warning could be resolved by the app referencing clrmd directly?

just throwing that into the discussion in case it helps

leculver:
@weltkante That would alleviate some of my concerns, but creating msbuild targets is not my forte.

The real problem is that for ClrMD (and primarily clrmd based tools), netstandard isn't an interesting concept.

ClrMD doesn't support Unity or Mono (the only places other than Microsoft runtimes that netstandard targets), so there's no reason to target netstandard in ClrMD to support those.  For libraries that use ClrMD, targeting netstandard isn't helpful for the same reason, unless the underlying library is more generally useful to mono/unity (and scopes its ClrMD usage only in the case where we are talking to desktop clr and/or .net core).

I _think_ there's also the chance we will load two different copies (or maybe the wrong copy) of clrmd if a library FOO targets netstandard2.0, then an application loads library FOO and ClrMD (netstandard3.1).  What happens in that case?  I need to ask some experts to find out.

I don't want to completely remove ClrMD's usefulness from netstandard libraries, so finding some way to mitigate the issue would be ideal, an msbuild warning would be nice.  I'm still working on this.

NextTurn:
Nuget will automatically choose the closest package. ClrMD for .NET Core will be referenced in a .NET Core project which only references BenchmarkDotNet.

Given the dependency chain

`core-only` -> `standard-only` -> `both-standard-and-core`

note the build order

```
> dotnet build standard-only\

  both-standard-and-core -> netstandard2.0\both-standard-and-core.dll
  standard-only -> netstandard2.0\standard-only.dll

> dotnet build core-only\

  both-standard-and-core -> netstandard2.0\both-standard-and-core.dll
  standard-only -> netstandard2.0\standard-only.dll
  both-standard-and-core -> netcoreapp3.1\both-standard-and-core.dll
  core-only -> netcoreapp3.1\core-only.dll
```

PR #522, opened 2020-01-18
Simplify DbgEng interfaces

NextTurn:
Remove attributes implied by parameter modifiers.


NextTurn:
For the last commit: `IDebugEventContextCallbacks` APIs are not documented. If `uint ContextSize` is the size of the struct type then `[In] DEBUG_EVENT_CONTEXT* Context` could be changed to an `in` parameter.

leculver:
I'm not sure on `IDebugEventContextCallbacks`.  The rest of the changes look good to me, so I'm merging at this point.

weltkante:
It seems to be somewhat documented in the [header via comments](https://github.com/tpn/winsdk-10/blob/master/Debuggers/inc/dbgeng.h#L17383). According to the header annotations (`_In_reads_bytes_(ContextSize) PVOID Context`) the ContextSize should indeed be the size of the pointed-to data.

Issue #532, opened 2020-01-28
Assertion failure for interior stack refs


NextTurn:
https://github.com/microsoft/clrmd/blob/2e00bfb1a067a07ec6793c0dc0d222a982adc695/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs#L256

This branch always triggers the assertion.

https://github.com/microsoft/clrmd/blob/2e00bfb1a067a07ec6793c0dc0d222a982adc695/src/Microsoft.Diagnostics.Runtime/src/Common/ClrObject.cs#L29

PR #533, opened 2020-01-28
Ensure valid ClrObject for by-ref stack refs

NextTurn:
Partial fix for #532 

||Ref to Reference Type on Heap|Ref to Reference Type on Stack|Ref to Value Type on Heap/Stack|
|-|-|-|-|
|**Before**|handled by DAC|invalid `ClrObject`|invalid `ClrObject`|
|**After**|handled by DAC|pointer indirection|skipped|


leculver:
Thanks for this pull request!  I've gone a different direction for this fix, using your test cases: https://github.com/microsoft/clrmd/pull/560.

This _should_ handle all relevant cases correctly while not losing any valid gc roots for objects.  It makes things a little more complicated (requiring a cast) if you want to directly interact with stack roots though.

PR #551, opened 2020-03-09
Remove System.Array from GetOrCreateBasicType

### LINKED ISSUE #506: Error handling element types for array types
See https://github.com/microsoft/clrmd/pull/505/commits/6b2669262b4c4560275d592ee9f4ff7828257a21

I haven't checked if changes are needed elsewhere.

NextTurn:
Fixes #506 

There is no "basic type" for `System.Array`.

`System.String` is missing here - not sure if it's intentional.


leculver:
> `System.String` is missing here - not sure if it's intentional.

We should always have System.String, it's enumerated via `GetCommonMethodTables` (or whatever that function is called).  Even if the runtime is barely started we should never fail to find that type, so I think we'd never fail (and therefore need it) in GetOrCreateBasicType.  That's from memory though, I'm not 100% sure.

PR #556, opened 2020-03-11
Make Read implementations consistent

NextTurn:
`IMemoryReader.Read(ulong, Span<byte>, out int)` is documented to return "true if any bytes were read at all".

The call site needs to determine whether incomplete reads make sense.

/cc @cshung


cshung:
I like the change, it would be better if you can assert/validation at various `IDataReader.Read()` implementation that callers don't ask for exactly 0 bytes.

PR #569, opened 2020-03-20
Fixes a StackOverflowException in FixGenerics

TheXenocide:
I wanted to add some test cases and was trying to research a bit and make sure I was doing things right, but the basic premise is that I encountered the following type, which has a `1 but no generic type parameters specified:

``System.Collections.Generic.List`1[[Syncfusion.EJ2.Blazor.Data.Group`1, Syncfusion.EJ2.Blazor]]``

If that type implies what I think it does I don't really know how an instance of it would get into memory, but that's what I encountered. Just to confirm what I thought I ran the following script in LINQPad:

```csharp
void Main()
{
	typeof(GenNameTest).AssemblyQualifiedName.Dump();
	typeof(GenNameTest<>).AssemblyQualifiedName.Dump();
	typeof(GenNameTest<int>).AssemblyQualifiedName.Dump();
}

public class GenNameTest { }

public class GenNameTest<T> : GenNameTest { }
```

and I got the following output (all types in a LINQPad script are nested types in a generated "UserQuery" class):

```
UserQuery+GenNameTest, query_hkfjhr, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
UserQuery+GenNameTest`1, query_hkfjhr, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
UserQuery+GenNameTest`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]], query_hkfjhr, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
```

I'm not sure how an instance would exist of a generic type definition as opposed to a generic type with type parameters specified, but that's what I perceive to be happening here. Maybe there are other conditions that translate to backticks in the naming system?

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
@@ -1163,7 +1163,7 @@ namespace Microsoft.Diagnostics.Runtime.Builders
             while (i < len)
             {
                 if (name[i] == '`')
-                    while (i < len && name[i] != '[')
+                    while (i < len && name[i] != '[' && name[i] != ',')
                         i++;
 
                 if (name[i] == ',')


TheXenocide:
For reference, my current fix only prevent the overflow. Conceptually, I think it doesn't translate the name quite right. The output of that patched function is:

`System.Collections.Generic.List<Syncfusion.EJ2.Blazor.Data.Group>`

but if my suppositions are correct, I think it should probably be:

`System.Collections.Generic.List<Syncfusion.EJ2.Blazor.Data.Group<>>`

NextTurn:
> I'm not sure how an instance would exist of a generic type definition as opposed to a generic type with type parameters specified, but that's what I perceive to be happening here.

I wouldn't expect it's possible to create an instance of a non-instantiated generic type...

Unrelated: You didn't configure the commit author correctly. You're using a private email address. For more information, see [Setting your commit email address](https://help.github.com/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address)

weltkante:
> I wouldn't expect it's possible to create an instance of a non-instantiated generic type...

Could it be that the type got optimized away? Desktop Framework used to merge class-based generics into a single implementation (i.e. `List<Control>` and `List<Form>` would use the same JIT compiled code) and depending on the context (e.g. if you just had the reference to a compiled method) you couldn't recover the type. Don't know how .NET Core does it and how it manifests in the type string.

>  I encountered the following type

I think the key question is _where_ did you encounter that type? `typeof(List<>)` is a valid type and you can do reflection to get MethodInfo and other properties off it, so I'd not rule out that somewhere the runtime has a representation of `List<>` you can come across via clrmd. Though if you found it as the type of a live object instance that would indeed be weird.

> my current fix only prevents the overflow. Conceptually, I think it doesn't translate the name quite right. The output of that patched function is

That has me worried, because it means when people will come across this kind of type they get false results (in particular dangerous because you can overload on number of generic parameters, so both `Syncfusion.EJ2.Blazor.Data.Group` and ``Syncfusion.EJ2.Blazor.Data.Group`1`` can exist)

I think if you fix the stack overflow but can't preserve the meaning of the type you should just throw some other exception rather than returning false results.

TheXenocide:
> You're using a private email address

I presume you mean from the Pull Request? My local commit log shows a non-github e-mail address. My license/cla requirement appears to be adequately met, but I suppose if that's required does that mean I'll need to close/resubmit the pull request (not that I'm sure this solution should be the accepted solution yet anyway).

> Could it be that the type got optimized away?

This has been known to happen in other ways so I suppose it's possible, though if it was combining types I would think it would still specify some type (e.g. `List<Control>`)? Just random speculation there and boy would that be misleadingly painful for some situations.

> I think the key question is _where_ did you encounter that type?

That type was the Name from a `ClrType` either on a `ClrObject` from `ClrHeap.EnumerateObjects` or it was from the `EnumerateTypes` extension method in the linked issue (`ClrModule.EnumerateTypeDefToMethodTableMap` -> `ClrRuntime.GetTypeByMethodTable`). It definitely wasn't obtained from a field or stack `Type` variable value, if that's what you're wondering.

> I think if you fix the stack overflow but can't preserve the meaning of the type you should just throw some other exception rather than returning false results.

If the memory dump heap is valid, should it ever be such that you can not enumerate it? Maybe this could be configurable, or a property can identify the certainty of the identification? I would think this could make the tool completely unusable on some dumps when it would still be useful, if not 100% accurate, which isn't always a guarantee of a managed dump anyway (e.g. GC Handles). Alternatively, names could be left in their non-friendly format, which would be accurate to how the runtime sees them, at least (not my favorite option, honestly, but another option with this would be a `TryGetCSharpName` or something like that).

weltkante:
>  does that mean I'll need to close/resubmit the pull request

You can fix it in your branch and do a force-push, the github PR infrastructure supports forced pushes

> though if it was combining types I would think it would still specify some type (e.g. List<Control>)?

Profilers show `List<__Canon>` or something along those lines (if I remember correctly), but I don't know who is responsible for that name. I certainly don't know how those names look in the runtime strings.

> [on the suggestion of throwing an exception instead of returning wrong data] ... I would think this could make the tool completely unusable on some dumps when it would still be useful, if not 100% accurate, which isn't always a guarantee of a managed dump anyway (e.g. GC Handles).

I do not agree, intentionally returning wrong data (by cutting off the arity of the type instead of understanding what it means) is not an option. Either
* get someone involved who understands what these types mean so it can be implemented correctly
* reliably throw or otherwise communicate that the program encountered data it did not understand
* don't try to process the data if you can't understand what it means

The problem I see is that you are introducing a bug in clrmd (dropping arity) which can cause users of the library to not notice at all that something is wrong.

If both ``Syncfusion.EJ2.Blazor.Data.Group`1`` and ``Syncfusion.EJ2.Blazor.Data.Group`` exist then returning `List<Syncfusion.EJ2.Blazor.Data.Group>` is just plain wrong. Tools using clrmd will not notice this, they'll happily attribute whatever they are doing to the wrong type. If both types have different fields this failure cascades itself, possible leading to a crash, possibly just bogus results in the dump analysis. If tool devs even notice that something is wrong they'll probably just think they got a bad dump and blame Windows, when in fact its all along a bug in clrmd not understanding how to parse the type, mixing it up with a different one.

I'd like to not introduce that kind of bug in clrmd, it makes me trust it less, probably having to fork it. Thats an amount of effort I'd like to avoid.

leculver:
Sorry for taking so long to get around to this.

Thank you for putting together the PR!  For context I asked to see a potential PR to understand the issue that was being reported.

> If the memory dump heap is valid, should it ever be such that you can not enumerate it?

Actually there are plenty of cases where ClrMD cannot properly walk stacks, or the heap, or other data structures.  The CLR debugging layer is built as a "shadow" compile of the original CLR code which carries with it all of the runtime's assumptions and requirements.

For example, let's say you used a debugger to suspend a process in a place where the CLR Stackwalker cannot walk the stack of a thread.  This would mean ClrMD would not be able to walk that thread either, nor would SOS, nor would ICorDebug...  The CLR runtime itself would never attempt to walk a stack under normal circumstances, but the diagnostic layer can encounter that though it's supposed to be rare-ish.

Long story short, there's plenty of times that diagnostic tools can't inspect a part of a dump.  That's supposed to be (and is) a relatively rare case.

> [with regards to returning a "slightly incorrect type name".
> I think if you fix the stack overflow but can't preserve the meaning of the type you should just throw some other exception rather than returning false results.

We should never return an incorrect type name.  If we do return an incorrect name, it's a bug we should fix.  Also keep in mind that ClrType.Name is intentionally defined as `string?`.  If we cannot properly determine the type name ClrType.Name will return null.  This could be because of missing metadata or (for example) because we couldn't parse the string we got back from CLR.

> > I'm not sure how an instance would exist of a generic type definition as opposed to a generic type with type parameters specified, but that's what I perceive to be happening here.
> 
> I wouldn't expect it's possible to create an instance of a non-instantiated generic type...

As for the original reported issue, NextTurn is correct.  We should never see that kind of type in ClrMD, afaik.  ClrMD doesn't implement a view of the full CLR typesystem, it only represents _constructed typehandles_, which should never be an open, non-instantiated generic type.  The closest we get is __CANON which has a different meaning.

I don't want to merge this kind of change unless there's a real way to encounter this infinite loop in the wild.

Anyway, thanks again for the info and report here!

TheXenocide:
OK, I think I can agree with the sentiment/priorities here. That said, the type mentioned here was encountered in a real-world dump which amounted to a StackOverflowException; I don't think this pull request is the correct solution to the problem as much as it got me past the issue well enough for me to analyze the things I was actually interested in, but based on what's noted here it seems like the correct solution when encountering the open type (that should not exist) scenario might well be to return null when the condition is detected?

TheXenocide:
Also as a point of curiosity, I really do wish I understood how the CLR wound up with that in memory in the first place or what it might actually have been describing

weltkante:
Regarding what it represents I'm pretty sure it can only represent an open generic, as the optimized generics with placeholders use the `__Canon` type in place of the concrete type which gets dropped (have seen another issue recently on this repo which had this type in its signature).

As for how it gets into the dump I have no idea, but things like `typeof(List<>).GetMethod("Add")` and `typeof(List<>).MakeGenericType(typeof(List<>)).GetMethod("Add")` do return valid methods. You can't invoke them or create instances of their containing type but they can be described, so maybe its possible to observe them in a dump if the code was juggling with open generics.

PR #570, opened 2020-03-21
Clean up tests

NextTurn:


src/Microsoft.Diagnostics.Runtime.Tests/src/MemoryReaderTests.cs
```diff
@@ -23,7 +23,7 @@ public void SearchMemoryTest()
                 HashSet<ulong> seen = new HashSet<ulong>() { 0 };
                 List<ClrObject> firstSeenObjectOfType = new List<ClrObject>();
 
-                // We will search for method tables so make sure we
```
NextTurn: Make sure what?

NextTurn: @leculver?

leculver: Bad comment left after a refactor.  If I remember correctly, the code was originally written to search for specific method tables at specific locations, but the GC heap isn't quite that deterministic.  I removed the code that comment was for but missed part of the comment.

PR #572, opened 2020-03-24
Temporarily disable test asset building

leculver:
Building test assets with "AfterTargets=Build" is currently breaking the arcade build system signing step.  This only happens within Arcade and no one on that team has been able to figure out why.  Disabling this part of the build for now so we can get 2.0-beta builds out the door.


leculver:
According to the build team, this is how we are supposed to do post-build targets in a supported way:

> AFAIK most repos take advantage of creating an eng/afterSolutionTargets.props file to add targets that should happen after the solutoin is built: 
> https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md#engaftersolutionbuildtargets-optional

PR #574, opened 2020-03-26
Use WINAPI calling convention for DAC APIs

NextTurn:
See https://github.com/dotnet/diagnostics/issues/846#issuecomment-604231358

Thanks @swift-kim

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataMethod.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataMethod.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             return hr == S_OK ? map : null;
         }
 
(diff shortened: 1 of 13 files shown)


swift-kim:
Thanks! @leculver Do you mind if we do the same for 1.1?

swift-kim:
Do you think https://github.com/microsoft/clrmd/compare/master...NextTurn:x86 or https://github.com/microsoft/clrmd/compare/release/1.1...NextTurn:v1-x86 needs any rework?

NextTurn:
> Do you think [master...NextTurn:x86](https://github.com/microsoft/clrmd/compare/master...NextTurn:x86) or [release/1.1...NextTurn:v1-x86](https://github.com/microsoft/clrmd/compare/release/1.1...NextTurn:v1-x86) needs any rework?

Not really. I didn't implement support for debug registers and floating point registers but I don't think they matter.

NextTurn:
We can update the DbgEng APIs to use WINAPI as well (for consistency).

PR #575, opened 2020-03-26
Support ARM/ARM64 contexts

swift-kim:
For release/1.1

Similar to #421 (master), but ProcessArchitecture == DataTarget.Architecture is assumed (we don't support cross-debugging scenario).


src/Microsoft.Diagnostics.Runtime/src/Common/ContextHelper.cs
```diff
@@ -53,14 +54,33 @@ private static void Init()
             if (_context != null)
                 return;
 
+#if !NET45
```
leculver: I need to go back and revisit how we determine the sizes here.  It's a much bigger change than this, so we'll check this in as-is.

PR #580, opened 2020-04-04
[v1] Use WINAPI calling convention for native APIs

NextTurn:
See https://github.com/dotnet/diagnostics/issues/846#issuecomment-606996179

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Native/src/DacInterface/SosNative.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Native/src/DacInterface/SosNative.cs
@@ -105,7 +105,6 @@ namespace Microsoft.Diagnostics.Runtime.Native.DacInterface
             return hr == S_OK;
         }
 
-        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         public delegate void ThreadRootCallback(IntPtr token, ulong symbol, ulong address, ulong obj, int pinned, int interior);
 
         public bool TraverseStackRoots(ulong threadAddr, IntPtr initialContext, int contextSize, ThreadRootCallback callback, IntPtr token)
@@ -119,7 +118,6 @@ namespace Microsoft.Diagnostics.Runtime.Native.DacInterface
             return hr == S_OK;
         }
 
-        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         public delegate void HandleCallback(IntPtr token, ulong handleAddress, ulong dependentTarget, int handleType, uint refCount, int strong);
 
         public bool TraverseHandleTable(HandleCallback callback, IntPtr token)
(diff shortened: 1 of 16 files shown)


weltkante:
Should it documented somewhere why things are as they are? Maybe in form of a unit test reflecting at delegate metadata to catch accidental changes? Just looking at unannotated delegates might be tempting to "fix" for someone starting to work on the project in the future.

leculver:
I just want to make sure I understand this change before merging it.  This change is asserting:

1. The calling convention of COM calls on Windows is stdcall.
2. The calling convention for "COM" code on Linux is cdecl (e.g. ISOSDac6::DacGetMethodTableCollectibleData's implementation is cdecl on Linux).
3. The calling convention _should_ be changed to WinAPI which sets the appropriate behavior (1 & 2) on the corresponding systems.
4. Simply omitting the [UnmanagedFunctionPointer] attribute altogether is equivalent to `[UnmanagedFunctionPointer(CallingConvention.WinApi)]`.

Is that correct?

A couple questions:

1.  What problem is being solved changing from StdCall to WinApi?  IE how does ClrMD function at all on linux without causing stack corruption if we've been providing the wrong calling convention?
2.  Does the correct behavior still happen on Desktop CLR?  I assume the default is stdcall on desktop clr?

NextTurn:
> 1. The calling convention of COM calls on Windows is stdcall.

`STDMETHODCALLTYPE` is defined as `__stdcall` in *winnt.h*.

> 2. The calling convention for "COM" code on Linux is cdecl (e.g. ISOSDac6::DacGetMethodTableCollectibleData's implementation is cdecl on Linux).

`STDMETHODCALLTYPE` is defined as `__cdecl` (since .NET Core 2.1).

https://github.com/dotnet/runtime/blob/8e0147ecdfd2717362ca0a859089968bad17aefc/src/coreclr/src/debug/daccess/dacimpl.h#L1190

https://github.com/dotnet/runtime/blob/8e0147ecdfd2717362ca0a859089968bad17aefc/src/coreclr/src/pal/inc/rt/palrt.h#L197

> 3. The calling convention _should_ be changed to WinAPI which sets the appropriate behavior (1 & 2) on the corresponding systems.

WinAPI is converted to the default calling convention.

https://github.com/dotnet/runtime/blob/8e0147ecdfd2717362ca0a859089968bad17aefc/src/coreclr/src/vm/dllimport.cpp#L3292

> 4. Simply omitting the [UnmanagedFunctionPointer] attribute altogether is equivalent to `[UnmanagedFunctionPointer(CallingConvention.WinApi)]`.

The default calling convention is `__stdcall` on Windows and `__cdecl` on Linux (since .NET Core 2.1).

https://github.com/dotnet/runtime/blob/8e0147ecdfd2717362ca0a859089968bad17aefc/src/coreclr/src/vm/dllimport.cpp#L3273

> 1. What problem is being solved changing from StdCall to WinApi?  IE how does ClrMD function at all on linux without causing stack corruption if we've been providing the wrong calling convention?

If we've been providing the wrong calling convention on Linux x86, ClrMD causes stack corruption in every reverse interop and the program crashes with segmentation fault.

> 2. Does the correct behavior still happen on Desktop CLR?  I assume the default is stdcall on desktop clr?

I assume yes.

leculver:
Ahh thanks for the clarifications!

> If we've been providing the wrong calling convention on Linux x86, ClrMD causes stack corruption in every reverse interop and the program crashes with segmentation fault.

The fact that it was Linux x86 was also the part I missed.  This isn't a scenario I tested before.

leculver:
I will make this same change in 2.0.  I'm working on changes closely related to this and will fold that into it.

PR #585, opened 2020-04-08
Use HRESULT instead of bool

leculver:
- Intentionally start accepting more SUCCEEDED(hr) instead of hr == S_OK.
- Created an HResult struct to wrap COM's HRESULT.
- HResult has an implicit int cast operator for convenience.
- HResult has an implicit bool cast operator which is the equivalent of "SUCCEEDED(hr)".
- Change most wrappers to use and return HResult instead of bool.
- Functionality should be mostly unchanged, but allows for better control/error flow.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataProcess.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataProcess.cs
@@ -16,14 +16,6 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
     public unsafe class ClrDataProcess : CallableCOMWrapper
     {
         private static readonly Guid IID_IXCLRDataProcess = new Guid("5c552ab6-fc09-4cb3-8e36-22fa03c798b7");
-
-        private FlushDelegate? _flush;
-        private GetTaskByOSThreadIDDelegate? _getTask;
-        private RequestDelegate? _request;
-        private StartEnumMethodInstancesByAddressDelegate? _startEnum;
-        private EnumMethodInstanceByAddressDelegate? _enum;
-        private EndEnumMethodInstancesByAddressDelegate? _endEnum;
-
         private readonly DacLibrary _library;
 
         public ClrDataProcess(DacLibrary library, IntPtr pUnknown)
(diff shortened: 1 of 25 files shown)


leculver:
This is complete, but awaiting work to un-break our internal arcade build system which I'm working on now.

leculver:
> There is a comment on L33.

L33 of what file?

> Is this a bug? I haven't seen anything fail before.

Not a bugfix, just general cleanup to more closely align with how I handle COM interfaces in other (internal) projects I work on.  I prefer to have the HRESULT around somewhere rather than a true/false when debugging failures.

src/Microsoft.Diagnostics.Runtime/src/DbgEng/DebugSystemObjects.cs
```diff
@@ -29,41 +29,45 @@ public uint GetProcessId()
...
-            DebugOnly.Assert(hr == S_OK);
+            HResult hr = _getNumberThreads(Self, out int count);
+            DebugOnly.Assert(hr);
             return count;
         }
 
         internal void Init()
         {
             InitDelegate(ref _getSystemId, VTable.GetCurrentSystemId);
-            _systemId = _getSystemId(Self, out int id);
```
leculver, line 66: @NextTurn Actually there was one bug "fixed" with this checkin, it's this line.  We originally set the _systemId to be the HRESULT of the return of `IDebugSystemObjects::GetSystemId`, which is a bug.

I fixed that here, but I'm seeing odd behavior in tests, so I'm debugging the problem now.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataProcess.cs
```diff
@@ -118,13 +109,12 @@ public IEnumerable<ClrDataMethod> EnumerateMethodInstancesByAddress(ClrDataAddre
...
 
-            int hr = _startEnum(Self, addr, IntPtr.Zero, out ClrDataAddress handle);
-            if (hr != S_OK)
+            if (!_startEnum(Self, addr, IntPtr.Zero, out ClrDataAddress handle))
                 return result;
 
             try
             {
-                while ((hr = _enum(Self, ref handle, out IntPtr method)) == S_OK)
+                while (_enum(Self, ref handle, out IntPtr method))
```
dailongchen, line 117: HResult.S_OK is zero, this change breaks this function.

NextTurn: Anything wrong?

#### Before

`hr == 0`

#### After

`hr >= 0`

leculver: Yep that's a bug.  Thanks!  I've fixed this here, which will be released in a nuget package this week:  https://github.com/microsoft/clrmd/pull/819

PR #587, opened 2020-04-08
Change targets for UnitTest dependency build

leculver:


NextTurn:
Do we still need to move the target to *eng/AfterSolutionBuild.targets* (https://github.com/microsoft/clrmd/pull/572#issuecomment-603345590)?

leculver:
@NextTurn I'm working with the .net core engineering team to try out a few theories as to why the current build works great outside of their system, but fails during their pack phase.  If it still fails after I've tried all of their suggestions then I'll file an issue detailing the situation.  The suggestion in that comment was one thing to try to do, but it's not the only option.

PR #588, opened 2020-04-08
Add ClrObject.EnumerateReferencesWithFields

leculver:
Add a way to enumerate references of an object with a source of where that reference came from.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrFieldReference.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrFieldReference.cs
@@ -1,0 +1,120 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    public struct ClrFieldReference
+    {
+        const ulong OffsetFlag = 1 << 61;
+        const ulong DependentFlag = 1 << 60;
+        const ulong ValueMask = ~(OffsetFlag | DependentFlag);
+
+        private readonly ulong _offsetOrHandle;
+
+        /// <summary>
+        /// The object that <see cref="Field"/> contained.
+        /// </summary>
+        public ClrObject Object { get; }
+
+        /// <summary>
+        /// The offset into the containing object this address is found at.  Only valid if <see cref="IsObjectInterior"/> is true.
+        /// </summary>
+        public int Offset
+        {
+            get
+            {
+                if ((_offsetOrHandle & OffsetFlag) == OffsetFlag)
+                {
+                    unchecked
+                    {
+                        // The (uint) cast will slice off the high bits
+                        return (int)(uint)_offsetOrHandle;
+                    }
+                }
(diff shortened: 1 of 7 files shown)


leculver:
Need to test it (and add tests).  This is the rough direction I'm going in.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrFieldReference.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        /// <summary>
+        /// The field this object was contained in.  This property may be null if metadata is not available, if
+        /// this reference came from a dependent handle/loader allocator, or if this came from an array.
+        /// Only valid to call if <see cref="IsObjectInterior"/> is true.
+        /// </summary>
+        public ClrInstanceField? Field { get; }
+
+        /// <summary>
+        /// If this reference came from a CollectableType, it means there's a collectable type which keeps this object alive.
```
NextTurn, line 51:
```suggestion
        /// If this reference came from a collectible type, it means there's a collectible type which keeps this object alive.
```

src/Microsoft.Diagnostics.Runtime/src/Common/ClrFieldReference.cs
```diff
@@ -0,0 +1,120 @@
...
+                // Return the loader allocator handle if this doesn't have the high bits set.
+                if ((_offsetOrHandle & ValueMask) == _offsetOrHandle)
+                    return ValueMask & _offsetOrHandle;
+
+                return 0;
+            }
+        }
+
+        /// <summary>
+        /// Returns true if this reference came from a LoaderAllocator handle (e.g. collectable types).
```
NextTurn, line 67:
```suggestion
        /// Returns true if this reference came from a LoaderAllocator handle (e.g. collectible types).
```

src/Microsoft.Diagnostics.Runtime/src/Common/ClrFieldReference.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        /// <summary>
+        /// Returns true if this reference came from a dependent handle.
+        /// </summary>
+        public bool IsDepenendentHandle => (_offsetOrHandle & DependentFlag) == DependentFlag;
+
+        /// <summary>
+        /// Returns true if this reference came from a field in another object.
+        /// </summary>
+        public bool IsObjectInterior => LoaderAllocator == 0;
```
NextTurn, line 79: ❓

leculver: Yeah, need to find a more descriptive term.  =)

NextTurn: This is asserting that a dependent handle is also a field in another object. I don't understand why this is not:

```suggestion
        public bool IsObjectInterior => (_offsetOrHandle & OffsetFlag) == OffsetFlag;
```

leculver: Ah right.  Also this isn't ready to merge, I was mostly only publishing it for a dev who was asking for the feature as a rough sketch of the design.

PR #589, opened 2020-04-09
Accept more spans

NextTurn:


NextTurn:
> src/DbgEng/DebugSystemObjects.cs(25,28): error CA1822: Member Enter does not access instance data and can be marked as static (Shared in VisualBasic)

PR #597, opened 2020-04-10
Remove project reference from Utilities to core library

NextTurn:


NextTurn:
> src/DbgEng/DebugSystemObjects.cs(25,28): error CA1822: Member Enter does not access instance data and can be marked as static (Shared in VisualBasic)

leculver:
I'll fix the error tomorrow.  Strange that my VS isn't picking up that warning.

PR #601, opened 2020-04-13
Remove UnitTest project

leculver:
I'm removing the UnitTets project for a few reasons:

- It's a pain to have two test projects, as it doesn't play well with tooling (VS code coverage not merging the two, for example).
- Overall code coverage for UnitTests is very low.
- I plan to expand M.D.R.Tests to make up for the small amount of code that was covered by UnitTests and not Tests.
- Tests written in the style of M.D.R.UnitTests requires a lot of verbage/code to accomplish narrow testing.
- When UnitTests fail, it's very difficult to step through and debug the issue, often requiring a deeper understanding of the test framework than I would like.

The last two points are probably a personal failing in my stubbornness to learn a new system of testing, but in reality I have a narrow amount of time to work on ClrMD and as a result these tests have just been rotting/ignored.  I generally agree that having true "unit" tests is a great thing when it works well, but these particular tests don't fit my mental model and are too expensive for me to keep up to date or expand.


NextTurn:
> - Tests written in the style of M.D.R.UnitTests requires a lot of verbage/code to accomplish narrow testing.

TBH I never knew what the unit tests were doing so I never touched them. I prefer the simplicity of xUnit.

PR #603, opened 2020-04-13
Usage of readonly structs and members on structs

drewnoakes:
- Make structs `readonly` where possible in the `Microsoft.Diagnostics.Runtime` assembly
- Remove some redundant `readonly` member modifiers
- Mark some members of `ClrArray` as `readonly`


leculver:
LGTM, thanks!

Issue #564, opened 2020-03-14
Early exit for corrupted PE images resources?


NextTurn:
When experimenting with my own version of *createdump* on x86, the number of resource entries were unreasonably large (over thousands) but we were unable to detect this error.

https://github.com/microsoft/clrmd/blob/69db395bbcc3db1fbdb7bfa0bad31cafebbe72ea/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/ResourceEntry.cs#L130

A known problem is that we didn't check if `PEImage.Read<T>` returned a valid result.

https://github.com/microsoft/clrmd/blob/69db395bbcc3db1fbdb7bfa0bad31cafebbe72ea/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs#L382

swift-kim:
I also have this issue on my machine (Tizen armel). Printing a simple stack trace of an application (which normally has 40+ dlls loaded in memory) took `1m43.092s` for a minidump and `0m13.167s` for a minidump with heap.
```
count = 18328 + 63042
count = 0 + 0
count = 37500 + 58526
count = 0 + 1
count = 0 + 1
...
```

mikem8361:
I have also noticed a problem like this in dotnet-dump's `lm` command. A huge perf problem that affects the first command run because the modules are enumerated to find the runtime module. Getting the Version for each PE image in the Linux core dump was taking way too long and failing.  

The problem is the "layout" of managed PE modules. Some PE's are "file" layout and some are "loaded" layout. The "isVirtual" flag in the PEImage constructor needs to be set based on the layout, but there is no way to determine that in the core dump data reader. If the layout/isVirtual flag is set incorrectly, PEImage doesn't enumerate through the resources correctly, takes a long time because it is reading really long garbage strings, etc.

The solution I've been working (for release/1.1 for now since that is what dotnet-dump needs), is to skip all the PE's in the ELF reader module list.  IDataReader.EnumerateModules() is meant to return just native modules (like libcoreclr.so, etc) but the PE's show up because it uses the NT_FILE entries in the coredump and because the PE's are memory mapped by the runtime as shared named memory,

[...]

PR #609, opened 2020-04-19
Improve ELF reader EnumerateModules() perf

mikem8361:
The dotnet-dump's `lm` command and any command run for the first time has a huge perf problem. The modules are enumerated to find the runtime module. Getting the Version for each PE image in the Linux core dump was taking way too long and failing.

The problem is the "layout" of managed PE modules. Some PE's are "file" layout and some are "loaded" (sections are aligned) layout in memory. The "isVirtual" flag in the PEImage constructor needs to be set based on the layout, but there is no way to determine that at the core dump data reader level. If the layout/isVirtual flag is set incorrectly, PEImage doesn't enumerate through the resources correctly, takes a long time because it is reading really long garbage strings, etc.

IDataReader.EnumerateModules() is meant to return just native modules (like libcoreclr.so, etc) but the PE's show up because it uses the NT_FILE entries in the coredump and because the PE's are memory mapped by the runtime as shared named memory,

The first part of the fix is to skip all the PE's (modules that ELFLoadedImage.Open can't open). The managed assemblies can be enumerated at a higher level (ClrRuntime.Modules) without any problems. dotnet-dump has a "clrmodules" command to do that.

The second part of the fix is to get the module's Version on demand and try both file and loaded layout of the PE images.

Fixed a problem in the ResourceEntry where resource id entries were trying to read a name string. Added check in the ResourceEntry read string code to limit strings to MaxPath (1024) characters for when the layout flag is wrong.

Fix PEImageResourceTests.FileInfoVersionTest. It wasn't comparing the same type of versions.

Issue: https://github.com/microsoft/clrmd/issues/564

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/PEImageResourceTests.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/PEImageResourceTests.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Diagnostics.Runtime.Tests
             Assert.NotNull(fileVersion.FileVersion);
 
             ClrInfo clrInfo = dt.ClrVersions[0];
-            Assert.True(clrInfo.Version.ToString() == fileVersion.VersionInfo.ToString());
+            Assert.Equal(clrInfo.Version, fileVersion.VersionInfo);
         }
 
         [Fact]
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/PEImageResourceTests.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/PEImageResourceTests.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Diagnostics.Runtime.Tests
             Assert.NotNull(fileVersion.FileVersion);
 
             ClrInfo clrInfo = dt.ClrVersions[0];
-            Assert.Contains(clrInfo.Version.ToString(), fileVersion.FileVersion);
+            Assert.True(clrInfo.Version.ToString() == fileVersion.VersionInfo.ToString());
         }
 
         [Fact]
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
@@ -2,25 +2,101 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Diagnostics.Runtime.Linux;
+using Microsoft.Diagnostics.Runtime.Utilities;
+using Microsoft.Diagnostics.Runtime.Utilities.Pdb;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
-using Microsoft.Diagnostics.Runtime.Linux;
-using Microsoft.Diagnostics.Runtime.Utilities;
 
 namespace Microsoft.Diagnostics.Runtime
 {
     internal class CoreDumpReader : IDataReader2
     {
+        private class CoreModuleInfo : ModuleInfo
+        {
+            internal readonly ElfFile _elfFile;
+            private readonly PEImage _pe;
+
+            public CoreModuleInfo(IDataReader reader, ElfLoadedImage image)
+                : base(reader)
+            {
+                Debug.Assert(reader != null);
+                _elfFile = image.Open();
+
+                uint fileSize = (uint)image.Size;
+                uint timeStamp = 0;
+
+                if (_elfFile == null)
+                {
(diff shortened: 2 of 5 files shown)


leculver:
@mikem8361 Currently the tests are disabled on checkin (and don't work on 1.1).  Did you run these through the UnitTests on Linux?  I wanted to confirm before merging because I don't want to have to do that myself.  =)

leculver:
Overall this looks fine for 1.1, assuming tests pass.  There's two issues here unrelated to this exact checkin:

1.  I want to make a similar change in 2.0 before merging this one (or shortly afterwards) otherwise the behavior will drift and I will forget about it.
2.  This approach of fixing it will need to be very different in 2.0 because ModulInfo is not meant to be inherited from in that version (it should be marked `sealed` but it's not yet).  I'll take a look at what needs to be done there Monday.

src/Microsoft.Diagnostics.Runtime.Tests/src/PEImageResourceTests.cs
```diff
@@ -23,7 +23,7 @@ public void FileInfoVersionTest()
             Assert.NotNull(fileVersion.FileVersion);
 
             ClrInfo clrInfo = dt.ClrVersions[0];
-            Assert.Contains(clrInfo.Version.ToString(), fileVersion.FileVersion);
+            Assert.True(clrInfo.Version.ToString() == fileVersion.VersionInfo.ToString());
```
NextTurn, line 26: Nit:

```suggestion
            Assert.Equal(clrInfo.Version, fileVersion.VersionInfo);
```

NextTurn:
Regarding the image layout, you may be interested in my magic idea in #468.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
```diff
@@ -89,44 +171,19 @@ public IList<ModuleInfo> EnumerateModules()
...
                 ulong interpreter = _core.GetAuxvValue(ElfAuxvType.Base);
 
-                _modules = new List<ModuleInfo>(_core.LoadedImages.Count);
-                foreach (ElfLoadedImage img in _core.LoadedImages)
-                    if ((ulong)img.BaseAddress != interpreter && !img.Path.StartsWith("/dev"))
-                        _modules.Add(CreateModuleInfo(img));
+                _modules = new List<CoreModuleInfo>();
+                foreach (ElfLoadedImage image in _core.LoadedImages)
+                {
+                    if ((ulong)image.BaseAddress != interpreter && !image.Path.StartsWith("/dev"))
```
leculver, line 177: This should be `!image.Path.StartsWith("/dev/")` to not ignore files in `/development` or something similar.

mikem8361:
Yes, I have run the tests locally. Microsoft.Diagnostics.Runtime.UnitTests succeeds.

Microsoft.Diagnostics.Runtime.Tests has one failure in GCRootTests.GCRoots:

```
Assemblies Run
Microsoft.Diagnostics.Runtime.Tests.dll
Summary
Tests run: 105 — Failures: 1, Run time: 8.626s, Finished: 04/20/2020 09:55:21

Failed tests
0.0311823s✘ Microsoft.Diagnostics.Runtime.Tests.GCRootTests.GCRoots
Assert.True() Failure\r\nExpected: True\r\nActual:   False
   at Microsoft.Diagnostics.Runtime.Tests.GCRootTests.GCRootsImpl(GCRoot gcroot) in C:\sd\clrmd\src\Microsoft.Diagnostics.Runtime.Tests\src\GCRootTests.cs:line 296
   at Microsoft.Diagnostics.Runtime.Tests.GCRootTests.GCRoots() in C:\sd\clrmd\src\Microsoft.Diagnostics.Runtime.Tests\src\GCRootTests.cs:line 271
```

mikem8361:
I was planning to attempt to cherry-pick this change to master, but if ModuleInfo is sealed, another way would be needed to get the module Version on-demand.  It still can take a while to get it for ELF modules because it is searching memory for the version string and for PE binaries because it tries both layouts.

leculver:
@mikem8361 hmm.  Yeah.  Let's get this checked in, I'm going to wrap similar changes here into main after we get https://github.com/microsoft/clrmd/pull/468 checked in.

I'll take care of the 2.0 change, but I'll have you review it to make sure it should fix the issue when you move forward to the next version.

mikem8361:
Thanks!

mikem8361:
Can you kick off an official build?

leculver:
Yeah, official builds are now automatic.  I just have to upload the corresponding nuget package.  A new NuGet package should be available later today.  I'll reply here with the link when it's uploaded.

leculver:
Whoops, I lied.  1.1 builds are NOT done automatically, only 2.0 builds are.  I just kicked off a build now.

leculver:
Published in this package:  https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/1.1.122004

Might still be a few hours before the build is indexed/usable.

Issue #607, opened 2020-04-17
Several issues in Debug build


swift-kim:
The following issues only occur in the Debug build. Maybe some are not reproducible on other platforms than Tizen armel.

1. (Linux) False-positive assertion when a memory mapped file path contains a space
https://github.com/microsoft/clrmd/blob/61f5f98030789f428d45b50a54a8242b6a5e9c12/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs#L415-L419

    ```
    Unhandled exception. Microsoft.Diagnostics.Runtime.AssertionException: Unknown data format
       at Microsoft.Diagnostics.Runtime.DebugOnly.Fail(String message) in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\Extensions\DebugOnly.cs:line 25
       at Microsoft.Diagnostics.Runtime.Linux.LinuxLiveDataReader.LoadMemoryMap() in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\Linux\LinuxLiveDataReader.cs:line 417
       ...
    ```

2. Casting failure in GetMethodTableSlot()
https://github.com/microsoft/clrmd/blob/31c031acd63f1dd00b15e0cb3d7e4dff88d3e9c9/src/Microsoft.Diagnostics.Runtime/src/DacInterface/SosDac.cs#L127-L128

    When `slot` is -1 (I don't know how this is possible), it turns into an OverflowException.
    ```
    Unhandled exception. System.OverflowException: Arithmetic operation resulted in an overflow.
       at Microsoft.Diagnostics.Runtime.DacInterface.SOSDac.GetMethodTableSlot(UInt64 mt, Int32 slot) in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\DacInterface\SosDac.cs:line 128
       at Microsoft.Diagnostics.Runtime.Builders.MethodBuilder.Init(SOSDac sos, UInt64 methodDesc, IMethodHelpers helpers) in E:\public\clrmd2.0\src\Microsoft.Diagnostics.Runtime\src\Builders\MethodBuilder.cs:line 36
    ```

    Although this can be easily escaped by wrapping with `unchecked` or adding a parameter validation (kind of `return 0 if slot < 0`), maybe I have to first figure out why `slot` is -1.

3. Huge DataOffset in ResourceEntry.GetChildren()
[...]

PR #610, opened 2020-04-20
Casting fixes

leculver:
- Use unsigned integers for everything related to MethodTable slots.
- Change DacTraverseModuleMap to accept an enum instead of using an int which has to be casted.


leculver:
Contributes to https://github.com/microsoft/clrmd/issues/607.

PR #611, opened 2020-04-20
Catch exceptions in ResourceEntry.Children

leculver:


leculver:
Fixes the last of https://github.com/microsoft/clrmd/issues/607.

Issue #605, opened 2020-04-14
Consider using named tuples in public and internal APIs


NextTurn:
... instead of Item1, Item2 everywhere. :/

leculver:
What's the syntax for what you are proposing?  I was reading through tuple types documentation for C#, but apparently this isn't correct:  `public abstract IEnumerable<(MethodTable: ulong, Token: int)> EnumerateTypeDefToMethodTableMap();`.  Or do you mean defining a struct with Deconstruct?

drewnoakes:
Type declaration is (int Age, string Name).

Instantiation is (1, "foo") or (Age: 1, Name: "Foo")

leculver:
I think I've found and fixed all instances of this.  Please let me know if you find one I missed.

I've also taken the time to remove the last remaining IReadOnlyList usage which shouldn't have been there.

PR #613, opened 2020-04-20
Use named tuples and remove IReadOnlyList usage

leculver:
Fixes https://github.com/microsoft/clrmd/issues/605.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrEnum.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrEnum.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Diagnostics.Runtime
         }
 
         public IEnumerable<string> GetEnumNames() => _values.Select(v => v.Name);
-        public IEnumerable<(string, object?)> EnumerateValues() => _values;
+        public IEnumerable<(string Name, object? Value)> EnumerateValues() => _values;
 
         private (string Name, object? Value)[] EnumerateValues(MetadataImport import, out ClrElementType elementType)
         {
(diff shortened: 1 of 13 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ClrHeap.cs
```diff
@@ -111,9 +112,9 @@ public abstract class ClrHeap
         /// <returns>The string representation of this heap.</returns>
         public override string ToString()
         {
-            double sizeMb = Segments.Sum(s => (long)s.Length) / 1000000.0;
-            int segCount = Segments != null ? Segments.Count : 0;
-            return $"ClrHeap {sizeMb}mb {segCount} segments";
+            ImmutableArray<ClrSegment> segments = Segments;
+            double sizeMb = segments.Sum(s => (long)s.Length) / 1000000.0;
+            return $"ClrHeap {sizeMb}mb {segments.Length} segments";
```
NextTurn, line 117: Unrelated:

```suggestion
            return $"ClrHeap {sizeMb}MB {segments.Length} segments";
```

src/Microsoft.Diagnostics.Runtime/src/Common/ClrEnum.cs
```diff
@@ -34,25 +34,25 @@ public ClrEnum(ClrType type)
...
+            object? value = _values.Single(v => v.Name == name).Value;
             if (value is null)
                 throw new InvalidOperationException($"Enum {Type.Name} had null '{name}' value.");
 
             return (T)value;
         }
 
-        public IEnumerable<string> GetEnumNames() => _values.Select(v => v.Item1);
+        public IEnumerable<string> GetEnumNames() => _values.Select(v => v.Name);
         public IEnumerable<(string, object?)> EnumerateValues() => _values;
```
NextTurn, line 51:
```suggestion
        public IEnumerable<(string Name, object? Value)> EnumerateValues() => _values;
```

PR #615, opened 2020-04-22
Rework ModuleInfo

leculver:
ModuleInfo changes:
- Change _isVirtual to be private
- Mark the class sealed
- Keep a reference to DataTarget instead of IDataReader
- Rename FileSize/Timestamp fields

PEImage changes:
- PEImage now contains a flag for whether to leave the stream open or not
- Clean up PEImage.Dispose related methods
- Make PEImage.Stream private
- Change PEImage.Reader to be nullable

Other:
- Update LinuxLiveDataReader to not need to write to IsVirtual
- Update DacInfo to not derive from ModuleInfo

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Diagnostics.Runtime
         {
             try
             {
-                PEImage image = new PEImage(new ReadVirtualStream(DataTarget.DataReader, (long)ImageBase, IndexFileSize), _isVirtual);
+                PEImage image = new PEImage(new ReadVirtualStream(DataTarget.DataReader, (long)ImageBase, IndexFileSize), leaveOpen: false, isVirtual: _isVirtual);
                 if (!_isManaged.HasValue)
                     _isManaged = image.IsManaged;
 
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System;
 using System.Collections.Immutable;
 using Microsoft.Diagnostics.Runtime.Utilities;
 
@@ -11,14 +10,16 @@ namespace Microsoft.Diagnostics.Runtime
     /// <summary>
     /// Provides information about loaded modules in a <see cref="DataTarget"/>.
     /// </summary>
-#pragma warning disable CA1001 // Types that own disposable fields should be disposable
-    public class ModuleInfo
-#pragma warning restore CA1001 // Types that own disposable fields should be disposable
+    public sealed class ModuleInfo
     {
-        private readonly IDataReader _dataReader;
         private bool? _isManaged;
         private VersionInfo? _version;
-        internal bool _isVirtual;
+        private readonly bool _isVirtual;
+
+        /// <summary>
+        /// The DataTarget which contains this module.
+        /// </summary>
+        public DataTarget DataTarget { get; internal set; }
 
         /// <summary>
         /// Gets the base address of the object.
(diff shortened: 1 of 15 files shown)


leculver:
@NextTurn This is what I was trying to say on how to handle `_isVirtual`.  I also did some cleanup around ModuleInfo/DacInfo as well.  Any thoughts?

Are you set up to test attaching to a live process on Linux?  If so would you please give this a try and see if the changes to [LinuxLiveDataTarget.cs](https://github.com/microsoft/clrmd/pull/615/files#diff-94eb6cdf033bfc33f80a71852c1956f5R111) were correct?

If not no worries, but I won't be able to get to testing it manually this week.

(There might be some build warnings, I will clean those up after CI checks the compile.)

NextTurn:
LINQ is powerful, yes.

Just one problem with `ModuleInfo.GetPEImage`: It now returns "invalid" `PEImage` for ELF images rather than `null` as documented.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
```diff
@@ -115,40 +116,38 @@ public VersionInfo Version
         {
             get
             {
-                if (_version is VersionInfo version)
+                if (_version.HasValue)
                 {
-                    return version;
+                    return _version.Value;
```
NextTurn, line 121: #396

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DacInfo.cs
```diff
@@ -7,10 +7,16 @@
 namespace Microsoft.Diagnostics.Runtime
 {
     /// <summary>
-    /// Represents the DAC dll.
+    /// This class provides information needed to located the correct CLR diagnostics DLL (this dll
+    /// is called the "dac").
```
NextTurn, line 11:
```suggestion
    /// is called the Debug Access Component (DAC)).
```

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -108,50 +108,47 @@ public void FlushCachedData()
...
+            }
 
-                ModuleInfo? module = result.FirstOrDefault(module => module.FileName == entry.FilePath);
-                if (module is null)
-                {
-                    int filesize = 0;
-                    int timestamp = 0;
-                    VersionInfo? version = null;
+            return result;
+        }
```
NextTurn, line 130: This method returns an enumerable.

```cs
public IEnumerable<ModuleInfo> EnumerateModules() =>
    from entry in _memoryMapEntries
    where !string.IsNullOrEmpty(entry.FilePath)
    group entry by entry.FilePath into image
    let filePath = image.Key
    let containsExecutable = image.Any(entry => entry.IsExecutable)
    let beginAddress = image.Min(entry => entry.BeginAddress)
    let _ = GetPEImageProperties(filePath)
    select new ModuleInfo(beginAddress, _.filesize, _.timestamp, filePath, containsExecutable, buildId: default, version: _.version);
```

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
leculver: hmm.  I know I merged that change but I'm rethinking that.  Overall I care more about readability/maintainability over micro optimizations.  (Though if something actually causes an actual perf issue in a hot loop I'm happy to optimize.)  In this case, this is a jit bug?  It's slow now, but at some point the JIT guys will fix the problem and the HasValue/Value will work as fast as other patterns.

Using `.HasValue` then `.Value` has symmetry and is much clearer in intent than "if (foo is Bar bar)" or `GetValueOrDefault`.  Additionally, `.Value` is saying "I really know this has a value or it'd thrown an exception this method isn't supposed to", versus `.GetValueOrDefault` could silently return the wrong value if the code above it isn't correct (or gets refactored incorrectly).

Essentially, I don't want to litter the code with micro optimizations which make the code harder to read or reason about.  If they have a noticeable perf impact or are in a tight loop that's when I want to use micro-optimizations.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
```diff
@@ -115,40 +116,36 @@ public VersionInfo Version
         {
             get
             {
-                if (_version is VersionInfo version)
-                {
-                    return version;
-                }
+                if (_version.HasValue)
+                    return _version.Value;
```
NextTurn, line 120: Fine with me as well. It's just bad that C# doesn't permit us to treat NVTs and NRTs the same.

PR #623, opened 2020-04-23
Do not permit write file share

NextTurn:


leculver:
This was intentional.  Was there a specific issue or failure related to this code?

The filesize/timestamp are found within the first 1k bytes of a binary (or so).  The specific use case here is that if there's a debugger (like windbg) or a parallel process containing ClrMD that's in the process of writing the file we could _possibly_ succeed in validating that the file they are currently writing and not attempt to re-download that file.

The subsequent File.Open where we use it does not allow write (so we wouldn't attempt to _use_ a file being written, we'd hit an IOException there and that's fine).

NextTurn:
> The filesize/timestamp are found within the first 1k bytes of a binary (or so). The specific use case here is that if there's a debugger (like windbg) or a parallel process containing ClrMD that's in the process of writing the file we could _possibly_ succeed in validating that the file they are currently writing and not attempt to re-download that file.

Ideally makes sense. IIRC WinDbg doesn't create the PE file before the download is complete, and ClrMD doesn't pre-allocate the full file size when downloading (which is necessary because the PE reader validates the metadata size). Any thoughts?

leculver:
Good point.  I haven't ever tested it and I'm not going to maintain it so I'll merge this one.

NextTurn:
What's worse, `File.Create` used by `SymbolServerLocator` specifies `FileShare.None` by default.

https://github.com/microsoft/clrmd/blob/4d94633b03eba2d88f81b0841654925210d3fe62/src/Microsoft.Diagnostics.Runtime/src/Implementation/SymbolServerLocator.cs#L244

PR #626, opened 2020-04-23
Don't call FixGenerics twice

leculver:
FixGenerics is called by ITypeHelpers.GetTypeName.  ClrmdType shouldn't also call it.


leculver:
Contributes to https://github.com/microsoft/clrmd/issues/627.  Just a workaround, need a real decision or fix on that issue.

Issue #625, opened 2020-04-23
Don't pre-calculate ModuleInfo's version/buildid on Linux


leculver:
On Linux it's more common to inspect a coredump from the local machine (or live process) than how ClrMD is used on Windows.  Since there's so much slowness associated with digging for version numbers (and potentially buildids), we should only request those on demand since we only really need them in the coredump case when the build is copied to another PC.

Also we'd need to update DacInfo to lazily request that data instead of taking it in a construct.  Probably need DacInfo to store the ModuleInfo for CLR and produce the information lazily.

PR #630, opened 2020-04-24
Lazily initialize ModuleInfo.BuildId

NextTurn:
Contributes to #625 

Some questions:

- `IDataReader.GetVersionInfo` is a `void` method with an `out` parameter?

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
@@ -144,19 +144,21 @@ namespace Microsoft.Diagnostics.Runtime
 
         public ImmutableArray<byte> GetBuildId(ulong baseAddress)
         {
-            ElfLoadedImage image = _core.LoadedImages.First(image => (ulong)image.BaseAddress == baseAddress);
-            ElfFile? file = image.Open();
-            return file?.BuildId ?? ImmutableArray<byte>.Empty;
+            return GetElfFile(baseAddress)?.BuildId ?? ImmutableArray<byte>.Empty;
         }
 
         public void GetVersionInfo(ulong baseAddress, out VersionInfo version)
         {
-            ElfLoadedImage image = _core.LoadedImages.First(image => (ulong)image.BaseAddress == baseAddress);
-            ElfFile? file = image.Open();
+            ElfFile? file = GetElfFile(baseAddress);
             if (file is null)
                 version = default;
             else
                 LinuxFunctions.GetVersionInfo(this, baseAddress, file, out version);
+        }
+
+        private ElfFile? GetElfFile(ulong baseAddress)
+        {
+            return _core.LoadedImages.First(image => (ulong)image.BaseAddress == baseAddress).Open();
         }
 
         public bool Read(ulong address, Span<byte> buffer, out int bytesRead)
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
```diff
@@ -141,6 +142,13 @@ public bool GetThreadContext(uint threadID, uint contextFlags, Span<byte> contex
             return false;
         }
 
+        public ImmutableArray<byte> GetBuildId(ulong baseAddress)
+        {
+            ElfLoadedImage image = _core.LoadedImages.First(image => (ulong)image.BaseAddress == baseAddress);
+            ElfFile? file = image.Open();
+            return file?.BuildId ?? default;
```
NextTurn, line 149:
```suggestion
            return file?.BuildId ?? ImmutableArray<byte>.Empty;
```

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -113,7 +114,7 @@ public void FlushCachedData()
             let containsExecutable = image.Any(entry => entry.IsExecutable)
             let beginAddress = image.Min(entry => entry.BeginAddress)
             let props = GetPEImageProperties(filePath)
-            select new ModuleInfo(beginAddress, props.Filesize, props.Timestamp, filePath, containsExecutable, buildId: default, version: props.Version);
+            select new ModuleInfo(beginAddress, props.Filesize, props.Timestamp, filePath, containsExecutable, buildId: ImmutableArray<byte>.Empty, version: props.Version);
```
NextTurn, line 117: Is build ID unavailable in live processes?

src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs
```diff
@@ -219,7 +220,7 @@ public IEnumerable<ModuleInfo> EnumerateModules()
                 for (int i = 0; i < bases.Length; ++i)
                 {
                     string? fn = _symbols.GetModuleNameStringWide(DebugModuleName.Image, i, bases[i]);
-                    ModuleInfo info = new ModuleInfo(bases[i], mods[i].Size, mods[i].TimeDateStamp, fn, isVirtual: true);
+                    ModuleInfo info = new ModuleInfo(bases[i], mods[i].Size, mods[i].TimeDateStamp, fn, isVirtual: true, buildId: ImmutableArray<byte>.Empty);
```
leculver, line 223: Maybe we should just remove the buildId parameter and always lazily calculate it?

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
leculver: That original code is wrong.  It exists but it isn't needed by ClrMD itself.  In the live scenario we should almost always have libmscordaccore.so on disk (unless the user published a build and/or deleted the file from disk) so we wouldn't need to hit the symbol server for it.  (I also realize ClrMD hasn't added symbol server support for buildid based files.)

In this case though, I think removing the parameter from the constructor and always lazily pulling the value is the right approach.

leculver:
> * `IDataReader.GetVersionInfo` is a `void` method with an `out` parameter?

That's one of many things I want to change on IDataReader, but I also want to do them all as one checkin because I know at least one team already using ClrMD 2.0 with a custom datareader and I'd like to minimize impact.

Some other IDataReader problems:

1. Remove `IsFullMemoryAvailable`, `EnumerateAllThreads`, etc that aren't used.
2. Remove `QueryMemory` or put it on IMemoryReader.
3. `GetVersionInfo` should return a bool.
3. Consider changing `IDataReader` to an abstract base class, or make `DataTarget`'s constructor take an abstract base class that contains a data reader instead of the current design.  Essentially I'd like to be able to add "optional" properties and functions to data readers without it being a breaking change.  Since we support desktop CLR, I can't use interface default methods...

It's on my todo list for the API review phase.  I'm wrapping up minidumps now.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/DbgEng/DbgEngDataReader.cs
NextTurn: Why lazily calculate if the build ID is known to be `Empty` on Windows?

leculver: Ah right.  =)

leculver:
@NextTurn This checkin may have conflicts with https://github.com/microsoft/clrmd/pull/635.  Please mark this PR as ready when it's ready for merge.  I'll be happy to merge this one first and resolve the conflicts if this one is done before https://github.com/microsoft/clrmd/pull/635.

PR #635, opened 2020-04-25
Change DataTarget startup paths

leculver:
- Remove unused IDataReader functions.  (QueryMemory, EnumerateAllThreads, etc).
- Don't throw "PlatformNotSupportedException" for just loading a dump.  DataTarget should be able to load dumps of any platform. I only intend to throw exceptions when attempting to CreateRuntime:  If a bitness mismatch occurs (x86 <==> amd64) then we'll throw `NotSupportedException` in CreateRuntime.  We'll also throw an exception if we cannot find an appropriate dac to load.  For example. if the user calls `CreateRuntime()` from windows targeting a linux coredump, we'll still attempt to locate a matching dac and load it if it exists.
- No longer accept .zip and .cab files.  It was always a quirk of DbgEng that we could load these files, and I'd like 2.0 to specifically require the input to LoadDump to be a supported dump format and not a compressed file that we have to deal with.
- Only have one function for loading dumps (LoadDump instead of LoadCoreDump, LoadCrashDump).
- Only have one function for attaching to a process.
- Use Microsoft.Diagnostics.NETCore.Client to create process "snapshots" on linux.
- Require DbgEng startup to go through a new path `DataTarget.CreateFromDbgEng(IntPtr pDebugClient)`
- No longer accept a raw IDataReader.  Instead use a pattern similar to `ProcessStartInfo` where the user needs to implement (or fill) a `CustomDataReader` which is an abstract base class (or maybe just an unsealed class).  This will let us add functionality to data reading in the future without making a breaking change to IDataReader.


leculver:
This is only a rough draft of this change, not ready for review.  Only posting as a heads up about upcoming changes.  The design around `CustomDataReader` is not finished yet and will change before checkin.  The code also runs but there's a lot of cleanup still left and unimplemented portions.

NextTurn:
Instead of enumerating and suspending all threads, can we simply call `NtSuspendProcess` (internal?) or non-invasively attach to the process?

leculver:
This actually is "non-invasive attach", as far as I know.  I looked at a bunch of options:

1)  `NtSuspendProcess` is not a documented win32 function so we are not allowed to use it.
2)  Be a real debugger.  We can use `DebugActiveProcess` but only one process can be a debugger for a process at a time, so we'd need a fallback anyway.
3)  Use SuspendThread on all threads along with OpenProcess/ReadProcessMemory/GetThreadContext to read data, but we have to worry about threads creating other threads.

AFAIK "non-invasive attach" is actually #3, but I could ask the windbg team to make sure they don't have some secret sauce to make this better.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -205,122 +208,163 @@ public IEnumerable<ModuleInfo> EnumerateModules()
...
             }
 
-            return new DataTarget(new DbgEngDataReader(processId, invasive: false, 5000));
+            throw new PlatformNotSupportedException(GetPlatformMessage(nameof(AttachToProcess), RuntimeInformation.OSDescription));
         }
 
         /// <summary>
-        /// Attaches to a snapshot process (see https://docs.microsoft.com/windows/win32/api/_proc_snap/).
-        /// This method is only supported on Windows.
+        /// Creates a snapshot of a running process and attaches to it.  This method will pause a running process
```
NextTurn, line 326: > This method will pause a running process

Really? Why? As I understand this used to be the only safe way to inspect a running process without suspending it.

leculver: I don't think I finished that comment.  It's supposed to be:  "This method will temporarily pause a running process to generate a snapshot, then resume the process and attach to the snapshot."

The design is:

1.  Use AttachToProcess(pid, suspend: true) - attach "non-invasively", the process is suspended by ClrMD
2.  Use AttachToProcess(pid, suspend: false) - Open process to read memory, this method is only supported if the user suspends the process themselves.
3.  Use CreateSnapshotAndAttach - Creates a "snapshot" and debugs that, fully supported.

For CreateSnapshotAndAttach:  On Windows we have PssCaptureSnapshot which is super fast, on Linux we'll create a coredump to the tmp folder and load it...which will take a lot longer than on Windows, but at least it's a consistent state to debug.

PR #640, opened 2020-04-27
Ensure ModuleInfo's BuildId and VersionInfo are lazily evaluated

leculver:
It can take a long time to produce the build id and version for various modules.  This change ensures that we calculate those values on demand.


leculver:
Fixes https://github.com/microsoft/clrmd/issues/625 (along with https://github.com/microsoft/clrmd/pull/630).

@mikem8361 In ClrMD 2.0 this should fix the issue where we spend too much time enumerating modules as in https://github.com/microsoft/clrmd/pull/609.  Let me know (after updating) if there's still more work to do here.

PR #641, opened 2020-04-28
Add Linux support for CreateSnapshotAndAttach

leculver:
(I haven't yet tested this checkin.)

Adding support for linux seemed oddly easy.  @mikem8361 do you think this would work taking a snapshot of your own process?  Have you ever tested that before?

I was reading through `DiagnosticsClient` and even though we'd pause before getting the IPC response everything _seems_ like it would work?  I'll do more testing before merging this change.

Also @NextTurn for code review, though it's a simple change.  I hate that I can't ask for reviews outside of the dotnet foundation...

I am using ArgumentException for errors attaching to a process.  That originally came from the fact that Process.GetProcessById will throw ArgumentException when the process isn't valid.  It still seems a bit odd to throw such a generic exception instead of something more meaningful, but I'm not sure what else I'd throw when we can't attach to a process for some reason.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/LinuxSnapshot/LinuxSnapshotTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/LinuxSnapshot/LinuxSnapshotTarget.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Diagnostics.Runtime
                 }
 
 #pragma warning disable CA2000 // Dispose objects before losing scope
-                LinuxSnapshotTarget result = new LinuxSnapshotTarget(new MinidumpReader(dumpPath), pid, dumpPath);
+                LinuxSnapshotTarget result = new LinuxSnapshotTarget(new CoreDumpReader(dumpPath, File.OpenRead(dumpPath)), pid, dumpPath);
                 dumpPath = null;
                 return result;
 #pragma warning restore CA2000 // Dispose objects before losing scope
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/LinuxSnapshot/LinuxSnapshotTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/LinuxSnapshot/LinuxSnapshotTarget.cs
@@ -1,0 +1,65 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.NETCore.Client;
+using System;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Runtime
+{
+    internal class LinuxSnapshotTarget : CustomDataTarget
+    {
+        private readonly int _pid;
+        private readonly string _filename;
+
+        public LinuxSnapshotTarget(IDataReader reader, int pid, string filename) : base(reader)
+        {
+            _pid = pid;
+            _filename = filename;
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            base.Dispose(disposing);
+
+            try
+            {
+                File.Delete(_filename);
+            }
+            catch
+            {
+            }
+        }
+
+        public override string ToString() => $"{_filename} (snapshot of pid:{_pid:x})";
+
+        public static LinuxSnapshotTarget CreateSnapshotFromProcess(int pid)
+        {
+            string? dumpPath = Path.GetTempFileName();
+            try
+            {
+                try
+                {
+                    DiagnosticsClient client = new DiagnosticsClient(pid);
+                    client.WriteDump(DumpType.Full, dumpPath, logDumpGeneration: false);
+                }
+                catch (ServerErrorException sxe)
+                {
+                    throw new ArgumentException($"Unable to create a snapshot of process {pid:x}.", sxe);
+                }
+
+#pragma warning disable CA2000 // Dispose objects before losing scope
+                LinuxSnapshotTarget result = new LinuxSnapshotTarget(new MinidumpReader(dumpPath), pid, dumpPath);
+                dumpPath = null;
+                return result;
+#pragma warning restore CA2000 // Dispose objects before losing scope
+            }
+            finally
+            {
+                if (dumpPath != null)
+                    File.Delete(dumpPath);
+            }
+        }
+    }
+}
--- a/src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
+++ b/src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
@@ -25,6 +25,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Diagnostics.NETCore.Client" Version="0.2.61701" />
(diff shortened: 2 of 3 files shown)


NextTurn:
> but I'm not sure what else I'd throw when we can't attach to a process for some reason

`ClrDiagnosticsException`...

leculver:
Fair point.  =)

NextTurn:
This doesn't seem right.

```
Unhandled exception. System.IO.InvalidDataException: File '/tmp/tmpHwVE1K.tmp' is not a Minidump.
   at Microsoft.Diagnostics.Runtime.Windows.Minidump..ctor(String crashDump, Stream stream) in /root/git/clrmd/src/Microsoft.Diagnostics.Runtime/src/Windows/Minidump.cs:line 91
   at Microsoft.Diagnostics.Runtime.MinidumpReader..ctor(String crashDump, Stream stream) in /root/git/clrmd/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs:line 41
   at Microsoft.Diagnostics.Runtime.MinidumpReader..ctor(String crashDump) in /root/git/clrmd/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs:line 27
   at Microsoft.Diagnostics.Runtime.LinuxSnapshotTarget.CreateSnapshotFromProcess(Int32 pid) in /root/git/clrmd/src/Microsoft.Diagnostics.Runtime/src/DataReaders/LinuxSnapshot/LinuxSnapshotTarget.cs:line 53
   at Microsoft.Diagnostics.Runtime.DataTarget.CreateSnapshotAndAttach(Int32 processId) in /root/git/clrmd/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs:line 347
```

NextTurn:
This doesn't seem right.

```
Unhandled exception. System.IO.InvalidDataException: File '/tmp/tmpHwVE1K.tmp' is not a Minidump.
   at Microsoft.Diagnostics.Runtime.Windows.Minidump..ctor(String crashDump, Stream stream) in clrmd/src/Microsoft.Diagnostics.Runtime/src/Windows/Minidump.cs:line 91
   at Microsoft.Diagnostics.Runtime.MinidumpReader..ctor(String crashDump, Stream stream) in clrmd/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs:line 41
   at Microsoft.Diagnostics.Runtime.MinidumpReader..ctor(String crashDump) in clrmd/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs:line 27
   at Microsoft.Diagnostics.Runtime.LinuxSnapshotTarget.CreateSnapshotFromProcess(Int32 pid) in clrmd/src/Microsoft.Diagnostics.Runtime/src/DataReaders/LinuxSnapshot/LinuxSnapshotTarget.cs:line 53
   at Microsoft.Diagnostics.Runtime.DataTarget.CreateSnapshotAndAttach(Int32 processId) in clrmd/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs:line 347
```

src/Microsoft.Diagnostics.Runtime/src/DataReaders/LinuxSnapshot/LinuxSnapshotTarget.cs
```diff
@@ -0,0 +1,65 @@
...
+                    DiagnosticsClient client = new DiagnosticsClient(pid);
+                    client.WriteDump(DumpType.Full, dumpPath, logDumpGeneration: false);
+                }
+                catch (ServerErrorException sxe)
+                {
+                    throw new ArgumentException($"Unable to create a snapshot of process {pid:x}.", sxe);
+                }
+
+#pragma warning disable CA2000 // Dispose objects before losing scope
+                LinuxSnapshotTarget result = new LinuxSnapshotTarget(new MinidumpReader(dumpPath), pid, dumpPath);
```
NextTurn, line 53:
```suggestion
                LinuxSnapshotTarget result = new LinuxSnapshotTarget(new CoreDumpReader(dumpPath, File.OpenRead(dumpPath)), pid, dumpPath);
```

leculver:
That was a silly mistake.  Thanks for the fix.  =)

I am not checking this in until I have a little time to test it.  My linux install is corrupted somehow so I will try to fix that and test this in the next day or so.

NextTurn:
Taking a snapshot of the own process or another process works well in my side.

Is this really a "snapshot"? The method name says it's a dump...

leculver:
It's conceptually a "snapshot".  On windows no dump is written, on linux and other platforms we have to (somehow) preserve the full memory address space, which is essentially a dump file written to disk.  If there's a more efficient way to make this work on Linux (and eventually OS X) I'd be happy to implement it a different way.  Or if there's a better term for what we are doing we could name it something else.

NextTurn:
@mikem8361 Just out of curiosity, have you ever explored the feasibility to fork a suspended child process as a snapshot (https://github.com/microsoft/clrmd/issues/303#issuecomment-556430260)?

leculver:
I think that would work well.  @mikem8361/@noahfalk:  Instead of the process dumping itself:

Parent process calls fork():
- Parent - Gets the child process id as return value, sends the child process ID back through IPC to dotnet-dump.  Parent process then continues on without any further action.
- Child - Returns 0 from fork, all threads except the current one are suspended (that's how fork works).  The last remaining thread immediately suspends its only remaining thread and does nothing more.

- Dotnet-dump: Gets the child process id from Parent's IPC call above.  It then reads process memory and writes out a minidump to disk.  If we sent the IPC to a previous version of coreclr, it would get back an error so send the regular self-dump IPC call.

Is there anything that sounds unreasonable?

noahfalk:
> Is there anything that sounds unreasonable?

We've been interested in the general premise for a while so entirely on-board there. The main issues thus far are getting time on the schedule to implement it and working out some technical challenges. Once we get into the details my understanding is the workflow you described will hit a few trouble-spots, though I suspect all of them are solvable:

1. I haven't personally investigated, but what I have heard 2nd hand is that there is only a single thread in the forked child process. The stack memory of the parent process threads is still there, but if you tried to enumerate the threads and get register contexts the OS would tell you only one exists and it has a different thread ID than the thread with the same stack had in the parent. If that description was accurate it would mean you need to do more work to suspend threads and capture thread register contexts before the fork() occurs.
2. If you want to avoid requiring super-user permissions then dotnet-dump can't be the process that reads the memory. Currently the way dotnet-dump works is it sends an IPC message to the target process and then that target process spawns CreateDump as a child. CreateDump is the one that captures the info and Linux allows child processes to use ptrace on their parent without super-user permission.

NextTurn:
> Linux allows child processes to use ptrace on their parent without super-user permission

Apparently CoreCLR calls `prctl` to grant PTRACE access to createdump.

https://github.com/dotnet/runtime/blob/3c0d86e584c6d549b70913ea94eedbc0f3b1b7ed/src/coreclr/src/pal/src/thread/process.cpp#L3328

leculver:
That first issue is a real killer.  For managed only diagnostics we could call SuspendRuntime before fork() and use clr!Thread along with the captured contexts for the GC to do diagnostics.  That would probably work for all of ClrMD, but generating real coredumps usable by other things (like !analyze) would be a challenge.  Hmm, I'll put more thought into it.

noahfalk:
For the threads issue my best current idea would be something like:
1. the thread about to call fork() sends all other threads a signal
2. use the signal to have them report their pre-signal handler register context back to the primary thread and block
3. fork() off the snapshot child process
4. In the parent resume all the suspended threads, in the child fork() again to initiate the CreateDump process.

I haven't dived into the details to know if the signaling mechanism would work the way I imagine and then the other risk is to see if it is possible that fork() could block. For example if fork() takes a lock inside its implementation then one of the suspended threads could be holding that lock.

PR #642, opened 2020-04-28
'try-finally' -> 'try-catch'

NextTurn:
The original pattern looks odd.


leculver:
Good catch, that's a better pattern.

I originally avoided that pattern because I (mistakenly) thought "throw;" was a rethrow, as in `catch (Exception ex) { throw ex; }` which destroys the stacktrace, but I ran a quick test and I was wrong.

Issue #529, opened 2020-01-25
Problem about module's name


ZQiu233:
ClrModule has 3 properties about name:
1.Name
2.FileName
3.AssemblyName
But it seems that they three return the same name.
If the assembly is loaded from byte[],then the three name return the full name of the assembly.
If the assembly is loaded from a .dll or .exe file,then the three name return the path of it.

Why this happend?I think at least "AssemblyName" should not be the path of an assembly.
Is it going on a wrong way?Or I'm wrong?

NextTurn:
`FileName` is `null` for dynamic modules (generated by `System.Reflection.Emit`).

`AssemblyName` and `Name` are usually identical since most assemblies only contain a single module with the same name.

ZQiu233:
> `FileName` is `null` for dynamic modules (generated by `System.Reflection.Emit`).
> 
> `AssemblyName` and `Name` are usually identical since most assemblies only contain a single module with the same name.

Got it.But how to get the full name of the assembly though.

NextTurn:
> Got it.But how to get the full name of the assembly though.

That could be available through a new `ClrAssembly` class, but I'm not sure if there are many use cases around it. @leculver

leculver:
I'm fine with adding a ClrAssembly class.  I was planning to do this for ClrMD 2.0 but the cases where the assembly is the module is _relatively_ rare and it's much more convenient to not have to go through another level of indirection.  It's true that the current implementation is likely not were we want to be.

Thanks for filing this issue.  This is something I want to tackle before we ship ClrMD 2.0 RTM.

NextTurn:
@leculver It seems a bit weird that `ClrModule.Name` returns the full path of PE file. It doesn't reflect the module's real name specified in *ModuleDef*.

I think it should instead return a short name which looks like the file name (with path removed).

leculver:
We are a bit hamstrung on how we report module "names" by the APIs that are provided by the dac.  I've cleaned this up somewhat, but Name and AssemblyName will continue to report filenames, because it's what the CLR diagnostics layer is reporting to us.  Essentially [GetPEFileName](https://github.com/dotnet/runtime/blob/master/src/coreclr/src/debug/daccess/request.cpp#L2047) doesn't _really_ give us a PEImage filename in all cases.  I suppose in a perfect world "name" would be the raw name in metadata as NextTurn suggests, but honestly I don't want to go digging through metadata (which might not be in crash dumps).

In the end, I've simplified this down to two things:

[...]

PR #649, opened 2020-04-30
Add ClrModule.SimpleName

NextTurn:
As proposed in #529.

This can be the real `ClrModule.Name`.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Builders/ModuleBuilder.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Builders/ModuleBuilder.cs
@@ -48,6 +48,8 @@ namespace Microsoft.Diagnostics.Runtime.Builders
             }
         }
 
+        public string? SimpleName { get; private set; }
+
         public string? AssemblyName
         {
             get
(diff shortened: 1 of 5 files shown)


NextTurn:
What are the conventions of loading and storing instance data?

Among `I*Data`, `I*Helper` and `Clr*`, which should:

- Initialize in constructors
- Lazily initialize fields
- Always fetch in getters

?

leculver:
I can't say I've put enough thought into making strong conventions.  What you've outlined is roughly what I was shooting for.

I would describe those interfaces more in the goal of what:

1.  Put a boundary between ClrMD implementation and the code which queries that data out of the dac.
2.  Most logic that deeply understands the CLR runtime should be hidden behind those interfaces.  Clr* implementations should just be querying data from I*Builder/I*Helper and not contain super deep logic.  (The reason here is that if we need to implement a new runtime type like .Net Native, we wouldn't really change ClrmdType at all, for example.  We'd either change RuntimeBuilder or create a new IRuntimeBuilder for .Net Native.)
3.  Initialize as much as we can in the constructor from basic data we can get quickly.  (IE when creating a ClrType, go ahead and request DacpMethodTableData and fill in as much as we can.)
4.  Lazily populate data that takes a long time to build (IE fields).
5.  Lazily populate data that takes up a lot of memory (like type names) so we keep memory usage low.

One alternative to all of these interfaces would have been to simply let Clrmd* objects directly use RuntimeBuilder.  If I went that route I would have made RuntimeBuilder an abstract base class and put much more thought into designing it.  However, there were a vocal set of folks who wanted to be able to mock Clr* objects, which his why I went this route.  I'm not sure I've really whether I've been successful or not with this design.  I'd guess not.

PR #651, opened 2020-04-30
Change ClrStatcField to require AppDomains

leculver:
The original design of ClrStaticField was to use the AppDomain associated with its parent type.  This made it very difficult to walk the value of that static field in all AppDomains as it's difficult to have ClrMD find and construct all of the concrete ClrTypes for each domain.  Instead we'll just require the user to pass the ClrAppDomain of the value to read.  They can always use "staticField.Parent.Module.AppDomain" to get the previous behavior.


NextTurn:
Hmm, in .NET Core I don't want to see `AppDomain` at all. :)

PR #654, opened 2020-05-02
Better type signature parsing

leculver:
This change essentially rewrites ClrmdField.GetTypeForFieldSig and instead implements that function as a real type signature parser.

- Move ClrmdField.GetTypeForFieldSig -> ITypeFactor.GetOrCreateTypeFromSignature.
- Implement most of ECMA 355 II.23.2.12 (type signature parsing) instead of the hacky version I wrote originally.
- Promote EnumerateGenericParameters to be a property instead of enum.
- Add ClrmdGenericType which is a "fake" constructed type for when we cannot properly look up a type.

During this change I promoted EnumerateGenericParameters from an enumeration to a property with a backing field.  This was a bad choice that I will revert before checkin, the original code was better.

Most types don't need that field, and we only need to go through GetOrCreateTypeFromSignature when the dac doesn't give us a real MethodTable for a legitimate field.  Marking this a draft until that's fixed.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
@@ -14,6 +14,7 @@ using System.Text;
 using System.Threading;
 using Microsoft.Diagnostics.Runtime.DacInterface;
 using Microsoft.Diagnostics.Runtime.Implementation;
+using Microsoft.Diagnostics.Runtime.Utilities;
 
 namespace Microsoft.Diagnostics.Runtime.Builders
 {
@@ -861,6 +862,10 @@ namespace Microsoft.Diagnostics.Runtime.Builders
         {
             CheckDisposed();
 
+            // We'll assume 'Class' is just System.Object
+            if (basicType == ClrElementType.Class)
+                basicType = ClrElementType.Object;
+
             ClrType?[]? basicTypes = _basicTypes;
             if (basicTypes is null)
             {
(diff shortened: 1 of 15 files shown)


src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
```diff
@@ -1002,6 +1012,136 @@ public ClrType CreateSystemType(ClrHeap heap, ulong mt, string typeName)
...
+                    return null;
+
+                if (!parser.GetToken(out int token))
+                    return null;
+
+
+                if (!parser.GetData(out int count))
+                    return null;
+
+                var builder = ImmutableArray.CreateBuilder<ClrType>(count);
```
leculver, line 1091: todo: Remove this builder.  We still need to fully walk the types here so that SigParser is left in a state pointing after the generic instantiation, but we don't need to build an immutable array since we don't use it.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrType.cs
```diff
@@ -86,7 +86,8 @@ public abstract class ClrType :
         /// <returns>True if this type is an object reference, false otherwise.</returns>
         public virtual bool IsObjectReference => ElementType.IsObjectReference();
 
-        public abstract IEnumerable<ClrGenericParameter> EnumerateGenericParameters();
```
leculver, line 89: todo: Put this back.  This was the correct choice from the beginning.

PR #657, opened 2020-05-04
Merge exception codes

NextTurn:
CA2000 is a really odd analyzer.


leculver:
> CA2000 is a really odd analyzer.

It is.  I'm not even sure it's useful, but I guess if I used something I forgot was disposable it'd be helpful.  If it continues to be annoying I'll globally disable it.

PR #658, opened 2020-05-09
Don't filter out managed assemblies in Linux data reader.

mikem8361:
It was a mistake to filter out the managed assemblies out of the ELF reader module list because there are coredumps generated with gdb running .NET Core 2.1.x with a stack overflow test that seems to have a bug or missing memory region where the ELF module can't be opened (ElfLoadedImage.Open() returns null). This causes most of the native modules (especially libcoreclr.so) to be missing. With this fix most dotnet-dump/SOS commands will work on this coredumps.

This is an regression caused by my module filtering change (sorry about that).

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
@@ -96,7 +96,7 @@ namespace Microsoft.Diagnostics.Runtime
         private readonly int _pointerSize;
         private readonly Architecture _architecture;
         private Dictionary<uint, IElfPRStatus> _threads;
-        private List<CoreModuleInfo> _modules;
+        private List<ModuleInfo> _modules;
         private readonly byte[] _buffer = new byte[512];
 
         public CoreDumpReader(string filename)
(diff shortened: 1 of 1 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
```diff
@@ -171,14 +165,10 @@ private IList<CoreModuleInfo> EnumerateAllModules()
                 // memory range overlaps with actual modules.
                 ulong interpreter = _core.GetAuxvValue(ElfAuxvType.Base);
 
-                _modules = new List<CoreModuleInfo>();
```
leculver, line 174: I think CoreModuleInfo was introduced in your last change unless my memory is wrong.  Is CoreModuleInfo still used or is it dead code?  If dead can you remove it?  (If it's still used then it's fine to leave it, I just wanted to make sure we still need the class and aren't leaving behind unused code.)

mikem8361:
Thanks Lee!

mikem8361:
Lee, did you ever get a 1.1 official build done/published with this change? Thanks.

leculver:
@mikem8361 So sorry about that, I'm juggling a LOT of work right now.  I did produce an official build but forgot to submit it to NuGet, thanks for the reminder!

Here's the package:  https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/1.1.126102.  It should be live in an hour or so.

Issue #660, opened 2020-05-12
ClrMD 2 EnumerateReferences much slower than ClrMD 1.1


jwdj:
For calculating the retention size of objects I use ClrMD 1.1 to visit all objects in a depth first search.
Out of curiositity I tried the beta version (2.0.0-beta.20230.2). Some code needed to be rewritten slightly but I like the API improvements.
But the performance of processing the same dump file was about 5 times slower.
I know that 2.0 is not finished yet, but before releasing 2.0 it would be wise to compare against ClrMD 1.1 visiting all objects and retrieving shallow size.

NextTurn:
I'd like to look into this but I need some guidance on performance profiling with *dotnet-trace*... :(

jwdj:
I recommend using [PerfView](https://github.com/Microsoft/perfview/releases). After recording a trace, you can go to the Flamegraph tab to see which methods take the most time. 
Here is an online [tutorial](https://mattwarren.org/2020/03/03/Analysing-.NET-Runtime-Startup-with-Flamegraphs/).
The traces are to big to send so I will try to summarize where more time is spent than before. This might take some time...

jwdj:
Well, the traces between ClrMD1.1 and ClrMD2.0 are very different. Hard to compare apples to apples. What I can tell is that ClrMD 2.0 suffers from I/O related slowdown. 83% is spent on reading the dump file. About 6% on my own code. So 11% on other ClrMD methods.
Samples spent in 1.1: 35893
Samples spent in 2.0: 155506

Here the methods that take the most time:
83%   microsoft.diagnostics.runtime!Microsoft.Diagnostics.Runtime.Windows.MinidumpMemoryReader.Read(unsigned int64,value class System.Span`1<unsigned int8>,int32&) (63809 inclusive samples, 37.88%)
63.8% microsoft.diagnostics.runtime!Microsoft.Diagnostics.Runtime.SpanExtensions.Read(class System.IO.Stream,value class System.Span`1<unsigned int8>)
60,7% mscorlib.ni!System.IO.FileStream.Read(Byte[], Int32, Int32)
17%   mscorlib.ni!System.IO.FileStream.set_Position(Int64)

drewnoakes:
@jwdj were you running this on .NET Framework?

https://adamsitnik.com/Span/#slow-vs-fast-span

NextTurn:
[...]

PR #661, opened 2020-05-13
Use memory mapped files for minidump

NextTurn:
Contributes to #660 

Interestingly enough, the simplest use of MMF increases performance significantly for my test case.

The MMF doesn't allow any file sharing by default so there can only be one stream at the same time. (Why do we need two streams?) I made `MinidumpReader` the `IDataReader` responsible for the disposal.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs
@@ -2,18 +2,21 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Microsoft.Diagnostics.Runtime.Windows;
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Collections.Immutable;
+using System.IO;
+using System.IO.MemoryMappedFiles;
 using System.Linq;
 using System.Runtime.InteropServices;
+using Microsoft.Diagnostics.Runtime.Windows;
 
 namespace Microsoft.Diagnostics.Runtime
 {
     internal sealed class MinidumpReader : IDataReader, IDisposable
     {
+        private readonly MemoryMappedFile _file;
+        private readonly Stream _stream;
         private readonly Minidump _minidump;
         private IMemoryReader? _readerCached;
 
@@ -22,11 +25,6 @@ namespace Microsoft.Diagnostics.Runtime
         public string DisplayName { get; }
 
         public IMemoryReader MemoryReader => _readerCached ??= _minidump.MemoryReader;
-
-        public MinidumpReader(string crashDump)
-            : this(crashDump, File.OpenRead(crashDump ?? throw new ArgumentNullException(nameof(crashDump))))
-        {
-        }
 
         public MinidumpReader(string crashDump, Stream stream)
         {
@@ -38,7 +36,10 @@ namespace Microsoft.Diagnostics.Runtime
 
             DisplayName = crashDump;
 
-            _minidump = new Minidump(crashDump, stream);
+            stream.Dispose();
+            _file = MemoryMappedFile.CreateFromFile(crashDump, FileMode.Open);
+            _stream = _file.CreateViewStream();
+            _minidump = new Minidump(crashDump, _stream);
 
             Architecture = _minidump.Architecture switch
             {
@@ -62,9 +63,9 @@ namespace Microsoft.Diagnostics.Runtime
 
         public void Dispose()
         {
-            _minidump.Dispose();
+            _stream.Dispose();
+            _file.Dispose();
         }
-
 
         public IEnumerable<ModuleInfo> EnumerateModules()
         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs
```diff
@@ -38,7 +36,10 @@ public MinidumpReader(string crashDump, Stream stream)
 
             DisplayName = crashDump;
 
-            _minidump = new Minidump(crashDump, stream);
+            stream.Dispose();
```
jwdj: Is it safe here to dispose of a stream you don't own? Probably better to leave the disposing to the caller.

NextTurn: See the second commit. :)

leculver:
Thanks!  I haven't had a chance to review it yet, but one concern I have is memory usage when inspecting large crash dumps.

I haven't used `MemoryMappedFile` before, so could you check what happens when you load a very large x86 dump file of a process (like with the /3gb bit set)?  Does it consume the entire address space or just a reasonable amount of it?

One of the reasons I started with a Stream (and hoped eventually to do some memory mapping) is that loading the whole file into the address space would run OOM in some x86 scenarios.  This is a super common use-case for ClrMD at Microsoft (inspecting Visual Studio dumps).  I want to make sure we don't eat too much memory.

NextTurn:
I'd guess it will take up the entire address space. Let's make this a 64-bit specific feature?

leculver:
> I'd guess it will take up the entire address space. Shall we make this a 64-bit specific feature?

I also am a little concerned about 32gb dumps out of SQL.  I would suggest instead making it a feature based on dump size.  For example, if the filesize of the dump we are loading is bigger than 256 megs on x86 or 4gb on x64 we fall back to streams?  (I picked those numbers at random, please feel free to choose better ones.)

We can always go back and let the user tune the numbers themselves later through `CachingOptions`.

leculver:
I will say my original plan for memory mapped files ages ago was to use a Least Recently Used cache to map segments into memory and keep them there for a while.  It'd let us use memory mapped files larger than we would allow with a filesize based approach, but it's also a lot more complicated to code up.

I'd say long term I'd still like to do that, but since it doesn't affect the public surface area of the API it's something I could always go back and do later.

NextTurn:
> to use a Least Recently Used cache to map segments into memory and keep them there for a while.

Yeah it would be my preference as well.

> I would suggest instead making it a feature based on dump size. For example, if the filesize of the dump we are loading is bigger than 256 megs on x86 or 4gb on x64 we fall back to streams

This can act as a transition. The size threshold on x64 can be much larger as the address space limit is in the TB level (128 TB on Windows 10).

I'll hold this for a while until you finish the perf testing to make sure this really improves performance.

leculver:
Sounds great!  I'll take a look at this as soon as I can get back to working on ClrMD.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs
NextTurn: No worries. The caller, `DataTarget`, was designed to transfer the ownership of the stream to the data reader. CA2000 doesn't know that either so we silenced it.

https://github.com/microsoft/clrmd/blob/ea1d957850c92a4f9a2ac73b5b72517bb83be0c6/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs#L245

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Minidump/MinidumpReader.cs
```diff
@@ -38,7 +36,17 @@ public MinidumpReader(string crashDump, Stream stream)
 
             DisplayName = crashDump;
 
-            _minidump = new Minidump(crashDump, stream);
+            if (new FileInfo(crashDump).Length <= (Environment.Is64BitProcess ? 0x100_0000_0000 : 0x1000_0000))
+            {
+                _file = MemoryMappedFile.CreateFromFile(stream, null, 0, MemoryMappedFileAccess.Read, HandleInheritability.None, leaveOpen: false);
+                _stream = _file.CreateViewStream(0, 0, MemoryMappedFileAccess.Read);
```
leculver, line 42: Since we aren't saving `stream` in this case, we should call `stream.Dispose()` on the parameter.

NextTurn: The MMF takes the stream and will dispose of it (`leaveOpen: false`).

leculver: Ahhh I see thanks!

leculver:
@NextTurn Still a draft, or ready to merge?

NextTurn:
@leculver Please feel free to merge.

PR #677, opened 2020-05-27
Don't read gcdesc pointers one at a time

leculver:
The previous code would individually read one pointer at a time as we walked objects for references.  Now read in the entire object into one buffer before walking the object.


leculver:
Contributes to https://github.com/microsoft/clrmd/issues/660.

PR #683, opened 2020-05-28
Nits

NextTurn:


src/Microsoft.Diagnostics.Runtime.Tests/src/GCRootTests.cs
```diff
@@ -133,9 +130,9 @@ public void EnumerateGCRefsArray()
 
             Assert.Equal("System.Object[]", obj.Type.Name);
 
-            ClrObject[] refs = obj.EnumerateReferences(false).ToArray();
-            Assert.Single(refs);
-            Assert.Equal("DoubleRef", refs[0].Type.Name);
+            IEnumerable<ClrObject> refs = obj.EnumerateReferences(false);
```
leculver, line 133: This makes debugging it harder.  In tests, I prefer to force the enumeration into an array so that it's easily to inspect the value while stepping through.  The remaining code is fine though.

```suggestion
            ClrObject[] refs = obj.EnumerateReferences(false).ToArray();
```

PR #689, opened 2020-05-28
Support MacOS PlatformFunctions.

mikem8361:
This is so dotnet-dump can support the ELF core dumps generated by the new createdump on MacOS.


leculver:
Published here, thanks!

https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/1.1.127808

PR #695, opened 2020-05-30
Save reflection calls

NextTurn:


leculver:
@ryanmolden as an FYI.

Issue #698, opened 2020-05-30
The Linux symlink code in GetOrCreateClrVersions() leaks temp directory/files


mikem8361:
I should have added some kind of cleanup with I added this DAC/symlink workaround months ago.  The result is leaked tmp directories/files on Linux:

https://github.com/microsoft/clrmd/blob/02b86fb3e87381b18d808c03a77c4e0cb2a3987c/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs#L167

These directories and files could be cleaned up during DataTarget.Dispose() with something like (saving a way the directory name in _tempDirectory:

        private void CleanupTempDirectory()
        {
            if (_tempDirectory != null)
            {
                try
                {
                    foreach (string file in Directory.EnumerateFiles(_tempDirectory))
                    {
                        File.Delete(file);
                    }
                }
                catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException)
                {
                }
            }
        }

PR #699, opened 2020-05-30
Add Linux symlink temp directory cleanup.

mikem8361:
Issue: https://github.com/microsoft/clrmd/issues/698

Added 3.1.0 runtime to global.json so the tests can run.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,11 @@
 {
   "tools": {
-    "dotnet": "5.0.100-preview.5.20251.2"
+    "dotnet": "5.0.100-preview.5.20251.2",
+    "runtimes": {
+      "dotnet": [
+        "3.1.0"
+      ]
+    }
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "5.0.0-beta.20261.9"
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
@@ -30,6 +30,7 @@ namespace Microsoft.Diagnostics.Runtime
         private ModuleInfo[]? _modules;
         private string? _symlink;
         private readonly Dictionary<string, PEImage?> _pefileCache = new Dictionary<string, PEImage?>(StringComparer.OrdinalIgnoreCase);
+        private readonly object _sync = new object();
 
         /// <summary>
         /// Gets the data reader for this instance.
@@ -176,10 +177,13 @@ namespace Microsoft.Diagnostics.Runtime
                     if (File.Exists(dacLocation))
                     {
                         // Works around issue https://github.com/dotnet/coreclr/issues/20205
-                        _symlink = Path.GetTempFileName();
-                        if (LinuxFunctions.symlink(dacLocation, _symlink) == 0)
+                        lock (_sync)
                         {
-                            dacLocation = _symlink;
+                            _symlink = Path.GetTempFileName();
+                            if (LinuxFunctions.symlink(dacLocation, _symlink) == 0)
+                            {
+                                dacLocation = _symlink;
+                            }
                         }
                     }
                     else
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Diagnostics.Runtime
         private bool _disposed;
         private ImmutableArray<ClrInfo> _clrs;
         private ModuleInfo[]? _modules;
-        private string? _tempDirectory;
+        private string? _symlink;
         private readonly Dictionary<string, PEImage?> _pefileCache = new Dictionary<string, PEImage?>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
@@ -90,7 +90,16 @@ namespace Microsoft.Diagnostics.Runtime
                     _pefileCache.Clear();
                 }
 
-                CleanupTempDirectory();
+                if (_symlink != null)
+                {
+                    try
+                    {   
+                        File.Delete(_symlink);
+                    }
+                    catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException)
+                    {
+                    }
+                }
 
                 _target.Dispose();
                 _disposed = true;
@@ -167,14 +176,10 @@ namespace Microsoft.Diagnostics.Runtime
                     if (File.Exists(dacLocation))
                     {
                         // Works around issue https://github.com/dotnet/coreclr/issues/20205
-                        int processId = Process.GetCurrentProcess().Id;
-                        _tempDirectory = Path.Combine(Path.GetTempPath(), "clrmd" + processId);
-                        Directory.CreateDirectory(_tempDirectory);
-
-                        string symlink = Path.Combine(_tempDirectory, dacFileName);
-                        if (LinuxFunctions.symlink(dacLocation, symlink) == 0)
+                        _symlink = Path.GetTempFileName();
+                        if (LinuxFunctions.symlink(dacLocation, _symlink) == 0)
                         {
-                            dacLocation = symlink;
+                            dacLocation = _symlink;
                         }
                     }
                     else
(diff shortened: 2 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -195,6 +199,28 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
...
+        /// Cleans up the temporary directory and symlinks for the Linux DAC workaround.
+        /// </summary>
+        private void CleanupTempDirectory()
+        {
+            string? tempDirectory = Interlocked.Exchange(ref _tempDirectory, null);
+            if (tempDirectory != null)
+            {
+                try
+                {
+                    foreach (string file in Directory.EnumerateFiles(tempDirectory))
```
NextTurn, line 212:
```suggestion
                    Directory.Delete(tempDirectory, recursive: true);
```

mikem8361:
@leculver, @NextTurn what do you guys think about adding a public DataTarget "TempDirectory" property that lazily creates it like https://github.com/microsoft/clrmd/blob/12da3fcbf17df12257a23de274108a6fd6698baa/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs#L167

dotnet-dump, SOS, etc. creates a similar per-process temp directory and it would be nice to share it and the cleanup.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -164,10 +168,10 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
                     {
                         // Works around issue https://github.com/dotnet/coreclr/issues/20205
                         int processId = Process.GetCurrentProcess().Id;
-                        string tempDirectory = Path.Combine(Path.GetTempPath(), "clrmd" + processId);
-                        Directory.CreateDirectory(tempDirectory);
+                        _tempDirectory = Path.Combine(Path.GetTempPath(), "clrmd" + processId);
+                        Directory.CreateDirectory(_tempDirectory);
 
-                        string symlink = Path.Combine(tempDirectory, dacFileName);
+                        string symlink = Path.Combine(_tempDirectory, dacFileName);
```
leculver, line 174: This design creates a subtle race between two ClrMD processes inspecting the same target process.

If Process A is in `CleanupTmpDirectory` in or just before `Directory.Delete` and Process B has just returned from `LinuxFunctions.symlink` then Process A may delete our symlink before we have a chance to open it, leading to an odd exception in Process B when it can't find the dac symlink we created.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -195,6 +199,24 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
             return _clrs;
         }
 
+        /// <summary>
+        /// Cleans up the temporary directory and symlinks for the Linux DAC workaround.
+        /// </summary>
+        private void CleanupTempDirectory()
+        {
+            string? tempDirectory = Interlocked.Exchange(ref _tempDirectory, null);
```
leculver, line 207: Why do we need `Interlocked.Exchange` here?  The user shouldn't call Dispose twice and it won't hurt if they do.  Using `Interlocked.Exchange` doesn't _hurt_ here but it makes the code a little harder to understand, as in someone walking up to the codebase might think this is part of some multithreaded locking scheme when it's not.

I'd normally recommend replacing this line with `if (_tempDirectory != null && Directory.Exists(_tempDirectory)` and never assigning `_tempDirectory` back to null.  It's ok if the user mistakenly calls Dispose, even from multiple threads, since the catch should protect us below.  (However, as mentioned above there's still a race in this overall design, so this is general feedback that I don't expect you to address before we figure out a design that doesn't race.)

global.json
```diff
@@ -1,6 +1,11 @@
 {
   "tools": {
```
leculver, line 2: What is the purpose of the change to global.json?  Is that expected in this review?

leculver: Oh I missed that comment.

leculver:
> what do you guys think about adding a public DataTarget "TempDirectory" property that lazily creates it like
> https://github.com/microsoft/clrmd/blob/12da3fcbf17df12257a23de274108a6fd6698baa/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs#L167
> 
> dotnet-dump, SOS, etc. creates a similar per-process temp directory and it would be nice to share it and the cleanup.

I worry that we will create hard to track races between tools creating and cleaning up files within the same directory.  I wonder if we should go in a different direction.  Instead of creating a directory for symlinks, it might make more sense to not create a directory at all and instead do something like:

    _symlink = Path.GetTempFileName();
    if (LinuxFunctions.symlink(dacLocation, _symlink) == 0)
        dacLocation = symlink;

Then deleting our `_symlink` temp file in `Dispose`.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
NextTurn: Aha? The temporary directory has the ClrMD process' ID in its name.

mikem8361:
I liked your last suggestion and push a new version.  It is a lot simpler.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
leculver: Ah you are right, I didn't notice it's the current process and not the target process.  What if two different threads in the same process create a DataTarget and clean it up?  I'm not sure the behavior of `LinuxFunctions.symlink` will be if two threads race to call that at the same time, will that fail and we won't set the dacLocation?  I think that same cleanup issues happens if two threads in the same process races.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -195,6 +204,13 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
             return _clrs;
         }
 
+        /// <summary>
+        /// Cleans up the temporary directory and symlinks for the Linux DAC workaround.
+        /// </summary>
+        private void CleanupTempDirectory()
```
leculver, line 210: We can remove this now right?

mikem8361: Oops, I forgot to delete it.

mikem8361:
I think with the new code, if two different threads enumerate the ClrInfos on the same DataTarget instance there could be a race so maybe I should put a lock around the creating the symlink:

```
              lock()
              {
                    if (_symlink == null)
                   {
                         _symlink = Path.GetTempFileName();
                        if (LinuxFunctions.symlink(dacLocation, _symlink) == 0)
                        {
                            dacLocation = _symlink;
                        }
                   }
              }
```

leculver:
Yep good point on the race.

You could also use Interlocked.CompareExchange to avoid a lock here, just make sure the symlink is fully created before publishing the value, and delete the temp file for the thread that lost the race.

Is there any way to load two different .Net Core runtimes into a single Linux process?  For example, two runtimes living in parallel directories?  We don't have to address that problem today, but I'd just point out that the current design only works in a non SxS scenario.  It's a problem for a different day if this is possible.

mikem8361:
Like you said before interlocked is harder to follow and I don't think a lock could cause any problems/deadlocks. I'm going to add a lock around the code unless you object.

Multiple .NET Core processes side-by-side isn't supported but actually might work.  The old hosting APIs (coreclr_initialize(), etc.) don't prevent it (the new ones may). It hasn't been tested in a long time.

We can have a desktop CLR and .NET Core runtime in the same process but that doesn't apply here.

mikem8361:
One thing I didn't think about is that `lock(this)` isn't a great practice because outside code can lock the DataTarget instance might cause problems/deadlocks.  So adding a private object instance is the usual practice if there isn't already some private. Seems overkill to just allocate another object instance for a lock, but I might be overthinking this.

leculver:
No worries, and definitely don't overthink it.  Just add a new `private readonly object _sync = new object();` to DataTarget.  I can always move it to interlocked later if need be.

NextTurn:
@leculver @mikem8361 

(I'm cleaning up my ancient email backlog.) This doesn't actually work. `Path.GetTempFileName` will create the file and `symlink` will fail with `EEXIST`.

Besides, the synchronization looks odd... If there is a race between two threads the temp file will be created twice?

PR #704, opened 2020-06-02
Change ObjectSet's fields to properties

NextTurn:
Fixes an ancient issue.

CA1819 recommends to return a collection but it's gonna hurt the performance.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ObjectSet.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ObjectSet.cs
@@ -16,17 +16,19 @@ namespace Microsoft.Diagnostics.Runtime
         /// <summary>
         /// The ClrHeap this is an object set over.
         /// </summary>
-        protected ClrHeap _heap;
+        protected ClrHeap Heap { get; }
 
         /// <summary>
         /// The minimum object size for this particular heap.
         /// </summary>
-        protected readonly int _minObjSize;
+        protected int MinObjSize { get; } = IntPtr.Size * 3;
 
         /// <summary>
         /// The collection of segments and associated objects.
         /// </summary>
-        protected HeapHashSegment[] _segments;
+#pragma warning disable CA1819 // Properties should not return arrays
+        protected HeapHashSegment[] Segments { get; }
+#pragma warning restore CA1819 // Properties should not return arrays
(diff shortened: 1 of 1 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ObjectSet.cs
```diff
@@ -16,17 +16,19 @@ public class ObjectSet
...
         /// </summary>
-        protected readonly int _minObjSize;
+        protected int MinObjSize { get; } = IntPtr.Size * 3;
 
         /// <summary>
         /// The collection of segments and associated objects.
         /// </summary>
-        protected HeapHashSegment[] _segments;
+#pragma warning disable CA1819 // Properties should not return arrays
+        protected HeapHashSegment[] Segments { get; }
```
leculver, line 30: Maybe we should just make all of these public and Segments an ImmutableArray?

Happy to merge as-is, but that would also satisfy the bug while making things a bit more open.

NextTurn: `ImmutableArray` implies the array elements won't change. Maybe something read-only like `ReadOnlyMemory`?

leculver: I don't read it that way.  Semantically ImmutableArray only enforces an immutable array of mutable objects.

In C++ turns, ImmutableArray is like `const Foo *` a "constant array of Foo objects" and not `Foo * const` which would be an "immutable array of mutable Foo objects" (and also not "const Foo * const" which is immutable/immutable).

leculver: Heh pretty sure I wrote that backwards.  It's been a really long day.

NextTurn: What???

NextTurn: We've made all of these read-only so it won't hurt to make them public.

leculver: Makes sense, thanks!

PR #708, opened 2020-06-02
Process ID should be signed

NextTurn:
Even `pid_t` is defined as signed.


src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -84,7 +84,7 @@ private void Dispose(bool _)
 
             if (_suspended)
             {
-                int status = (int)ptrace(PTRACE_DETACH, (int)ProcessId, IntPtr.Zero, IntPtr.Zero);
+                int status = (int)ptrace(PTRACE_DETACH, ProcessId, IntPtr.Zero, IntPtr.Zero);
```
leculver, line 87: Unrelated to this change, maybe we should simply define `ptrace` and `process_vm_readv` as returning int instead of casting?

NextTurn: Yeah we could do that.

PR #709, opened 2020-06-03
Optimize process readers

NextTurn:
- All segments/regions must be in ascending order by begin address.
- The first one is located by a binary search.
- Read next only if no partial read occurs.
- The subsequent ones must be adjacent and readable (only checked by process readers).

Did I miss anything?

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
     {
         private readonly int _task;
 
-        private List<MemoryRegion> _memoryRegions;
+        private ImmutableArray<MemoryRegion>.Builder _memoryRegions;
 
         private bool _suspended;
         private bool _disposed;
@@ -26,13 +26,14 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
                 throw new ArgumentException("The process is not running");
 
             ProcessId = processId;
-            _memoryRegions = LoadMemoryRegions();
 
             int kr = Native.task_for_pid(Native.mach_task_self(), processId, out int task);
             if (kr != 0)
                 throw new ClrDiagnosticsException($"task_for_pid failed with status code 0x{kr:x}");
 
             _task = task;
+
+            _memoryRegions = LoadMemoryRegions();
 
             if (suspend)
             {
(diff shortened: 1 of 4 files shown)


leculver:
Overall this is a great cleanup but it doesn't consider the case where segments overlap.  Once that's fixed we should definitely get this checked in.

I've definitely Minidumps with overlapping segments happen happen in the real world but I haven't encountered a coredump created this way.  I think it can still happen in coredumps...nothing would prevent someone from putting overlapping sections in, so I think we should handle it there as well just in case.

Here's an example, let's say these MINIDUMP_SEGMENTS ranges are defined and the user requests to read the memory range `[1000-2000]`:

```
_segments[index] - [va start, va end]

_segments[0] = [1000, 1500] <== contains the first 500 bytes
_segments[1] = [1000, 1200]   <== This completely overlaps 1, we need to skip it
_segments[2] = [1400, 1900]   <== This partially overlaps 1, we need to read [1500-1900]
_segments[3] = [1500, 2000]   <== This contains the remaining bytes, we need to read [1900-2000]
```

I haven't fully reviewed the change, but this code in UncachedMemoryReader would fail:

``` csharp
                        if (bytesRead != 0 && segment.VirtualAddress != prevEndAddress)
                            break;
```

I think overall your solution is better (not continuing to binary search in some cases), but instead of the code `segment.VirtualAddress != preEndAddress`, we have to keep enumerating segments and trying to fill the buffer until either we've read all of the requested bytes, or until the segment we see in our linear walk no longer lives in the original `[1000, 2000]` requested range.

Lastly, we _may_ also need to change `int i = GetSegmentContaining(address);` to be followed by `while (i > 0 && _segments[i-1].Contains(address)) i--;` to ensure we've found the bottom most segment.  In the example above if the binary search stops on _segment[1] instead of _segment[0], then we will fail to read all of the memory.

@ryanmolden as an FYI since we were chatting about this last week.  I think NextTurn's fix for this issue here is better continually binary searching like I had in my code.

NextTurn:
Thanks @leculver! Segment overlap is definitely something I didn't consider. I've excluded changes to dump readers from this PR and will open a separate one when I get to it.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
```diff
@@ -401,13 +416,13 @@ internal static class Native
             }
         }
 
-        internal sealed class MemoryRegion
+        internal struct MemoryRegion
```
leculver, line 419: Can this struct be marked readonly?

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -274,69 +274,87 @@ private void LoadThreads()
...
-            if (startIndex < 0)
-            {
-                return 0;
+                bytesToRead -= regionSize;
+                bytesReadable = regionSize;
+                prevEndAddr = entryEndAddr;
             }
 
-            for (int i = startIndex + 1; i < _memoryMapEntries.Count; i++)
+            for (i += 1; i < _memoryMapEntries.Count; i++)
```
leculver, line 307: This block of code looks similar (or a duplicate of) the code in `MacOSProcessDataReader.cs` (I didn't look very closely to confirm they are the same).

If these are the same, is there any way we could put this in a static helper method, extension method, etc, so that we don't have it duplicated?  I don't like having two copies of the same algorithm that have to be updated in parallel.

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -274,69 +274,87 @@ private void LoadThreads()
...
-                bytesRequested -= regionSize;
+                bytesToRead -= regionSize;
                 bytesReadable += regionSize;
                 prevEndAddr = entryEndAddr;
             }
 
             return bytesReadable;
         }
 
+        private int GetMapEntryContaining(ulong address)
```
leculver, line 332: Same question here with `GetRegionContaining`/`GetMapEntryContaining`.  It would be better if we could find a way to merge that code together, maybe by unifying MemoryMapEntry/MemoryRegion...even if that means having to read in one form and convert to another.

NextTurn:
As different types are used I couldn't think of a safe way to merge these without templates in C++. I'll push a commit to demonstrate an unsafe way to achieve this.

leculver:
No, if it doesn't make sense to merge them then let's not merge them.  I only wanted to make sure and didn't have time today to look closer.  Let's merge this one as-is.  Thanks!

Issue #475, opened 2019-12-24
Layout/Size related naming


NextTurn:
As discussed in some other threads, we need more appropriate layout/size related names.

For reference, there are three major layouts in CoreCLR:

https://github.com/dotnet/runtime/blob/bd6c81a944f609b88908864827b7797c6b127c5b/src/coreclr/src/vm/peimage.h#L287

https://github.com/dotnet/runtime/blob/bd6c81a944f609b88908864827b7797c6b127c5b/src/coreclr/src/vm/peimagelayout.h#L40

- Flat layout

Specifies that the image was directly mapped via a single `mmap`/`CreateFileMapping` call (file layout).

- Mapped layout

Specifies that the sections of the image were mapped to their corresponding virtual addresses, as if it was loaded with `LoadLibrary` (not file layout).

- Loaded layout

Windows specific: Specifies that the image was loaded using `LoadLibrary` (not file layout).

leculver:
I'm not going to be able to get to this in the 2.0 RTM timeline.  If this requires breaking changes to the API surface it'll likely have to be addressed in 2.1.

leculver:
@NextTurn I am not planning to fix this issue before RTM, but can you think of any function name/enum name/etc we should rename before we ship?

NextTurn:
No good ideas. I haven't had time to visit all public APIs. But super long names like `returnOnlyFullyUniquePaths` won't be what I would prefer. ;)

As for this specific issue, the "size" here was referring to `ModuleInfo.FileSize` which could be a virtual size sometimes. Now the problem has automatically disappeared since we only have `ModuleInfo.IndexFileSize` for PE images.

leculver:
I'm going to close this issue for now since we don't have a concrete issue outlined here.  Please feel free to open a new one or reply back here if there's an actionable bug related to this.

PR #715, opened 2020-06-03
Change ModuleInfo._isVirtual to an enum

leculver:
Start tracking peimage load layout by enum in accordance with https://github.com/dotnet/runtime/blob/bd6c81a944f609b88908864827b7797c6b127c5b/src/coreclr/src/vm/peimage.h#L287.

Note that this change is NOT intended to make this work correctly for all file layouts, but instead to change the public surface area of ClrMD to support future improvments here before RTM.

Contributes to https://github.com/microsoft/clrmd/issues/475.


leculver:
@NextTurn Any thoughts on this?

NextTurn:
`isVirtual` is not too bad. In most cases we only need a boolean to decide whether to use virtual address/size or file offset/size. I assume that the "mapped layout" and the "loaded layout" should always be equivalent.

PR #719, opened 2020-06-04
Update Samples/Documentation

leculver:
Also renamed `ClrObject.IsValidObject` -> `IsValid` to make it more in line with other functions.

Closes https://github.com/microsoft/clrmd/issues/417.


doc/GettingStarted.md
```diff
@@ -8,75 +8,29 @@ how to create a `ClrRuntime` instance from a crash dump and a dac.
...
-For example, how a GC Segment is laid out, how object references inside an
-object can be found, and so on.
-
-ClrMD is an attempt to bridge the gap between private APIs and tool writers. The
-API itself is built on top of the Dac private API, but abstracts away all of the
-gory details you are required to know to use them successfully. Better yet,
-ClrMD is a publicly shipping, documented API. You may build your own programs
-on top of it and ship them outside of Microsoft (which is not true if you build
-on top of the raw Dac private APIs).
+All of .NET debugging support is implemented on top of a dll we call "The Dac" (usually named `mscordacwks.dll`).  This library is the building block for our private debugging APIs which should not be directly used.  The dotnet runtime team has implemented several public APIs using these private ones for anyone to consume.  This includes the ClrMD library, which is a process and runtime inspection API as well as `ICorDebug`, the official debugger API for CLR.  Note that ClrMD is not a debugging api and does not support setting breakpoints, stepping, and so on.
```
NextTurn, line 11: Nits:

```suggestion
All of .NET debugging support is implemented on top of a dll we call "The DAC" (usually named *mscordacwks.dll*/*mscordaccore.dll*).  This library is the building block for our private debugging APIs which should not be directly used.  The .NET runtime team has implemented several public APIs using these private ones for anyone to consume.  This includes the ClrMD library, which is a process and runtime inspection API as well as `ICorDebug`, the official debugger API for CLR.  Note that ClrMD is not a debugging API and does not support setting breakpoints, stepping, and so on.
```

(GitHub doesn't render diff for long lines.)

PR #728, opened 2020-06-05
Fix threading bug in HeapSegmentDataCache for the ArrayPoolBasedCache…

leculver:
Cherry-pick fix from @ryanmolden.


leculver:
Contributes to #737.

Issue #727, opened 2020-06-05
OOM when calling Flush


leculver:
The new data readers in ClrMD currently have a concurrency issue which causes some memory leaks under extremely heavy load.  This seems to happen primarily when ClrMD is used from multiple threads, with a smaller than default cache size, and under heavy usage.

This is my top priority to fix.

leculver:
Contributes to https://github.com/microsoft/clrmd/issues/727.

leculver:
I've tracked this down to benchmarks calling `ClrRuntime.FlushCachedData`.  There were a few reasons why SOSHandleEnume would leak memory.  Some are fixed in #756 and some are only worked around because parts of the leak are unavoidable.

PR #730, opened 2020-06-05
Fix for AWE oom

leculver:
Fixes https://github.com/microsoft/clrmd/issues/727.


src/Microsoft.Diagnostics.Runtime/src/Windows/AWEBasedCacheEntry.cs
```diff
@@ -330,12 +330,12 @@ public override bool PageOutData()
             }
 
             // Revert to our minimum size
-            CurrentSize = MinSize;
+            CurrentSize = Math.Min(MinSize, CurrentSize - (uint)sizeRemoved);
```
leculver, line 333:
```suggestion
            CurrentSize = Math.Max(MinSize, CurrentSize - (uint)sizeRemoved);
```

Issue #736, opened 2020-06-07
Allow more ways to create a DataTarget not just by filename


0blu:
Hey,
after looking into the code a bit, I've noticed that only the _isTinyDump_ in `Minidump` with `MemoryMappedFile.CreateFromFile` requires a real `FileStream`, all other segments are using a normal `Stream`.
Would it be possible to remove the `FileStream` restriction and `File.Exists(crashDump)` from the code in favor of a new additional factory like `DataTarget.LoadFromStream(string displayName, Stream stream, CacheOptions? cacheOptions = null)`?

NextTurn:
Ideally `LoadDump` could accept a `Stream`, an `ImmutableArray<Byte>` or even a `Byte*` just like what `PEReader` does.

PR #738, opened 2020-06-07
Create DataTarget by Stream

0blu:
Part of #736.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
@@ -2,8 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Diagnostics.Runtime.Implementation;
+using Microsoft.Diagnostics.Runtime.Linux;
+using Microsoft.Diagnostics.Runtime.MacOS;
+using Microsoft.Diagnostics.Runtime.Utilities;
 using System;
-using System.Buffers;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
@@ -12,11 +15,6 @@ using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using System.Threading;
-using Microsoft.Diagnostics.Runtime.Implementation;
-using Microsoft.Diagnostics.Runtime.Linux;
-using Microsoft.Diagnostics.Runtime.MacOS;
-using Microsoft.Diagnostics.Runtime.Utilities;
 
 namespace Microsoft.Diagnostics.Runtime
 {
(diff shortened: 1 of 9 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -237,88 +235,102 @@ public IEnumerable<ModuleInfo> EnumerateModules()
...
-                throw new FileNotFoundException($"Could not open dump file '{filePath}'.", filePath);
-
-            cacheOptions ??= new CacheOptions();
-
-            (FileStream stream, DumpFileFormat format) = OpenDump(filePath);
             try
             {
+                if (displayName is null)
+                    throw new ArgumentNullException(nameof(displayName));
+                if (stream is null)
```
leculver, line 252: You also need to check that `stream.CanSeek` and if not throw here, since we require seekable streams.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -237,88 +235,102 @@ public IEnumerable<ModuleInfo> EnumerateModules()
...
+                    throw new ArgumentNullException(nameof(stream));
+
+                cacheOptions ??= new CacheOptions();
+
+                DumpFileFormat format = ReadFileFormat(stream);
```
leculver, line 257: The entire library assumes the dump is at the beginning of the stream.  You should either set `stream.Position = 0` or require the user set the position to 0 and throw an exception if `stream.Position != 0`.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -237,88 +235,102 @@ public IEnumerable<ModuleInfo> EnumerateModules()
...
         /// </summary>
-        /// <param name="filePath">The path to the dump file.</param>
-        /// <param name="cacheOptions">The caching options to use </param>
-        /// <returns>A <see cref="DataTarget"/> for the given dump file.</returns>
-        public static DataTarget LoadDump(string filePath, CacheOptions? cacheOptions = null)
+        /// <param name="displayName">The name of this DataTarget, might be used in exceptions.</param>
+        /// <param name="stream">The stream that should be used. Ownership will be transferred to the DataTarget.</param>
+        /// <param name="cacheOptions">The caching options to use. (Only used for FileStreams)</param>
+        /// <returns>A <see cref="DataTarget"/> for the given dump.</returns>
+        public static DataTarget LoadDumpFromStream(string displayName, Stream stream, CacheOptions? cacheOptions = null)
```
leculver, line 246: Should this just be named `LoadDump` instead?  It's fine to have an overload that takes a stream, but I don't think it needs a unique name.

leculver, line 246: If we provide a `Stream` constructor, we should also take a `leaveOpen` flag similar to other C# apis.  That would have to be plumbed through all of the data readers to make sure that they respect whether to leave the stream open or dispose of it.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -239,86 +237,102 @@ public IEnumerable<ModuleInfo> EnumerateModules()
...
+                    DumpFileFormat.Minidump => new MinidumpReader(displayName, stream, cacheOptions, leaveOpen),
+                    DumpFileFormat.ElfCoredump => new CoredumpReader(displayName, stream, leaveOpen),
 
                     // USERDU64 dumps are the "old" style of dumpfile.  This file format is very old and shouldn't be
                     // used.  However, IDebugClient::WriteDumpFile(,DEBUG_DUMP_DEFAULT) still generates this format
                     // (at least with the Win10 system32\dbgeng.dll), so we will support this for now.
-                    DumpFileFormat.Userdump64 => new DbgEngDataReader(filePath, stream),
-                    DumpFileFormat.CompressedArchive => throw new InvalidDataException($"File '{filePath}' is a compressed archived instead of a dump file."),
-                    _ => throw new InvalidDataException($"File '{filePath}' is in an unknown or unsupported file format."),
+                    DumpFileFormat.Userdump64 => new DbgEngDataReader(displayName, stream, leaveOpen),
```
0blu, line 270: I've just noticed that Userdump64 cannot be created by a stream :/

NextTurn:
Don't forget `isVirtual`. ;)

0blu:
Are you sure I should add `isVirtual` to this pull request?
I literally have no idea where this is used other than the memory reader.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
leculver: It's fine, just throw a NotSupportedException for that case.

leculver:
> Are you sure I should add `isVirtual` to this pull request?
> I literally have no idea where this is used other than the memory reader.

You don't need to handle anything related to that in this PR.

FYI, I had always intended to provide a way to open dumps via stream, but I haven't gotten around to it yet because of all the little details related to stuff like `leaveOpen`, the dump being at an offset instead of 0 in the stream, and other corner cases.  I'm happy you are tackling this, but there might be slightly more work than you expected.  =)

The only work that needs to be done before this is merged is outlined in my feedback above.

0blu:
Okay, thanks for all the feedback, I am glad that everything went so fast. :)
I think the current build issue comes from the master branch.

leculver:
Yeah that's my doing.  =)

leculver:
I'm still debating what to name the method.  I think "LoadDump" is correct and what I want, but I want a little time to think about it so it might be a week or so before I ship a new build.  Thanks for the pr!

0blu:
Great! 😃

PR #751, opened 2020-06-10
ClrObject extension methods

raffaeler:
This PR contains two extension methods for ClrObject and related tests:

1. `public static string GetStringValue(this ClrObject clrObject, int maxLength = int.MaxValue)`
This is a convenient method to return the string value when the object is a string.
Null is returned when the object is not a string.
By default the string length is not limited, but can be passed as parameter

2. `public static UInt64 GetGraphSize(this ClrObject clrObject)`
This other method uses the `EnumerateReferencesWithFields` method to walk through the object graph and calculate the total size of the graph.
As an example, this is handy when you find something in the statics and need to know the amount of memory kept allocated. In general, when looking for leaks, it is a good way to evaluate the overall weight of the leak for each reference.

**Note**: The test method of `GetGraphSize` is only testing against an int[] because otherwise I don't know how to compute that same size using dotnet-dump / SOS. Any ideas on how to improve the test are welcome.

Finally, there is also a using statement reordered in HeapTests.cs

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Utilities/ClrObjectExtensions.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Utilities/ClrObjectExtensions.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+using Microsoft.Diagnostics.Runtime;
+
+namespace Microsoft.Diagnostics.Runtime.Utilities
+{
+    public static class ClrObjectExtensions
+    {
+        public static string GetStringValue(this ClrObject clrObject, int maxLength = int.MaxValue)
+        {
+            if (clrObject.Type.ElementType != ClrElementType.String) return null;
+            return clrObject.Type.ClrObjectHelpers.ReadString(clrObject.Address, maxLength);
+        }
+
+        public static UInt64 GetGraphSize(this ClrObject clrObject)
+        {
+            ulong size = 0;
+            var visited = new HashSet<UInt64>();
+            Accumulate(clrObject, visited, ref size);
+
+            static void Accumulate(ClrObject clrObject, HashSet<UInt64> visited, ref ulong size)
+            {
+                if (clrObject.IsNull) return;
+                if (visited.Contains(clrObject.Address)) return;
+
+                size += clrObject.Size;
+                visited.Add(clrObject.Address);
+                foreach (var childReference in clrObject.EnumerateReferencesWithFields())
+                {
+                    Accumulate(childReference.Object, visited, ref size);
+                }
+            }
+
+            return size;
+        }
+
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Runtime.Utilities/ClrObjectExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+using Microsoft.Diagnostics.Runtime;
+
+namespace Microsoft.Diagnostics.Runtime.Utilities
+{
+    public static class ClrObjectExtensions
+    {
+        public static string GetStringValue(this ClrObject clrObject, int maxLength = int.MaxValue)
```
leculver, line 11: Why not use `ClrObject.AsString`?  We probably need to add `ClrObject.IsString` but can't this extension method be replaced with:  `clrObject.Type.IsString ? clrObject.AsString(maxLength) : null` ?

See: https://github.com/microsoft/clrmd/blob/17bea31ca88e89ec695b2b2073eb9b61cc425e6e/src/Microsoft.Diagnostics.Runtime/src/Common/ClrObject.cs#L334

src/Microsoft.Diagnostics.Runtime.Utilities/ClrObjectExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            if (clrObject.Type.ElementType != ClrElementType.String) return null;
+            return clrObject.Type.ClrObjectHelpers.ReadString(clrObject.Address, maxLength);
+        }
+
+        public static UInt64 GetGraphSize(this ClrObject clrObject)
```
leculver, line 17: I'm planning to build something this week which will make this not needed.  Can you pause for a day or two?  I'm going to add a function to ClrObject which enumerates the entire graph pointed to by that object.  You should be able to use that with a linq expression to calculate graph size.

src/Microsoft.Diagnostics.Runtime.Utilities/ClrObjectExtensions.cs
raffaeler: You are right. I migrated my own library to ClrMD 2.0 few weeks ago and I probably missed that.
It can still be useful to avoid the check on the type. Do you want me to change the extension method name?

src/Microsoft.Diagnostics.Runtime.Utilities/ClrObjectExtensions.cs
raffaeler: Sure. I can pause any PR or just this one. I have more code in my own library that could be useful. Just let me know.

BTW in my own library I created new classes to represent a graph (the starting node and the children). This avoids the need to enumerate each ClrObject every time you are working on the graph and the need to check if it has been already visited. If you are making changes to ClrObject we can discuss this separately in order to ease this scenario as well.

raffaeler: With regards to the tests, do you have any idea on how to compare the graph size obtained through this (or a new method you are going to implement)?
The `sizeof` returned by `dotnet-dump` is slightly different for reference types only.

PR #765, opened 2020-06-16
Eliminate memory mapped file in array pool cache entry

ryanmolden:
In favor of simpler CreateFile/SetFilePointer/ReadFile calls


leculver:
Ryan and I worked on this offline.  The overall perf improvements turned out to be modest, and this would make it so we can't use the ArrayPool reader on Linux/OS X in the future so we are not merging it.

PR #776, opened 2020-06-23
Add ClrType.AssemblyLoadContextHandle

NextTurn:
Contributes to dotnet/runtime#11157


NextTurn:
@leculver Why do we name an *address* as a *handle* (`LoaderAllocatorHandle`)?

leculver:
> @leculver Why do we name an _address_ as a _handle_ (`LoaderAllocatorHandle`)?

That's just me misunderstanding what the address represents.

I'm going to go ahead and merge this now and not block on it, but if you feel there's a better term for that field please feel free to submit a PR to mark `LoaderAllocatorHandle` as Obsolete with a new field that's more descriptive.

FYI I would like to avoid as much code-rot as happened with ClrMD 1.1 by releasing sub-versions more often.  I'd like to get to a place where minor issues like this don't linger in the product for years.  Instead I plan to put together a 2.1 within the next 2-6 weeks or so with some minor breaking changes/renames that we missed.  Hopefully folks who are developing against ClrMD 2 can more easily upgrade their code if it's only a few things to fix.

PR #778, opened 2020-06-24
Rename AssemblyLoadContext{Handle -> Address}

NextTurn:
As per #776.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrType.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrType.cs
@@ -57,6 +57,9 @@ namespace Microsoft.Diagnostics.Runtime
         /// Gets the address of the <c>AssemblyLoadContext</c> object.
         /// </summary>
         public virtual ulong AssemblyLoadContextAddress => 0;
+        
+        [Obsolete("Use AssemblyLoadContextAddress.")]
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ClrType.cs
```diff
@@ -54,9 +54,9 @@ public abstract class ClrType :
         public virtual ulong LoaderAllocatorHandle => 0;
 
         /// <summary>
-        /// Gets the address of the <c>System.Runtime.Loader.AssemblyLoadContext</c> object.
+        /// Gets the address of the <c>AssemblyLoadContext</c> object.
         /// </summary>
-        public virtual ulong AssemblyLoadContextHandle => 0;
+        public virtual ulong AssemblyLoadContextAddress => 0;
```
leculver, line 59: This is a breaking change.  The original version of this PR had the old `AssemblyLoadContextHandle` marked [Obsolete], which is the right thing to do here.  Otherwise this PR will have to wait until 2.1.

```suggestion
        public virtual ulong AssemblyLoadContextAddress => 0;
        
        [Obsolete("Use AssemblyLoadContextAddress.")]
        public virtual ulong AssemblyLoadContextHandle =>AssemblyLoadContextAddress;
```

NextTurn: This API was added very lately and hasn't been published, I think.

PR #792, opened 2020-08-13
Add IThreadReader

leculver:
- Added an optional interface for IDataReaders which preserves some ClrMD 1.1 functionality.
- WindowsLiveDataTarget doesn't currently implement this.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/ThreadReaderTests.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/ThreadReaderTests.cs
@@ -1,0 +1,60 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Runtime.DacInterface;
+using Microsoft.Diagnostics.Runtime.DataReaders.Implementation;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace Microsoft.Diagnostics.Runtime.Tests
+{
+    public class ThreadReaderTests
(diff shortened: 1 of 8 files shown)


mikem8361:
Is this ready to merge?

mikem8361:
I didn’t catch this until now but are you sure you want IThreadReader in the Microsoft.Diagnostics.Runtime.DataReaders.Implementation namespace? The rest of data readers are in just Microsoft.Diagnostics.Runtime.

mikem8361:
It also looks like the EnumerateOSThreads() on the Windows minidump reader no longer puts the "Main" thread first.  

This is different kind of breaking behavior for dotnet-dump.

leculver:
@mikem8361 Do you need signed packages to test this?  I can give them to you offline if so.

> are you sure you want IThreadReader in the Microsoft.Diagnostics.Runtime.DataReaders.Implementation namespace?

I'm about 75% sure I want it in the Implementation namespace.  =)

The underlying idea here is that this is a very, very special case that really only SOS/dotnet-diagnostics is using.  Having "hidden" interfaces is wacky enough, and I want to make sure people are truly opting into them.  I also considered changing it from .Implementation to M.D.R.Optional as a way to say "hey these might not be implemented."

> It also looks like the EnumerateOSThreads() on the Windows minidump reader no longer puts the "Main" thread first.

I'll see how fixable this is.  I need to chase down some issues that came in over the weekend this morning (and mondays I have a lot of meetings) but I'll do this today between meetings.

mikem8361:
The namespace is ok with me. I just wanted to make sure it wasn't a typo. 

I understand about how busy you.  Whenever you get time to look at the thread enumeration order.

leculver:
> I understand about how busy you. Whenever you get time to look at the thread enumeration order.

This is still my top priority, I want to get this wrapped up while it's fresh.

PR #795, opened 2020-08-13
Add a fake "SizeOfImage" on Linux

leculver:
SOS requires us to be able to know the size of an image in memory.  This unblocks some functionality by letting us use the SizeOfImage on coredumps for this purpose.


leculver:
@mikem8361 This change makes it so that SOS can use ModuleInfo.SizeOfImage in the same way that it used to use ModuleInfo.Size.  The underlying values being given out should be the same here, let me know if this doesn't resolve the problem.

mikem8361:
Is this ready to merge?  I would like to test it (and the IThreadReader changes) with my diagnostics repo branch.

leculver:
Need to fix LinuxLiveDataReader, but will push a new build first thing Monday morning.

PR #801, opened 2020-08-18
Ensure threads are enumerated in the same order as dumps

leculver:
This change moves 2.0 to match ClrMD 1.1 behavior, where we ensure that threads are enumerated in the order they exist in minidumps and coredumps.

Note that this does _not_ guarantee any particular ordering, such as assume the first thread is the "main" thread.  Programs that create crash and coredumps are able to set any particular ordering in the dumps.  It's valid to create crash dumps and coredumps where the first thread in the stream is not the main thread, so you should not take any dependency on ordering.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/ThreadReaderTests.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/ThreadReaderTests.cs
@@ -42,9 +42,9 @@ namespace Microsoft.Diagnostics.Runtime.Tests
         public void EnumerateOSThreadIdsTest()
         {
             using DataTarget dt = TestTargets.Types.LoadFullDump();
-            IThreadReader threadReader = (IThreadReader)dt.DataReader;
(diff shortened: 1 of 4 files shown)


leculver:
@mikem8361 This fixes the thread ordering difference in ClrMD 2.0, but as the message above says, it's probably a bad thing that SOS or dotnet- related tools take a dependency on the ordering of threads.  It's entirely possible that the OS or program writing the crash dump could change ordering on you and break any functionality that requires the first thread to be the main thread.

mikem8361:
It is kind of a user experience and more a diagnostics test issues more than SOS itself.  Most of the other debuggers (windbg and lldb) “seem” to always have main thread as the current on dumps.

Thanks for doing this. I’ll test it with my clrmd20 branch if you want to wait to kick off the official 2.0 build.

leculver:
I'm trying to wrap up two last changes and I'll publish an official package.  ETA later today, or tomorrow afternoon depending on how long it takes.

PR #802, opened 2020-08-18
Support pinned object heap

leculver:
Having trouble coming up with a reasonable test for this since we build against desktop CLR for our test targets.


leculver:
@chrisnas FYI

PR #807, opened 2020-08-19
Add ClrArray.GetStructValue()

chrisnas:
Add GetStructValue() to ClrArray and refactor GetObjectValue() accordingly.
Should fix https://github.com/microsoft/clrmd/issues/786


src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -157,15 +157,65 @@ public int GetUpperBound(int dimension)
...
+            }
+            else
+            {
+                return Address + (ulong)valueByteOffset;
+            }
+        }
+
+        public ClrValueType GetStructValue(int index)
+        {
+            if (Type.ComponentType.IsObjectReference)
```
leculver, line 201: ComponentType needs to be null checked.  Unfortunately there's some types we just can't get information on, and an array of generic types sometimes has a null ComponentType.  I think returning an empty ClrValueType (which will has `ClrValueType.IsValid == false`) is the only sensible thing to do in that case.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -157,15 +157,65 @@ public int GetUpperBound(int dimension)
...
+
+            ulong address = GetItemAddress(index, false);
+            return new ClrValueType(address, Type.ComponentType, false);
+        }
+
+
         public ClrObject GetObjectValue(int index)
         {
-            if (!Type.IsObjectReference)
+            if (!Type.ComponentType.IsObjectReference)
```
leculver, line 211: This needs a null check, build warning.  Same as above.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -157,15 +157,65 @@ public int GetUpperBound(int dimension)
...
+            {
+                return Type.ClrObjectHelpers.DataReader.Read<UIntPtr>(Address + (ulong)valueByteOffset).ToUInt64();
+            }
+            else
+            {
+                return Address + (ulong)valueByteOffset;
+            }
+        }
+
+        public ClrValueType GetStructValue(int index)
```
NextTurn, line 199: `struct` is merely a C# concept.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -157,15 +157,65 @@ public int GetUpperBound(int dimension)
 
         public T GetValue<T>(params int[] indices) where T : unmanaged => GetValue<T>(false, indices);
 
+        // This helper method returns the address of the item at a given index of an array.
+        // - reference type object: the array item at the given index contains the address so it has to be read.
+        // - value type object: it will be embedded inside the array itself so the address is computed relatively to the array itself.
+        private ulong GetItemAddress(int index, bool isObjectReference)
```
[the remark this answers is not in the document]
NextTurn: This has been covered by `(uint)valueOffset` checks below.

src/Microsoft.Diagnostics.Runtime.Tests/src/BasicArrayTests.cs
```diff
@@ -27,6 +27,7 @@ public BasicArrayTests(ArrayConnection connection)
             _connection = connection;
         }
 
+
```
sungam3r, line 30:
```suggestion
```

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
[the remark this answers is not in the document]
chrisnas: I prefer an ArgumentOutOfRangeException to a System.OverflowException : Arithmetic operation resulted in an overflow so I'm fixing it.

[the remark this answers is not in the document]
NextTurn: The casts will not be checked.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
chrisnas: Good idea.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
chrisnas: ok

src/Microsoft.Diagnostics.Runtime.Tests/src/BasicArrayTests.cs
chrisnas: line removed

leculver:
LGTM.

PR #809, opened 2020-08-21
Add missing boolean interior parameter to ClrArray.GetStructValue() 

chrisnas:
This is needed to support array of struct: I'm not sure that false is the right default value but it was the initial implementation.


src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -199,7 +199,7 @@ private ulong GetItemAddress(int index, bool isObjectReference)
             }
         }
 
-        public ClrValueType GetStructValue(int index)
+        public ClrValueType GetStructValue(int index, bool interior = false)
```
leculver, line 202: This is technically a breaking change.  I think we'll need to structure it this way to preserve binary compat:

```suggestion
        public ClrValueType GetStructValue(int index) => GetStructValue(index, false);

        public ClrValueType GetStructValue(int index, bool interior)
```

leculver: @chrisnas This change LGTM, if you wouldn't mind committing the above change.

Thanks!

chrisnas: I hesitated between the two solutions when I wrote it  :^)
I will do the change tomorrow

leculver:
@chrisnas Shipped here:  https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/2.0.142501

PR #814, opened 2020-08-27
FIX: missing case in GetMethodByInstructionPointer() 

chrisnas:
Same implementation as in ClrMD 1.1


chrisnas:
I still have the unnecessary commit list even though the modified file is fine.
I'm changing the way to create the PR and resubmitting it ASAP.

Issue #624, opened 2020-04-23
API Design and Issues thread


NextTurn:
It seems to me that "helpers" are usually some small utilities that complement the main functions, while in ClrMD the "helpers" build everything (especially the `RuntimeBuilder` which implements 10 `I*Helpers`).

leculver:
I'm terrible at naming things.  =)

The original idea was that `Builders` actually construct Clr* types, where as `Helpers` are what various Clr* implementations use to implement its functionality.  That's not really how it's worked out in practice.

These kinds of naming issues are what I'm hoping to fix in "API Review" which I call out here:  https://github.com/microsoft/clrmd/issues/614.  Feel free to use this issue (or create new issues) for things that look weird/should be renamed in the surface area.  I want to do a full pass over the entire library on or before the date listed in 614.

(Though also I don't plan for I*Helpers and I*Builders to be a "stable" API surface.  I haven't documented that fact yet, but those are implementation details that are there in case someone wants to mock the internals of ClrMD without having to construct an entire object model.  I don't plan to support a stable version of those APIs.)

leculver:
Fix data readers: https://github.com/microsoft/clrmd/pull/630#issuecomment-619144883

leculver:
Fix inconsistent "Minidump" vs "MiniDump" and "Coredump" vs "CoreDump".

NextTurn:
Clarify that `IDataReader.GetVersionInfo` only works for native modules?

leculver:
I've mostly finished API breaking changes/fixes.  I'm trying to scope my work down to things that need to be fixed before RTM, IE breaking changes.  The I*Builder/I*Helper issue I'm leaving until a bit later since I'm not promising those interfaces will remain stable.  They are primarily meant for testing/mocking parts of the library.

Any feedback on API issues or changes within the `Microsoft.Diagnostics.Runtime` is greatly appreciated as I'm trying to ship 2.0 RTM on the 15th.

> Clarify that `IDataReader.GetVersionInfo` only works for native modules?

I'm still debating what to do about this.  Having the version information for all modules (native or managed) is super useful for tools to have.  Especially when listing out modules to the user.  Removing `ModuleInfo.Version` or making it return bogus data for managed code makes me sad. 
 File layout issues with managed code make it very difficult to do anything sensible though.

I think `IDataReader.GetVersionInfo` is the last major API issue I have planned to work on unless you or other folks raise other API concerns.  I'm not sure what I'm going to do with it yet.

[...]

PR #816, opened 2020-08-28
Tweak ClrArray

NextTurn:
- Reverts #787 - it did nothing.
- Deprecates `GetStructValue(int, bool)` - in ClrMD you simply can't get a non-interior struct value.
- Adds multi-dimensional `GetStructValue`.
- Fixes multi-dimensional `GetObjectValue`.
- Resolves code duplication.
- Reduces generic instantiation.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
@@ -153,128 +153,118 @@ namespace Microsoft.Diagnostics.Runtime
             return length + lowerBound - 1;
         }
 
-        public T GetValue<T>(int index) where T : unmanaged => GetValue<T>(false, index);
-
-        public T GetValue<T>(params int[] indices) where T : unmanaged => GetValue<T>(false, indices);
-
-        // This helper method returns the address of the item at a given index of an array.
-        // - reference type object: the array item at the given index contains the address so it has to be read.
-        // - value type object: it will be embedded inside the array itself so the address is computed relatively to the array itself.
-        private ulong GetItemAddress(int index, bool isObjectReference)
+        public unsafe T GetValue<T>(int index) where T : unmanaged
         {
-            if (Rank != 1)
-                throw new ArgumentException($"Array {Address:x} was not a one-dimensional array.  Type: {Type?.Name ?? "null"}");
+            if (sizeof(T) != Type.ComponentSize)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
```diff
@@ -153,128 +153,118 @@ public int GetUpperBound(int dimension)
...
+
+            ulong address = GetElementAddress(Type.ComponentSize, indices);
+            return new ClrValueType(address, Type.ComponentType, interior: true);
+        }
 
         public ClrObject GetObjectValue(int index)
         {
-            if (Type.ComponentType == null)
-                return new ClrObject();
+            if (Type.ComponentType is null)
```
leculver, line 211: Thinking about this more, I think we should remove this null check/return default.  We can successfully get an object value if `ComponentType` is null, we don't use it at all in the size calculation, and the ClrType used for the object comes from reading the method table with ClrHeap.GetObject.

leculver:
@NextTurn A while back someone suggested adding a NuGet package which generates a public API .txt file so that it's easy to see when someone makes a breaking change to the surface area of the library through diffs.  Do you happen to remember that at all?  I can't seem to find the issue or comment that brought it up and it seems like a good time to add it to the library.

NextTurn:
> A while back someone suggested adding a NuGet package which generates a public API .txt file so that it's easy to see when someone makes a breaking change to the surface area of the library through diffs. Do you happen to remember that at all? I can't seem to find the issue or comment that brought it up and it seems like a good time to add it to the library.

It was suggested by @drewnoakes in https://github.com/microsoft/clrmd/issues/624#issuecomment-623246855.

src/Microsoft.Diagnostics.Runtime/src/Common/ClrArray.cs
NextTurn: What about the reference/value type validation?

```cs
if (Type.ComponentType?.IsObjectReference == false) throw
```

leculver: Yeah that seems right to me.  I'll wrap that into a change and merge this one as is.

leculver:
@chrisnas Shipped here:  https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/2.0.145301

drewnoakes:
>> @NextTurn A while back someone suggested adding a NuGet package which generates a public API .txt file so that it's easy to see when someone makes a breaking change to the surface area of the library through diffs.
>
> It was suggested by @drewnoakes in #624 (comment).

Implemented in #826.

PR #820, opened 2020-09-01
Fix ILOffsetMap in Minidumps

leculver:
The dac reads a single byte of memory out of the instruction pointer you pass to EnumerateMethodInstancesByAddress to see if it's a valid location or not.  Unfortunately in a Minidump scenario it's normal for that read to fail since we simply did not put the addres into the dump file.

This change forces that single read to succeed.


leculver:
I don't like this solution because it's not thread safe.  I also don't like the fact that I'll have to take a giant lock around reads (or at least an interlocked instruction), so I'm at a loss to fully fix this properly.  I'm going to think about this problem for a bit.

PR #821, opened 2020-09-01
Allow null ComponentType for ClrArray.GetObjectValue

leculver:


NextTurn:
You might want to make the same change to `GetStructValue`.

leculver:
> You might want to make the same change to `GetStructValue`.

I'm not sure I understand.  `ComponentType` must be non-null and a representative type, because we use it to construct the underlying `ClrValueType` we return:

    return new ClrValueType(address, Type.ComponentType, interior);

So the same change in GetStructValue would result in us returning a ClrValueType with a null Type (essentially making it useless).

PR #825, opened 2020-09-03
Fix ELF reader for MacOS generated dumps and other changes

### LINKED ISSUE #3: ClrMD memory leak Thread.BlockingObjects
The code:

```
using System;
using System.Collections.Generic;
using System.Diagnostics;
using System.Globalization;
using System.Linq;
using System.Threading;
using Microsoft.Diagnostics.Runtime;

namespace ConsoleApplication1
{
    internal class MyClass
    {
        public static void Main()
        {
            var thread = new Thread(
                () =>
                {
                    while (true)
                    {
                        Console.WriteLine("123");
                    }
                });

            thread.Start();

            var pid = Process.GetCurrentProcess().Id;

            using (var dataTarget = DataTarget.AttachToProcess(pid, 5000, AttachFlag.Passive))
            {
                string dacLocation = dataTarget.ClrVersions[0].TryGetDacLocation();
                var runtime = dataTarget.CreateRuntime(dacLocation);

                while (true)
                {
                    Thread.Sleep(200);


                    foreach (var t in runtime.Threads)
                    {
                        var temp = t.BlockingObjects;
                    }

                    runtime.Flush();

                    GC.Collect();
                }
            }
        }
    }
}
```

When you run this code managed memory stays consitent, but native memory grows consistently. The detalization of most memory allocations below:
clrmd_memoryleak

In my test run I got from 2.2 MB to 10.5 MB in 8 minutes, so I guess it is easily reproducible.

Original discussion is here: https://github.com/Microsoft/dotnetsamples/issues/19

_Upd1_: in order to run code in 4.6 use this version https://github.com/Microsoft/dotnetsamples/issues/11
_Upd2_: Some of memory leak is due to: https://github.com/dotnet/coreclr/issues/1379, but probably this is not all of them.

mikem8361:
This fixes the following issues:

1) SOS for Linux dumps generated on MacOS. The ModuleInfo ImageSize wasn't being set in this case. The ElfLoadedImage.Size is set to an appropriate value even for MachO in-memory images but when ElfLoadedImage.Open fails and PEImage.ReadIndexProperties is called the file size variable is set to 0. Rearranged the code to preserve the Size.

2) Made PEImage.ReadPdbs() more bullet proof and changed it not to return "default" which causes PEImage.DefaultPdb to hit an null ref. The try/catch added isn't as neccesary anymore because of fix #3.

3) Fixed module PDBInfo and Version for PE images in ELF coredumps. When building the module infos changed the CoreDumpReader's to always pass isVirtual to true. The _containsExecutable flag is the right value. It meant that the ELF image was an executable not the managed PE assembly. The only problem is that are still some PEs that the layout (isVirtual) is wrong and PDBInfo/Version doesn't work. This happen to be the app's DLLs. All the system BCL assemblies work.

4) Removes " (deleted)" from the ELF image path for coredumps with unloaded modules. Fixes issue https://github.com/dotnet/diagnostics/issues/1268 for dotnet-dump even though the issue was filed against dotnet-symbol. They both have the same problem.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Core/CoreDumpReader.cs
@@ -92,15 +92,19 @@ namespace Microsoft.Diagnostics.Runtime
                 using Stream stream = image.CreateStream();
                 PEImage.ReadIndexProperties(stream, out timestamp, out filesize);
             }
-            else
+
+            // It's true that we are setting "IndexFileSize" to be the raw size on linux for Linux modules,
+            // but this unblocks some SOS scenarios.
+            if (filesize == 0)
             {
-                // It's true that we are setting "IndexFileSize" to be the raw size on linux for Linux modules,
-                // but this unblocks some SOS scenarios.
-                filesize = (int)image.Size;
+                filesize = unchecked((int)image.Size);
             }
 
+            // This substitution is for unloaded modules for which Linux appends " (deleted)" to the module name.
+            string path = image.Path.Replace(" (deleted)", "");
+
             // We set buildId to "default" which means we will later lazily evaluate the buildId on demand.
-            return new ModuleInfo(this, (ulong)image.BaseAddress, image.Path, image._containsExecutable, filesize, timestamp, buildId: default);
+            return new ModuleInfo(this, (ulong)image.BaseAddress, path, true, filesize, timestamp, buildId: default);
         }
 
         public void FlushCachedData()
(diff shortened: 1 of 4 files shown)


leculver (approved):
The `_containsExecutable` change is worrisome a bit.  Your reasoning here is sound but I worry there's some other odd corner case we aren't covering.  I assume you all will likely find that faster than I would though as our test coverage isn't great there (and my one-off usage rarely ventures to Linux).

We can turn around a fix pretty quickly so I'm going to go ahead and ship this.

leculver:
@mikem8361 Shipped here:  https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/2.0.145301

mikem8361:
Thanks for the fast turnaround.

mikem8361:
Yes, we need to discuss and deal with the filelayout/isVirtual for the PE images in the module list.  For getting the version in CoreDumpReader’s GetVersionInfo(), the only way I can come up with is to try both isVirtual = true and if that fails then isVirtual = false. I did that in the 1.1 CoreDumpReader code.  This would also have to be done  ModuleInfo’s PDBInfo and IsManaged properties. This gets a little messy and makes the ModuleInfo’s _isVirtual field obsolete/incorrect.

And the LinuxLiveDataReader needs all of my fixes and these two.

When can discuss this when we get some time.

PR #826, opened 2020-09-05
Update analyzers, and add public API analyzer

drewnoakes:
This adds `Roslyn.Diagnostics.Analyzers`, which includes the public API analyzers and addresses all new warnings.

Also update `Microsoft.CodeAnalysis.FxCopAnalyzers` to 3.3.0 and address new warnings (in commits after 59136c65164895354d174d37944ee318da5e01c2).

One diagnostic per commit to make it easier to review/amend.


leculver:
@drewnoakes This is awesome, thank you for doing the work here!

One quick question here.  What's the difference between "PublicAPI.Shipped.txt" and "PublicAPI.Unshipped.txt"?  I read what sparse documentation I could find, but I didn't really understand the split between these two files.  In my mind, the entire API surface at this point is "shipped" but maybe these mean something else?

src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
```diff
@@ -1570,7 +1568,11 @@ private static bool IsArraySubstring(string name, int start, int end)
         [return: NotNullIfNotNull("name")]
         public static string? FixGenerics(string? name)
         {
-            if (name == null || name.IndexOf("[[") == -1)
+#if NETCOREAPP3_1
```
leculver, line 1571: Instead of putting #if statements in the code, we have been solving this via adding extension methods on platforms where functions are missing.  For example:  https://github.com/microsoft/clrmd/blob/master/src/Microsoft.Diagnostics.Runtime/src/Extensions/SpanExtensions.cs

Would you mind changing this to just be `name.Contains` for both cases, and add an extension method for non-NETCOREAPP3_1?

src/Microsoft.Diagnostics.Runtime/src/Common/GCRoot.cs
```diff
@@ -174,7 +174,7 @@ public IEnumerable<GCRootPath> EnumerateGCRoots(ulong target, bool returnOnlyFul
             CancellationToken cancellationToken)
         {
             IClrRoot? root;
-            while ((root = queue.Take()) != null)
+            while ((root = queue.Take(cancellationToken)) != null)
```
leculver, line 177: This will throw an OperationCancelledException, right?  This will be an unhandled exception on a worker thread which will tear down the process.  If you are adding `cancellationToken` to Add and Take you also need to wrap these with a catch/return.

NextTurn:
Why use `Roslyn.Diagnostics.Analyzers` here?

> This package contains rules that are very specific to the .NET Compiler Platform ("Roslyn") project, i.e. [dotnet/roslyn](https://github.com/dotnet/roslyn) repo. This analyzer package is _not intended for general consumption_ outside the Roslyn repo. [More info about rules in this package](src/Roslyn.Diagnostics.Analyzers/Roslyn.Diagnostics.Analyzers.md)

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
```diff
@@ -96,6 +96,9 @@ internal override unsafe bool GetFileVersion(string dll, out int major, out int
             long address = dataOffset;
             long endAddress = address + dataSize;
 
+            Span<byte> bytes = stackalloc byte[1];
+            Span<char> chars = stackalloc char[1];
```
leculver, line 100: I'm curious about this change, what is it for?  I don't care if this is just a cosmetic thing you added, but if this is due to a new build warning I'd like understand what it's warning about.

src/Microsoft.Diagnostics.Runtime/src/Windows/Minidump.cs
```diff
@@ -198,7 +198,11 @@ private async Task<ThreadReadResult> ReadThreadData(Stream stream)
                             continue;
 
                         int count = ResizeBytesForArray<MinidumpThread>(numThreads, ref buffer);
+#if NETCOREAPP3_1
```
leculver, line 201: What does converting buffer to Memory<T> buy us here?

I am trying to avoid littering the code with `#if [version]` #defines.  Is there a way to make this into an extension method of some sort?  I'm not sure what that would look like.  (It would fit here: https://github.com/microsoft/clrmd/blob/master/src/Microsoft.Diagnostics.Runtime/src/Extensions/SpanExtensions.cs#L20).

If there's a clear pattern to doing this it'd really need to be spread throughout the codebase, and I'd rather not duplicate this #if pattern everywhere.  (Maybe these are the only places, but I'm _never_ going to remember `AsMemory` and splitting the codepath when reading from streams.)

leculver (changes requested):
Overall this looks good.  I have a few minor things I'd like to see if there's a way to avoid `#if [version]` throughout the code as well as some minor comments.

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
NextTurn: Just like in C++, the memory allocated on the stack isn't freed after each iteration.

NextTurn: Just like in C++, stack space allocated is only released at the end of the method.

NextTurn:
> What's the difference between "PublicAPI.Shipped.txt" and "PublicAPI.Unshipped.txt"?

As I understand, "unshipped" APIs refer to those which are present in the code base but haven't been released in NuGet packages.

leculver:
> As I understand, "unshipped" APIs refer to those which are present in the code base but haven't been released in NuGet packages.

I guess I don't understand the mechanism.  Do I manually move APIs from unshipped to shipped by cutting/pasting the text files?

drewnoakes:
> What's the difference between "PublicAPI.Shipped.txt" and "PublicAPI.Unshipped.txt"?

The analyzer treats them both the same way, so whether an API is in one or the other is only important if you want it to be.

The distinction probably makes sense when you have long periods between shipping your API, in which time the unshipped file grows. When you ship you can document the API changes, then push everything into shipped and start fresh. It's my opinion that for many use cases this distinction isn't helpful, so I ignore it. The analyzer comes with a codefix which allows automatically updating the unshipped file, which is very convenient as editing the files manually is hard to get right. At some point I'd like to contribute a change to the analyzer that allows using only one file (eg `PublicAPI.txt`).

drewnoakes:
> Why use `Roslyn.Diagnostics.Analyzers` here?
> 
>> This package contains rules that are very specific to the .NET Compiler Platform ("Roslyn") project, i.e. [dotnet/roslyn](https://github.com/dotnet/roslyn) repo. This analyzer package is _not intended for general consumption_ outside the Roslyn repo. [More info about rules in this package](src/Roslyn.Diagnostics.Analyzers/Roslyn.Diagnostics.Analyzers.md)

@NextTurn these analyzers are generally useful and used in many places outside of Roslyn.

`Roslyn.Diagnostics.Analyzers` brings in several other packages, and it's easier to just include `Roslyn.Diagnostics.Analyzers` directly so that the versions of those other packages are compatible.

Changes in response to `Roslyn.Diagnostics.Analyzers`:

- RS0038 82f71031b6ea1f9dd593322310003d465ef9c89a
- RS0042 e697de74c57ca0b9ae93cf931541cd61e2147c5a
- RS0100 d9db26bfbbac020a64b9299ce67bfa4afcdfc7d7
- RS0101 dadabd97e84722fe2506d540c9caebdbe025f6da

If there are any diagnostics that don't fit the project, it's easy to exclude them.

src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/LinuxFunctions.cs
drewnoakes: This is not cosmetic. By using `stackalloc` in a loop, the stack pointer is incremented each time the loop iterates. So we allocate N bytes on the stack instead of one, where only one is needed.

src/Microsoft.Diagnostics.Runtime/src/Windows/Minidump.cs
drewnoakes: An extension method makes sense. Agree that `#if` is ugly and good to remove.

> What does converting buffer to Memory buy us here?

I will look into whether this is actually a useful change for those frameworks. If not we can disable the diagnostic or suppress these occurrences.

> I'm never going to remember AsMemory and splitting the codepath when reading from streams.

The good news is you don't have to. The analyzer will nag you with a warning. All these changes are a result of analyzer warnings.

If it isn't already, I'd consider making CI builds fail on warnings.

drewnoakes: The rule's description suggests there is a perf improvement to using the memory overload. From https://docs.microsoft.com/en-gb/visualstudio/code-quality/ca1835?view=vs-2019:

> The memory-based method overloads have a more efficient memory usage than the byte array-based ones.

Adding an extension method addresses the `string.Contains`/`IndexOf` comment above. It's not clear to me how to avoid the `#if` here with an extension method.

So it's your call as to whether you wish to keep the `#if`, suppress the warnings and keep the old code, or disable the rule altogether.

src/Microsoft.Diagnostics.Runtime/src/Common/GCRoot.cs
drewnoakes: Good catch. In 0ccf44b37c7115c754bc1370a6949a261ac1246e I've removed the token from `Add` calls as it's unbounded, so won't block. To address the resulting warning, I've added a localised suppression using `#pragma warning disable`, as other suppression styles will suppress the diagnostic in the entire method.

Please also review:
- a8f8cbb9a549e0d935e3d344dac5b200e5c67316 which adds `BlockingCollection<>.CompleteAdding`
- 4e135291cbd3f7f4ed6f29bfbdd29ea09cfb683a which returns earlier in case of cancellation, which may also yield fewer results

src/Microsoft.Diagnostics.Runtime/src/Windows/Minidump.cs
leculver: Thanks for checking!  Let's leave your change as-is since there's a real perf benefit here.  I'll try to figure out whether I want to refactor the codepaths better in the future.

leculver (approved):
Thanks for doing this!  LGTM.

sungam3r:
[PublicApiGenerator](https://github.com/PublicApiGenerator/PublicApiGenerator) also may be helpful.

drewnoakes:
@sungam3r thanks for the pointer. That tool looks identical in functionality to the analyzer. Am I missing something?

sungam3r:
They solve approximately the same problem, but in different ways.

PR #836, opened 2020-09-25
Fix the architecture bit in the arm64/x86 context flags

mikem8361:
This is important because if the context is returned back to DBI/DAC like it is in dotnet-dump, the context isn't internally copied correctly causing the SOS `clrstack -i` commands to fail.  

DAC/DBI code does this:

  if ((dstFlags & srcFlags & DT_CONTEXT_CONTROL) == DT_CONTEXT_CONTROL)


leculver:
LGTM.  I have some Arcade build issues to work through before I can ship and I'd like to fix a couple of outstanding issues.  I _should_ be able to get a new build out today assuming I can get the build system back into good shape.

PR #845, opened 2020-10-13
Fix public api files for the recently changed context flags

mikem8361:
Fix DacDataTargetWrapper.ReadVirtual bug that caused my "clrmodules" command to loop forever.


src/Microsoft.Diagnostics.Runtime/src/DacInterface/DacDataTargetWrapper.cs
```diff
@@ -154,6 +154,7 @@ public HResult ReadVirtual(IntPtr _, ClrDataAddress cda, IntPtr buffer, int byte
                         return HResult.E_NOTIMPL;
                     }
 
+                    Trace.TraceInformation($"ReadVirtual {address:X16} size {bytesRequested}");
```
leculver, line 157: I probably don't want to put this in shipping code.  It would be very noisy when debugging as there's a lot of ReadVirtual calls.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -15,9 +15,6 @@
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-#if NET5_0
```
leculver, line 18: It's fine to strip this.  The real reason I left it in is because it's _supposed_ to work, and would work if I do the versioning fix and pick up a new copy of arcade.  I planned to do that work and had other things going on...

leculver (changes requested):
Can you pull out the Trace statement?  Otherwise looks good to me.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/DacDataTargetWrapper.cs
mikem8361: Done. That was supposed to be temporary.

leculver:
@mikem8361 Should be fixed here:  https://www.nuget.org/packages/Microsoft.Diagnostics.Runtime/2.0.151903

PR #854, opened 2020-10-26
Fix triage dumps

mikem8361:
As it turns out Linux triage dumps on Windows and Linux are missing some memory that causes getting these basic types to fail so getting them needs to be delayed until they are actually used.  This is true in 3.1 and 5.0. For 6.0, we can fix the DAC enum regions to include these types, but clrmd/dotnet-dump needs to support them. System.String was the type failing (Free and System.Object succeeded) but to be consist I made all 4 types lazily initialized.

Not sure how you feel about using Lazy<T> here because I don't see it being used any where else but some tests.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdHeap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdHeap.cs
@@ -16,6 +16,10 @@ namespace Microsoft.Diagnostics.Runtime.Implementation
     {
         private const int MaxGen2ObjectSize = 85000;
         private readonly IHeapHelpers _helpers;
+        private readonly Lazy<ClrType> _freeType;
+        private readonly Lazy<ClrType> _objectType;
+        private readonly Lazy<ClrType> _stringType;
+        private readonly Lazy<ClrType> _exceptionType;
 
         private readonly object _sync = new();
         private volatile VolatileHeapData? _volatileHeapData;
@@ -36,13 +40,13 @@ namespace Microsoft.Diagnostics.Runtime.Implementation
 
         public override int LogicalHeapCount { get; }
 
-        public override ClrType FreeType { get; }
+        public override ClrType FreeType => _freeType.Value;
 
-        public override ClrType StringType { get; }
+        public override ClrType StringType => _stringType.Value;
 
-        public override ClrType ObjectType { get; }
+        public override ClrType ObjectType => _objectType.Value;
 
-        public override ClrType ExceptionType { get; }
+        public override ClrType ExceptionType => _exceptionType.Value;
 
         public override bool IsServer { get; }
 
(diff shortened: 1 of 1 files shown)


leculver:
> As it turns out Linux triage dumps on Windows and Linux are missing some memory that causes getting these basic types to fail so getting them needs to be delayed until they are actually used. 

That's really bad.  Does this affect SOS commands and !analyze, or just ClrMD because we request more data?

> Not sure how you feel about using Lazy here because I don't see it being used any where else but some tests.

Generally speaking I avoid it.  ClrMD 1.1 used a lot of Lazy<T> and the extra delegates laying around ends up eating a non-trivial amount of heap space.  It's also really annoying to debug.  Pre-loading these types was intentional.  I need to take a look at what's going on to see if this is the right fix

mikem8361:
It is that ClrMD is requesting more info on initialization.  For triage dumps, from SOS point of view, only stacks (clrstack), exceptions (pe) and managed threads (clrthreads) need to work.  Dumping the heap with types/stats (dumpheap) is supported.  I assume this would also be true for clrmd, too.  I assuming unless you are enumerating the GC heap with clrmd, those pre-initialized type wouldn’t be needed.  I hit this in my extensions side-project branch where I was attempting to use ClrRuntime to get the ClrDataProcess instance for the native SOS.  In the current dotnet-dump I hit it with the “clrmodules” command which uses clrmd to enumerate the managed assemblies.  Not sure if some of the newly add Criteo commands would work.

leculver:
Mike and I chatted offline, but I wanted to capture all of that context here:

I need to think about this for a bit.  Fundamentally this runs against the design of ClrMD 2.0:
 
	We pre-fetch the object types that we expect to always work.
	Things which use ClrMD expect that if you can create a ClrHeap, then basic operations should behave "normally" and not fail/throw.
	The rest of the system assumes that if ClrHeap is created then we have a valid System.String type
	We can never have a null value for System.String because it would break a lot of other invariants.  (This change doesn't do that, but it's also something that won't work.)

This PR won't work due to 2 above.  Someone consuming ClrMD would write this really reasonable code in which msg should be null but instead they get an exception when they lazily try to get the String type:
 
        ClrRuntime runtime = dt.ClrVersions.Single().CreateRuntime();
        ClrHeap heap = runtime.Heap;
        ClrException ex = runtime.Threads.First().CurrentException;
        string msg = ex.Message;

Fundamentally this is my oversight in how I designed ClrHeap, assuming System.String the basic types would always exist in the dump.  It's hard to change that assumption now, so I'll have to put some thought into how to fix this cleanly.  I can likely special case triage dumps in some way.  Ugly, but it will work.

mikem8361:
I'm closing this PR for now.

leculver:
Sounds good.  This is on my short-list to fix, but I haven't gotten around to ClrMD work recently.  Let me know if this is high priority.  It might be a week or two before I get to it otherwise.

mikem8361:
Not a high priority.

Issue #860, opened 2020-11-04
Fix triage dumps


leculver:
https://github.com/microsoft/clrmd/pull/854

PR #861, opened 2020-11-05
Fix triage dumps

leculver:
The data for System.String is not present in triage dumps, but ClrMD assumes we can always construct some basic types including System.String.  This change creates a "fake" string type when we cannot create the type for String specifically.

Fixes https://github.com/microsoft/clrmd/issues/860.


leculver:
@mikem8361 This change fixes the triage dump/string issue.

mikem8361 (approved):
Thanks.

Issue #855, opened 2020-10-27
Windows Minidump data reader doesn't return process id 


mikem8361:
Always returns -1.

PR #863, opened 2020-11-05
Implement MinidumpReader.ProcessId

leculver:
Fixes https://github.com/microsoft/clrmd/issues/855.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Windows/Minidump.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Windows/Minidump.cs
@@ -19,9 +19,7 @@ namespace Microsoft.Diagnostics.Runtime.Windows
     internal sealed class Minidump : IDisposable
     {
         private readonly string _displayName;
-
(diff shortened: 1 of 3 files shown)


leculver:
@mikem8361 This fixes Minidump process ids.

src/Microsoft.Diagnostics.Runtime/src/Windows/Minidump.cs
```diff
@@ -290,6 +284,15 @@ private async Task<ThreadReadResult> ReadThreadData(Stream stream)
                 Threads = threadBuilder.ToImmutable()
             };
         }
+
+        private static async Task<int> ReadAsync(Stream stream, byte[] buffer, int count)
+        {
+#if NETCOREAPP3_1 || NET5_0
```
NextTurn, line 290:
```suggestion
#if NETCOREAPP
```

https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#preprocessor-symbols

PR #865, opened 2020-11-10
Reimplement DAC name parsing

ryanmolden:
Reimplement DAC name parsing to handle nested classes of generic classes and a bunch of other corner cases (+ tests for same)

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Common/DACNameParser.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Common/DACNameParser.cs
@@ -1,0 +1,831 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.Diagnostics.Runtime.Builders
+{
+    // This class handles stateless, non-recursive parsing of DAC names, converting things like this:
+    //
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Runtime/src/Common/DACNameParser.cs
```diff
@@ -0,0 +1,831 @@
...
+                // that parsing level
+                if (candidateSegment.ParsingArgDepth < parsingGenericArgListDepth)
+                    break;
+
+                // Don't fold any that still have unfulfilled generics, leave them so they can be fulfilled
+                //
+                // TODO: Is this a thing? The need to check for HasUnfulfilledGenericArgs? The code before this that propagates generic args only fixes a single generic type target
+                // per pass...can there ever be more than one? The closing of a generic param list triggers this code, so the only way for there to be more than one generic type needing
+                // patching is EITHER a nested generic type in a generic parent or a generic param to a generic type, but the latter will trigger two seperate levels of generic arg parsing,
+                // the former is handled in the back-propagation code before this, and this collapses the nested generic (that has been resolved) into the parent generic (that may not
```
ryanmolden, line 512: Whoops, never took this out (the TODO) :P I think I landed on this was necessary (hence I left it in), at worst it isn't actively harmful. Can probably remove the TODO at the very least.

PR #873, opened 2020-12-04
Replace delegates with function pointers

nxtn:
Just FYI. The major flaw of this change is that Linux x86 is abandoned again, because the "platform default" calling convention is supported only on .NET 5 and above, and there is no simple way to specify calling conventions per platform.


leculver:
Is it only x86 Linux that's affected?  Do arm32 and arm64 work as well as x64?

NextTurn:
Yeah. Calling conventions are consistent on those platforms.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/MetadataImport.cs
```diff
@@ -171,15 +159,16 @@ private void CloseEnum(IntPtr handle)
...
             IntPtr handle = IntPtr.Zero;
             int[] tokens = ArrayPool<int>.Shared.Rent(32);
             try
             {
-                while (_enumGenericParams(Self, ref handle, token, tokens, tokens.Length, out int count) && count > 0)
+                while (Enum(ref handle, token, tokens, tokens.Length, out int count) && count > 0)
                     for (int i = 0; i < count; i++)
                         yield return tokens[i];
+
+                bool Enum(ref IntPtr handle, int token, int[] tokens, int length, out int count) =>
```
leculver, line 170: Do we need this?  I can't figure out why this doesn't just call `VTable.EnumGenericParams` directly

src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataMethod.cs
```diff
@@ -24,23 +22,19 @@ public ClrDataMethod(DacLibrary library, IntPtr pUnk)
 
         public ILToNativeMap[]? GetILToNativeMap()
         {
-            InitDelegate(ref _getILAddressMap, VTable.GetILAddressMap);
```
leculver, line 27: Is InitDelegate still used anywhere?  We should probably remove the original function in `CallableCOMWrapper.cs` if it's now dead code.

leculver (commented):
I have a couple of questions but overall I think this is a great improvement.  What's left for it to be finished and not a draft?

src/Microsoft.Diagnostics.Runtime/src/DacInterface/MetadataImport.cs
NextTurn: > CS1629 Unsafe code may not appear in iterators

NextTurn:
> What's left for it to be finished

Special handling of strings (which should be marshaled as char pointers).

src/Microsoft.Diagnostics.Runtime/src/DbgEng/DebugClient.cs
```diff
@@ -80,21 +64,21 @@ public HResult OpenDumpFile(string dumpFile)
...
-        public readonly IntPtr AttachProcess;
+        public readonly delegate* unmanaged[Stdcall]<IntPtr, ulong, uint, DebugAttach, HResult> AttachProcess;
         public readonly IntPtr CreateProcess;
         public readonly IntPtr CreateProcessAndAttach;
         public readonly IntPtr GetProcessOptions;
         public readonly IntPtr AddProcessOptions;
         public readonly IntPtr RemoveProcessOptions;
         public readonly IntPtr SetProcessOptions;
-        public readonly IntPtr OpenDumpFile;
+        public readonly delegate* unmanaged[Stdcall]<IntPtr, string, HResult> OpenDumpFile;
```
NextTurn, line 74: > The Unicode version of this method is OpenDumpFileWide.

leculver (approved):
Thanks for doing this!  I read through it and nothing jumped out at me.  With such a big change I think it's best to merge and do some testing, so that's what I'll do.

PR #882, opened 2021-01-12
Fix build warning

leculver:


leculver:
Using this checkin to test the build system.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Registers/VectorRegisterArea.cs
```diff
@@ -13,13 +14,19 @@ public struct VectorRegisterArea
...
         [MarshalAs(UnmanagedType.ByValArray, SizeConst = VectorRegisterSize)]
-        public M128A[] VectorRegister;
+        public M128A[]? VectorRegister;
 
         [FieldOffset(0x1a0)]
         public ulong VectorControl;
 
         public VectorRegisterArea(VectorRegisterArea other) : this()
         {
+            if (other.VectorRegister is null)
```
nxtn, line 24: This condition is unnecessary. It must be null here.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Registers/VectorRegisterArea.cs
```diff
@@ -13,13 +14,19 @@ public struct VectorRegisterArea
...
+                throw new ArgumentException("Cannot have empty VectorRegister.", nameof(other));
+
+            if (VectorRegister == null)
```
nxtn, line 27: This condition is unnecessary. It must be null here.

PR #889, opened 2021-02-22
* Handle arithmetic overflow when dealing with large memory dump files.

Kompiler:
When opening a large memory dump file (10+ GB), any segment that is larger than 2GB in size is missing.
This is because an arithmetic overflow occurs when both loading the dump and again when creating the runtime.

Loading the same dump file in ClrMd 1.1 works fine so I suspect this is a regression.

Having the segments missing means that walking the heap objects and enumerating the roots will not return the correct data.


leculver (approved):
LGTM

PR #890, opened 2021-02-23
Cleanup delegates

leculver:
- Cleaned up the delegate conversion pull request.
- Fixed compiler errors.
- Moved from passing arrays to passing pointers.


nxtn:
Passing arrays is OK. Arrays (except for `char[]`) are marshaled by reference as pointers by default.

leculver:
Almost every test was failing with arrays when I ran them locally.  I didn't dig too deeply into the root cause, but swapping to a fixed pointer seemed to resolve the issue so I made it consistent throughout.

Issue #881, opened 2020-12-24
ClrMD 2.0 reports less number of pinned handles than ClrMD 1.1


crui3er:
Hello,
I am investigating process dump with ClrMD and noticed that after upgrading from version 1.1.1 to 2.0.3 it reports less pinned handles. I prepared 2 similar programs (for both version) that collect and report to file information about pinned handles.
Here is extract from reports that show difference:

**v.2.0.3**
Total pinned: 599
Total overall: 78725
System.Byte[] -> 4
System.Object[] -> 28
System.Threading.OverlappedData -> 445

**v.1.1.1**
Total pinned: 1042
Total overall: 79168
System.Byte[] -> 326
System.Object[] -> 149
System.Threading.OverlappedData -> 445

Can you explain where this difference come from?

Also what is interesting is that when I analyze my dump file with JetBrains dotMemory tool it reports 1283 pinned objects
![image](https://user-images.githubusercontent.com/20703715/103086270-00142380-4606-11eb-8864-126962020212.png),
according to it there are 566 pinned byte arrays on a heap. Which is even more than I get with ClrMD 1.1.1.

Here is an example of program that I uses to report pinned handles information with ClrMD 2.0.3

```
private static string HandleToString(ClrHandle handle)
{
	return string.Format("{0} @{1:x12} -> {2}", (object)handle.HandleKind, handle.Address, HandleObjectToString(handle.Object));
}

private static string HandleObjectToString(ClrObject obj)
{
	ClrType type = obj.Type;
	return string.Format("{0} {1:x} {2}", type?.Name, obj.Address, obj.Size);
}

private static void ReportPinnedHandles(DataTarget dataTarget)
{
	var runtime = dataTarget.ClrVersions[0].CreateRuntime();
	
	var fileName = @"C:\Logs\PinnedHandlesReport_new_2_0_3.txt";
	if (File.Exists(fileName))
		File.Delete(fileName);

	using(var stream = new FileStream(fileName, FileMode.CreateNew))
	using (var writer = new StreamWriter(stream)) {
		var totalPinned = 0;
		var totalOverall = 0;
		var types = new Dictionary<ClrType, ulong>();
		foreach (var handle in runtime.EnumerateHandles()) {
			totalOverall++;
			if (!handle.IsPinned)
				continue;
			totalPinned++;
			var type = handle.Object.Type;
			if (!types.TryGetValue(type, out var currentNumber))
				types.Add(type, 1);
			else
[...]

PR #896, opened 2021-02-25
Enumerate inner async pinned handles

leculver:
Bring async pinned handles into line with ClrMD 1.1 behavior.

Contributes to https://github.com/microsoft/clrmd/issues/881.


leculver:
There's still a discrepancy in the number of stack roots that ClrMD 2.0 reports.  I will take a look at that before merging this change.

leculver:
I still have more work to do on #881 but for now I'm going to merge this change and publish it in the next release.

PR #901, opened 2021-03-12
Fix a couple of SOS issues

mikem8361:
Fix cases where windbg/dbgeng returns the coreclr module not in lower case (i.e. CoreCLR.dll).

Fixed Apple Silicon/Issue https://github.com/microsoft/clrmd/issues/888

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfoProvider.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfoProvider.cs
@@ -13,9 +13,9 @@ namespace Microsoft.Diagnostics.Runtime
     /// </summary>
     public static class ClrInfoProvider
     {
-        private const string c_desktopModuleName1 = "CLR.DLL";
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfoProvider.cs
```diff
@@ -45,7 +45,7 @@ public static bool IsSupportedRuntime(ModuleInfo moduleInfo, out ClrFlavor flavo
             if (moduleName is null)
                 return false;
 
-            switch (moduleName)
+            switch (moduleName.ToUpperInvariant())
```
leculver, line 48: This is an ok fix.  Instead of switching on module name should we be converting this to if statements using `moduleName.Equals(..., OrdinalIngoreCase)`?  Or will that not matter here?

mikem8361: Yes, I like that better.

leculver (approved):
LGTM

PR #906, opened 2021-03-23
 Add single-file support for Linux

mikem8361:
Add symbol export support to ELF reader. Add related structs and enums.

Clean up ELF program header creation

Add IExportReader interface

Update public API definitions

Add public RuntimeInfo/SingleFileRuntimeInfo property to ClrInfo

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
@@ -164,6 +164,7 @@ namespace Microsoft.Diagnostics.Runtime
 
             IExportReader? exportReader = DataReader as IExportReader;
             Architecture arch = DataReader.Architecture;
+            RuntimeInfo? singleFileRuntimeInfo = null;
             ImmutableArray<ClrInfo>.Builder versions = ImmutableArray.CreateBuilder<ClrInfo>(2);
             foreach (ModuleInfo module in EnumerateModules())
             {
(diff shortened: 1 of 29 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -161,12 +162,55 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
...
+
+                if (ClrInfoProvider.IsSupportedRuntime(module, out ClrFlavor flavor, out OSPlatform platform))
+                {
+                    runtimeTimeStamp = module.IndexTimeStamp;
+                    runtimeFileSize = module.IndexFileSize;
+                    runtimeBuildId = module.BuildId;
+                }
+                else
+                {
+                    if (exportReader is null || !exportReader.TryGetSymbolAddress(module.ImageBase, RuntimeInfo.SymbolValue, out ulong runtimeInfoAddress))
```
leculver, line 183: Is there any way to further narrow what modules we need to check for single module export symbols?  It's going to be painful and slow if we have to check every module for a particular symbol when single file is going to be present in very rare circumstances.

leculver: Is it always the main module that has the CLR export?  IE the executable image?

mikem8361: I know this is a perf problem but I haven't come up with any way around enumerating all the modules.  The single-file executable header isn't marked as executable (ET_EXEC); it is a shared module (ET_DYN). I have no idea why the host team did that. The normal hosts (i.e. "dotnet", "apphost", etc.) are also marked this way. It doesn't bother Linux.

PR #908, opened 2021-04-05
More single-file fixes

mikem8361:
Add a ICLRRuntimeLocator impl to DAC data target so single-file runtime can be found.

Fix single-file module names. The DAC GetPEFileName() GetAssemblyName() don't return a module/assembly file path for single file apps.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataModule.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataModule.cs
@@ -48,6 +48,17 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             return hr ? name : null;
         }
 
+        public string? GetFileName()
+        {
+            // GetFileName will fault if buffer pointer is null. Use fixed size buffer.
+            char[] buffer = new char[1024];
+            fixed (char* bufferPtr = buffer)
+            {
+                HResult hr = VTable.GetFileName(Self, 1024, out int nameLength, bufferPtr);
+                return hr && nameLength > 0 ? new string(buffer, 0, nameLength - 1) : null;
+            }
+        }
+
         [StructLayout(LayoutKind.Sequential)]
         private readonly unsafe struct IClrDataModuleVTable
         {
@@ -78,7 +89,7 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             private readonly IntPtr EnumDataByName;
             private readonly IntPtr EndEnumDataByName;
             public readonly delegate* unmanaged[Stdcall]<IntPtr, int, out int, char*, HResult> GetName;
-            private readonly IntPtr GetFileName;
+            public readonly delegate* unmanaged[Stdcall]<IntPtr, int, out int, char*, HResult> GetFileName;
             private readonly IntPtr GetFlags;
             private readonly IntPtr IsSameObject;
             private readonly IntPtr StartEnumExtents;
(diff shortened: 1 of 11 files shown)


mikem8361:
@leculver is there anyway you could take a look this week?  Thanks!

Also, the "ArityTest" test is failing on TypeTests.cs line 183. I really don't think my changes caused this.

actual:
```
Microsoft.Diagnostics.Runtime.Tests.TypeTests+GenericTest[[System.String, System.Private.CoreLib],[System.Collections.Generic.List[[System.Collections.Generic.IEnumerable[[System.Int32, System.Private.CoreLib]][,], System.Private.CoreLib]][], System.Private.CoreLib]]
```
expected:
```
Microsoft.Diagnostics.Runtime.Tests.TypeTests+GenericTest<System.String, System.Collections.Generic.List<System.Collections.Generic.IEnumerable<System.Int32>[,]>[]>
```

src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataModule.cs
```diff
@@ -48,6 +48,17 @@ public HResult GetModuleData(out ExtendedModuleData data)
             return hr ? name : null;
         }
 
+        public string? GetFileName()
+        {
+            // GetFileName will fault if buffer pointer is null. Use fixed size buffer.
+            char[] buffer = new char[1024];
```
leculver, line 54: Use ArrayPool rent and return with a try/finally here.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/ClrDataModule.cs
```diff
@@ -48,6 +48,17 @@ public HResult GetModuleData(out ExtendedModuleData data)
...
+            fixed (char* bufferPtr = buffer)
+            {
+                HResult hr = VTable.GetFileName(Self, 1024, out int nameLength, bufferPtr);
```
leculver, line 57: Pass buffer.Length instead of hardcoded 1024.

src/Microsoft.Diagnostics.Runtime/src/DacInterface/DacDataTargetWrapper.cs
```diff
@@ -51,6 +54,10 @@ public DacDataTargetWrapper(DataTarget dataTarget)
             builder = AddInterface(IID_IMetadataLocator, false);
             builder.AddMethod(new GetMetadataDelegate(GetMetadata));
             builder.Complete();
+
+            builder = AddInterface(IID_ICLRRuntimeLocator, false);
```
leculver, line 58: I would wrap this with `if (runtimeBaseAddress > 0) { ... }`.  IE don't expose the interface if we don't have a runtimeBaseAddress.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DacLibrary.cs
```diff
@@ -86,7 +86,12 @@ public DacLibrary(DataTarget dataTarget, IntPtr pClrDataProcess)
             DacDataTarget = new DacDataTargetWrapper(dataTarget);
         }
 
-        public unsafe DacLibrary(DataTarget dataTarget, string dacPath)
+        public DacLibrary(DataTarget dataTarget, string dacPath)
+            : this(dataTarget, dacPath, runtimeBaseAddress: 0)
+        {
+        }
+
+        internal unsafe DacLibrary(DataTarget dataTarget, string dacPath, ulong runtimeBaseAddress)
```
leculver, line 94: This constructor should be public.  Otherwise folks will have no way to create this manually if they so chose to do so.

leculver (changes requested):
I have some minor feedback on a few things.  This looks good to me as soon as those are cleaned up.

Thanks!

leculver:
> Also, the "ArityTest" test is failing on TypeTests.cs line 183. I really don't think my changes caused this.

Yes, please ignore that.

leculver (approved):

PR #910, opened 2021-04-06
DbgEngDataReader.PointerSize should return the pointer size of the dumped process

catstrike:
Currently, ```DbgEngDataReader.PointerSize``` returns ```IntPtr.Size``` instead of the pointer size of the dumped process.
Fixed.


leculver (approved):

leculver:
Thanks!

PR #911, opened 2021-04-11
Expose ELF coredump reader as public

leculver:
I need to expose the ELF reader for another project.  It's still strongly encouraged that folks use DataReader/IDataReader/IMemoryReader and go through its abstractions but I've done the work here to clean up the ELF file and coredump readers in a way that matches the conventions of ClrMD.  Specifically:

- Moved the namespace to Microsoft.Diagnostics.Runtime.Utilities (same as PEImage).
- The ELF code used `long` for addresses, this change fully brings all public surface to use `ulong`.  (There may still be some longs lurking the code, but they should not be public.)
- Not all classes and concepts are exposed, this is just enough to get basic information for now.
- Longer term I'd like to look into doing more with the export symbol table, but I'd like to do so in a way that's unified between PEImage/ElfFile.


leculver:
@nxtn FYI, if you have any feedback here too.

mikem8361:
Here is what I used to test the symbol lookup code. It uses hard coded paths to some local dumps which I put here: \\bvtsrv2\team\mikem\LineNums

```
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

using Microsoft.Diagnostics.Runtime.Linux;
using System.IO;
using System.Linq;
using Xunit;

namespace Microsoft.Diagnostics.Runtime.Tests
{
    public class ElfTests
    {
        [Fact]
        public void ElfCoreSymbolTests()
        {
            Stream stream = File.OpenRead(@"c:\sd\dumps\TestAssets\Linux\x64\Debug\5.0.0\LineNums\SOS.LineNums.Heap.dmp");
            var dump = new ElfCoreFile(stream);
            ElfLoadedImage image = dump.LoadedImages.Values.FirstOrDefault((image) => image.Path.Contains("libcoreclr.so"));
            Assert.NotNull(image);

            ElfFile file = image.Open();
            Assert.NotNull(file);

            Assert.True(file.DynamicSection.TryLookupSymbol("g_dacTable", out ElfSymbol symbol));
            Assert.True(symbol.Value == 0x00000000006d01a8);
        }

        [Fact]
        public void ElfFileSymbolTests()
        {
            Stream stream = File.OpenRead(@"c:\sd\dumps\TestAssets\Linux\x64\Debug\5.0.0\LineNums\libcoreclr.so");
            var file = new ElfFile(new Reader(new StreamAddressSpace(stream)));

            Assert.True(file.DynamicSection.TryLookupSymbol("g_dacTable", out ElfSymbol symbol));
            Assert.True(symbol.Value == 0x00000000006d01a8);
        }
    }
}
```

leculver:
> Here is what I used to test the symbol lookup code. It uses hard coded paths to some local dumps which I put here:

Up until now we've tried to avoid putting dump files into the repo to avoid large downloads.  I can add that test to Linux though for when we generate cores there.

mikem8361:
I was suggesting adding this test locally/temporarily just to make sure the symbol lookup isn't broken.

I am working on a "TestAsset" nuget package of dumps for diagnostics repo testing that maybe clrmd can leverage. It may be a while before I finish all of that up, but I'll let you know when it is done.

leculver:
> I was suggesting adding this test locally/temporarily just to make sure the symbol lookup isn't broken. I am working on a "TestAsset" nuget package of dumps for diagnostics repo testing that maybe clrmd can leverage.

That would be great!

FYI, this code doesn't work _before_ my changes:

```C#
        [Fact]
        public void ElfCoreSymbolTests()
        {
            ElfCoreFile dump = new ElfCoreFile(File.OpenRead(@"D:\linux_core\GCRoot_wks.dmp"));
            ElfLoadedImage image = dump.LoadedImages.Values.FirstOrDefault((image) => image.Path.Contains("libcoreclr.so"));
            Assert.NotNull(image);

            ElfFile file = image.Open();
            Assert.NotNull(file);

            Assert.True(file.DynamicSection.TryLookupSymbol("g_dacTable", out ElfSymbol symbol));
            Assert.NotEqual(0, symbol.Value);
        }
```

This is using a linux coredump generated from my test targets using coreclr 3.1.13.

```
  Message: 
    System.IO.IOException : I/O error occurred.
  Stack Trace: 
    Reader.Read[T](Int64& position) line 55
    ElfSymbolGnuHash.ctor(Reader reader, Boolean is64Bit, Int64 address) line 22
    ElfDynamicSection.ctor(Reader reader, Boolean is64Bit, Int64 address, Int64 size) line 65
    ElfFile.get_DynamicSection() line 63
    ThreadReaderTests.ElfCoreSymbolTests() line 87
```

Is the right data not put into the dump unless it's 5.0+?

mikem8361:
I haven't tried symbol lookup for any 3.1 coredumps yet but it wasn't until 5.0 that the g_dacTable export was added and that createdump added the symbol related memory. You need 5.0 core dumps to test this.

leculver:
Ah got it.  Yes, it passes on the 5.0 dumps you gave me, but the original code throws an exception from a property when you give it a 3.1 dump file.  That's not something that should happen.  Changeset 486591e fixes that.

mikem8361:
One thing I forgot to ask if there way to get the ElfFile for a ModuleInfo?

leculver:
> One thing I forgot to ask if there way to get the ElfFile for a ModuleInfo?

Not in this checkin, but in the future yes.  I still want people to use ModuleInfo/DataTarget/IDataReader/etc for 99% of cases.  I'm adding ElfCoreFile and ElfFile as a way to "break glass in case you can't get the info any other way".

Longer term my plans here are:

1. Eventually make other concepts public (e.g. ElfDynamicSection).
2. Allow you to get ElfFile ModuleInfo.
3. Do more work with export symbols to be able to find near matches for stack unwinds etc.  I need this for several other projects.

I'm intentionally going slow here though because I want to make sure the public API surface we ship is solid.  I made a lot of mistakes early on in ClrMD by building something that was half thought through then having to clean it up later with breaking changes, obsolete, etc.

I'm trying to break this work into manageable chunks that I (hopefully) can get right the first time around.

PR #915, opened 2021-04-30
Implement IDataReader.GetThreadContext() for MacOSProcessDataReader

k15tfu:
This allows us to walk through the stack on macOS.

Related issue: #888

The changes for macOS ARM64 will be in a different PR.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
@@ -3,10 +3,13 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
+using Microsoft.Diagnostics.Runtime.Utilities;
 
 namespace Microsoft.Diagnostics.Runtime.MacOS
 {
@@ -15,6 +18,7 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
         private readonly int _task;
 
         private ImmutableArray<MemoryRegion>.Builder _memoryRegions;
+        private readonly Dictionary<ulong, uint> _threadActs = new(); // map of thread id (uint64_t) -> thread (thread_act_t)
 
         private bool _suspended;
         private bool _disposed;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
```diff
@@ -221,10 +225,62 @@ private unsafe T Read<T>(ref ulong address)
...
+
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(regSize);
+            try
+            {
+                fixed (byte* data = buffer)
+                {
+                    int stateCount = Native.x86_THREAD_STATE64_COUNT;
+                    int kr = Native.thread_get_state(threadAct, Native.x86_THREAD_STATE64, new IntPtr(data), ref stateCount);
+                    if (kr != 0)
+                        throw new ClrDiagnosticsException($"thread_get_state failed with status code 0x{kr:x}");
```
leculver, line 244: If we can't get the context, you should return `false` here instead of throwing.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
```diff
@@ -221,10 +225,62 @@ private unsafe T Read<T>(ref ulong address)
...
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(buffer);
+            }
+
+            return true;
+        }
+
+        private unsafe void LoadThreads()
```
leculver, line 257: Similarly, it would be good to not throw in this function if threads cannot be loaded.

leculver (changes requested):
I have a minor request to change the exception behavior here, otherwise this looks good to me.

Thanks for working on this!

src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
k15tfu: Hi! Thanks for your review. I agree with the another one, but in this case I'm not sure if this is what we want. It behaves similarly to private `LinuxLiveDataReader.LoadThreads` and `MacOSProcessDataReader.LoadMemoryRegions`.

Do you suggest to get Native.thread_info() for all threads, even if some of them fail?

leculver: Good point.  The behavior should be consistent across all platforms, but it's not today...  I need to go back and take a look and make it all consistent.

In general, we want functions which return `bool` to return false on failure, and push that upwards in the stack.  So LoadThreads both here and in Linux should probably have a `bool` return value instead of throwing.

The intent is that even if we cannot walk threads or walk callstacks, a tool should still be able to walk the heap without hitting exceptions.  I'll flag an issue to take a look at this more broadly.  We won't block this checkin on that though since, as you point out, it's just not consistent today.

FYI, I am out sick today.  I will take a closer look at the other Pull Requests on monday.

leculver (approved):
I will take a closer look at exception handling later, so we can merge this as-is.  Thanks!

I will take a look at the other PRs early next week.

PR #917, opened 2021-04-30
Add support for macOS ARM64 in MacOSProcessDataReader

k15tfu:
Also adds arm_thread_state64_t struct for the case when ptrauth is disabled.

Related issue: #888

Based on #915, review it first pls.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
@@ -12,6 +12,7 @@ using System.Runtime.InteropServices;
 using System.Text;
 using Microsoft.Diagnostics.Runtime.DataReaders.Implementation;
 using Microsoft.Diagnostics.Runtime.Utilities;
+using ProcessArchitecture = System.Runtime.InteropServices.Architecture;
 
 namespace Microsoft.Diagnostics.Runtime.MacOS
 {
@@ -56,6 +57,13 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
 
                 _suspended = true;
             }
+
+            Architecture = RuntimeInformation.ProcessArchitecture switch
+            {
+                ProcessArchitecture.X64 => Architecture.Amd64,
+                ProcessArchitecture.Arm64 => Architecture.Arm64,
+                _ => Architecture.Unknown
+            };
         }
 
         ~MacOSProcessDataReader() => Dispose(false);
@@ -66,7 +74,7 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
 
         public OSPlatform TargetPlatform => OSPlatform.OSX;
 
-        public Architecture Architecture => Architecture.Amd64;
+        public Architecture Architecture { get; }
 
         public int ProcessId { get; }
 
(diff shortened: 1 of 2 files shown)


leculver:
I will review once #915 is checked in and this PR doesn't contain that change too.  Thanks for working on this!

src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
```diff
@@ -231,20 +239,34 @@ public unsafe bool GetThreadContext(uint threadID, uint contextFlags, Span<byte>
...
                 return false;
 
-            int regSize = sizeof(x86_thread_state64_t);
+            int stateFlavor;
+            int stateCount;
+            int regSize;
+            (stateFlavor, stateCount, regSize) = Architecture switch
+            {
+                Architecture.Arm64 => (Native.ARM_THREAD_STATE64, Native.ARM_THREAD_STATE64_COUNT, sizeof(arm_thread_state64_t)),
+                _ => (Native.x86_THREAD_STATE64, Native.x86_THREAD_STATE64_COUNT, sizeof(x86_thread_state64_t))
```
leculver, line 248: This second switch should be `Architecture.X86` and the default case should be `_ =>throw new PlatformNotSupportedException`.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
```diff
@@ -231,20 +239,34 @@ public unsafe bool GetThreadContext(uint threadID, uint contextFlags, Span<byte>
...
                         return false;
                 }
 
-                Unsafe.As<byte, x86_thread_state64_t>(ref MemoryMarshal.GetReference(buffer.AsSpan())).CopyContext(context);
+                switch (Architecture)
+                {
+                    case Architecture.Arm64:
+                        Unsafe.As<byte, arm_thread_state64_t>(ref MemoryMarshal.GetReference(buffer.AsSpan())).CopyContext(context);
+                        break;
+                    default:
```
leculver, line 266: This should be `case Architecture.X86`.  In the future if more architectures are added and supported for OS X, it makes the code less error prone.

leculver (changes requested):
A couple of minor issues, but otherwise it looks good when those are fixed.  Thanks!

src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
mikem8361: Don't you mean "Architecture.X64"? or both X86/X64?  Currently .NET Core only supports X64 and ARM64 (mono may support an X86 iOS, but I don't know for sure).

leculver: Ah, right.  This is also why it should be explicit and not the default case.

k15tfu: Yep, `default` is for `Architecture.Amd64`. Surely I can change it, but please also note that this style is used in other places, for example:
https://github.com/microsoft/clrmd/blob/d0409daab70c27dcf4c8b8d39cf91c11ae0773e6/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs#L370
https://github.com/microsoft/clrmd/blob/d0409daab70c27dcf4c8b8d39cf91c11ae0773e6/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs#L229

leculver (approved):
LGTM thanks!

PR #918, opened 2021-05-01
Add Destroy() override to ComCallableIUnknown

mikem8361:
This allows wrappers that use ComCallableIUnknown to clean up any
internal state on ref count == 0.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Utilities/COMInterop/ComCallableIUnknown.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Utilities/COMInterop/ComCallableIUnknown.cs
@@ -119,20 +119,19 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
                     {
                         Destroy();
                     }
-                    catch
+                    finally
                     {
-                    }
+                        foreach (IntPtr ptr in _interfaces.Values)
+                        {
+                            IntPtr* val = (IntPtr*)ptr;
+                            Marshal.FreeHGlobal(*val);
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Runtime/src/Utilities/COMInterop/ComCallableIUnknown.cs
```diff
@@ -115,6 +115,14 @@ private int ReleaseImpl(IntPtr self)
                 // Only free memory the first time we reach here.
                 if (_handle.IsAllocated)
                 {
+                    try
+                    {
+                        Destroy();
+                    }
+                    catch
```
leculver, line 122: Should we have a blanket catch here?  I would think we'd rather not have anyone who implemented `Destory` throw an exception...they could wrap in try/catch.

If the intent is to ensure cleanup, maybe a `try { Destroy(); } finally { ...cleanup code below... }` pattern instead.  We'd end up releasing the memory correctly and then propagate the exception back out.

leculver (commented):
I'm not sure the intent of the blanket `catch` block.  Maybe a try/finally is more appropriate?

src/Microsoft.Diagnostics.Runtime/src/Utilities/COMInterop/ComCallableIUnknown.cs
mikem8361: I switch it to a try/finally.

leculver (approved):

PR #926, opened 2021-05-20
Kill the process created by DataTarget.CreateSnapshotAndAttach

### LINKED ISSUE #925: CreateSnapshotAndAttach does not kill the created process
I would expect calling `Dispose` on the returned target to kill the suspended process created by `CreateSnapshotAndAttach`.

The problem can be seen with this simple sample:
```csharp
static void Main()
{
    for (int i = 0; i < 10; i++)
    {
        using var dt = DataTarget.CreateSnapshotAndAttach(Process.GetCurrentProcess().Id);
    }
}
 ```

 If you break on the last line you can see that the 10 processes are still there: 
 
image

While looking at the code, I noticed that adding in `WindowsProcessDataReader.Dispose`   

```csharp
Process.GetProcessById(ProcessId)?.Kill();
```
does fix the issue.

I'd be happy to propose a PR!

jairbubbles:
Closes #925

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Windows/WindowsProcessDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Windows/WindowsProcessDataReader.cs
@@ -104,8 +104,9 @@ namespace Microsoft.Diagnostics.Runtime
                         // Kill the cloned process as we don't need it anymore
                         Process.GetProcessById(ProcessId).Kill();
                     }
-                    catch (Win32Exception)
+                    catch (Exception e)
                     {
+                        Trace.WriteLine($"Unable to kill the target process: {e}");
                     }
                 }
 
--- a/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Windows/WindowsProcessDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataReaders/Windows/WindowsProcessDataReader.cs
@@ -99,8 +99,14 @@ namespace Microsoft.Diagnostics.Runtime
 
                 if (_originalPid != 0)
                 {
-                    // Kill the cloned process as we don't need it anymore
-                    Process.GetProcessById(ProcessId)?.Kill();
+                    try
+                    {
+                        // Kill the cloned process as we don't need it anymore
+                        Process.GetProcessById(ProcessId).Kill();
+                    }
+                    catch (Win32Exception)
+                    {
+                    }
                 }
 
                 _disposed = true;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataReaders/Windows/WindowsProcessDataReader.cs
```diff
@@ -97,6 +97,18 @@ private void Dispose(bool _)
...
                     WindowsFunctions.NativeMethods.CloseHandle(_process);
 
+                if (_originalPid != 0)
+                {
+                    try
+                    {
+                        // Kill the cloned process as we don't need it anymore
+                        Process.GetProcessById(ProcessId).Kill();
+                    }
+                    catch (Win32Exception)
```
leculver, line 107: This should be a blanket `catch {}` statement.

Usually I don't like adding those, but the original reason that block was removed is that exceptions were escaping this exact `catch (Win32Exception)` block and making it impossible to properly have a `using DataTarget dt = ...` statement without wrapping the whole thing in a try/catch.

leculver: Or better yet, something like this (or similar):

``` C#
catch (Exception e)
{
    Trace.WriteLine($"Unable to kill the target process: {e.ToString()}");
}
```

leculver (changes requested):
Interesting...I didn't realize the original change left hanging processes.

There seems to be some odd state where the process is gone by the time we get to `Kill` and sometimes not...  In either case, adding this back as it was originally causes issues.  We need to catch all possible exceptions coming out of GetProcessById/Kill with a blanket catch block to ensure `Dispose` doesn't throw.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Windows/WindowsProcessDataReader.cs
jairbubbles: Done!

leculver (approved):

leculver:
LGTM!  I should have a new version released early next week with this fix.

ltrzesniewski:
Shouldn't the `Process` object be disposed even though it's killed?

jairbubbles:
That's a good remark @ltrzesniewski! I discovered recently that `Process`  is disposable but I've never seen patterns where we dispose it like that:
```
using (Process.Start())
{
}
```

I guess we rely on the finalizer.

src/Microsoft.Diagnostics.Runtime/src/DataReaders/Windows/WindowsProcessDataReader.cs
```diff
@@ -97,6 +97,19 @@ private void Dispose(bool _)
                 if (_process != IntPtr.Zero)
                     WindowsFunctions.NativeMethods.CloseHandle(_process);
 
+                if (_originalPid != 0)
+                {
+                    try
+                    {
+                        // Kill the cloned process as we don't need it anymore
+                        Process.GetProcessById(ProcessId).Kill();
```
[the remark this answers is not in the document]
sungam3r: ok

PR #930, opened 2021-06-17
Fix the check of DAC & CoreCLR versions when they are not versioned

### LINKED ISSUE #929: Could not find matching DAC for this runtime if both DAC & CoreCLR are not versioned
Hi! I'm getting this error on dev (local) builds of coreclr:
https://github.com/microsoft/clrmd/blob/957981f36eeccb6e9d266407df6522ca5cfbd899/src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs#L94

Actually, `Microsoft.Diagnostics.Runtime.IDataReader.GetVersionInfo()` fails for my libcoreclr.dylib, but then it returns the `default` as non-nullable property `Version`:
https://github.com/microsoft/clrmd/blob/957981f36eeccb6e9d266407df6522ca5cfbd899/src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs#L151-L161

So when we check the DAC matches CoreCLR in `PlatformFunctions.IsEqualFileVersion()`, on the one hand `Microsoft.Diagnostics.Runtime.PlatformFunctions.GetFileVersion()` fails for my libmscordaccore.dylib either, but on the other hand the version of libcoreclr.dylib is shown as `0.0.0.0`:
https://github.com/microsoft/clrmd/blob/957981f36eeccb6e9d266407df6522ca5cfbd899/src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/PlatformFunctions.cs#L31-L37

Can we do this check only if at least one of them has a version? cc @leculver 
Is `Microsoft.Diagnostics.Runtime.ModuleInfo.Version` a non-nullable property on purpose?

k15tfu:
Fixes #929

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/PlatformFunctions.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Utilities/Platform/PlatformFunctions.cs
@@ -31,7 +31,12 @@ namespace Microsoft.Diagnostics.Runtime
         public virtual bool IsEqualFileVersion(string file, VersionInfo version)
         {
             if (!GetFileVersion(file, out int major, out int minor, out int revision, out int patch))
+            {
+                if (version == default)
+                    return true;  // the runtime module has no version info either
+
                 return false;
+            }
 
             return major == version.Major && minor == version.Minor && revision == version.Revision && patch == version.Patch;
         }


k15tfu:
@leculver Friendly ping.

leculver:
Hello!

I have been on vacation for the past two weeks.  I unfortunately forgot to set a GitHub away message.  I will take a look at this early next week as I'm just back into the office today and have a lot of meetings.

PR #946, opened 2021-08-24
Fix ELF the module base address for single file apps

mikem8361:
Verify that ELF image has a base address

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Linux/ElfCoreFile.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Linux/ElfCoreFile.cs
@@ -211,7 +211,7 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
                 ArrayPool<byte>.Shared.Return(bytes);
             }
 
-            return lookup.Values.ToImmutableDictionary(i => i.BaseAddress);
+            return lookup.Values.ToImmutableDictionary(i => i.BaseAddress != 0 ? i.BaseAddress : throw new InvalidDataException($"Module {i} has no base address"));
         }
 
         public void Dispose()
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/Linux/ElfCoreFile.cs
```diff
@@ -211,7 +211,7 @@ private void LoadAuxvTable()
                 ArrayPool<byte>.Shared.Return(bytes);
             }
 
-            return lookup.Values.ToImmutableDictionary(i => i.BaseAddress);
+            return lookup.Values.ToImmutableDictionary(i => i.BaseAddress != 0 ? i.BaseAddress : throw new InvalidDataException($"Module {i} has no base address"));
```
leculver, line 214: I'm not sure this should be a throw here.  We don't control coredump generation for the world, and if we end up seeing a BaseAddress == 0 in a coredump generated by a 3rd party program then we won't be able to debug the coredump at all, even if that part of the file table is irrelevant to CLR debugging.

What about this?

```suggestion
            return lookup.Values.Where(i => i.BaseAddress != 0).ToImmutableDictionary(i => i.BaseAddress);
```

leculver (changes requested):

src/Microsoft.Diagnostics.Runtime/src/Linux/ElfCoreFile.cs
mikem8361: I was wondering if throwing was too extreme. Your suggestion is fine but it will hide that there is a module with no base address.  Since the base address is used as a key, leaving them in the dictionary doesn't work either. 

I'll push your suggestion.

leculver: I'm happy to hear other suggestions if there's another way to fix it.  My only concern is that if there are _multiple_ `BaseAddress == 0` I would assume that ToImmutableDictionary would throw due to duplicate keys.  I'm not sure of another solution if there happens to be two files with baseaddress == 0 then the whole concept of using a dictionary is suspect...  I would imagine that'd take more thought to know what to sensibly do there.

PR #952, opened 2021-08-30
Fix .NET 5.0 MacOS ELF dump module list

mikem8361:
This fixes what a problem with my previous single-file module fix. The hacky MacOS ELF dumps that 5.0 generates (6.0 now generates actual MachO dumps) have modules with no PageOffset == 0 file entries so my previous change doesn't add them to the list of modules.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Linux/ElfLoadedImage.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Linux/ElfLoadedImage.cs
@@ -4,6 +4,7 @@
 
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 
 namespace Microsoft.Diagnostics.Runtime.Utilities
 {
@@ -12,7 +13,7 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
     /// </summary>
     public class ElfLoadedImage
     {
-        private readonly List<ElfFileTableEntryPointers64> _fileTable = new(4);
+        private readonly SortedList<ulong, ElfFileTableEntryPointers64> _fileTable = new(4);
         private readonly Reader _vaReader;
         private readonly bool _is64bit;
         private ulong _end;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/Linux/ElfLoadedImage.cs
```diff
@@ -12,7 +13,7 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
     /// </summary>
     public class ElfLoadedImage
     {
-        private readonly List<ElfFileTableEntryPointers64> _fileTable = new(4);
+        private readonly SortedList<ulong, ElfFileTableEntryPointers64> _fileTable = new(4);
```
leculver, line 16: We don't actually need a sorted list do we?  If Is _fileTable used anywhere else?

What about `private int _minimumPointer;`.

In AddTableEntryPointers:  `if (_minimumPointer == 0 || pointers.Start < _minimumPointer) _minimumPointer = pointers.Start;`

And BaseAddress can be replaced by:

    private ulong _baseAddress;
    public BaseAddress { get => _baseAddress == 0 ? _minimumPointer : _baseAddress; } set => _baseAddress = value; }

Or something similar?  If we need to keep a sorted list of file table pointers for other reasons then this is fine as-is, but there's no reason to keep a sorted list just to find the minimum address.

leculver (changes requested):

leculver (approved):
Nevermind my previous comments, this is fine.

mikem8361:
Your idea/comment sounded good. I would have to think about it some more.  I liked the idea of removing the sorted _fileTable list. I guess it is too late now.  What changed your mind about your comment?

leculver:
> What changed your mind about your comment?

We have to linearly search the entries for the smallest entry which is O(n), and keeping a sorted list I think is still O(n) so it doesn't save us much eliminating the sorted list.  I also didn't have a strong opinion either way, so I figured it was easiest to just merge it.  If you want to submit another PR to remove it feel free.

PR #955, opened 2021-09-29
Add IExportSymbols support to PEImage modules

mikem8361:
Added PEImage.TryGetExportSymbol public method.

Clean up and refactor of ELF symbol export lookup. Added ElfFile.TryGetExportSymbol public method.

Added ElfFile constructor that allows a memory stream to be passed.

Added IExportReader implementation to MinidumpReader and DbgEngDataReader.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Linux/ElfFile.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Linux/ElfFile.cs
@@ -64,6 +64,29 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
         }
 
         /// <summary>
+        /// Returns the address of a module export symbol if found
+        /// </summary>
+        /// <param name="symbolName">symbol name (without the module name prepended)</param>
+        /// <param name="offset">symbol offset returned</param>
+        /// <returns>true if found</returns>
+        public bool TryGetExportSymbol(string symbolName, out ulong offset)
+        {
+            try
+            {
+                if (DynamicSection is not null && DynamicSection.TryLookupSymbol(symbolName, out ElfSymbol? symbol) && symbol is not null)
+                {
+                    offset = (ulong)symbol.Value;
+                    return true;
+                }
+            }
+            catch (Exception ex) when (ex is IOException || ex is InvalidDataException)
+            {
+            }
+            offset = 0;
+            return false;
+        }
+
+        /// <summary>
         /// The ELFDynamicSection for this file, if it exists.
         /// </summary>
         internal ElfDynamicSection? DynamicSection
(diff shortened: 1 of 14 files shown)


src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -257,6 +259,77 @@ public int Read(int virtualAddress, Span<byte> dest)
...
+                }
+            }
+            catch (Exception ex) when (ex is IOException || ex is InvalidDataException)
+            {
+            }
+            offset = 0;
+            return false;
+        }
+
+        private string ReadNullTerminatedAscii(int rva)
```
leculver, line 311: This should take a max length.  The default max length can be very high (like 4096 or 8192), but there should definitely be a cap to the number of bytes read since malformed or weird crash dumps could cause an incredibly, incredibly slow read until we OOM.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -257,6 +259,77 @@ public int Read(int virtualAddress, Span<byte> dest)
...
+        /// Returns the address of a module export symbol if found
+        /// </summary>
+        /// <param name="symbolName">symbol name (without the module name prepended)</param>
+        /// <param name="offset">symbol offset returned</param>
+        /// <returns>true if found</returns>
+        public bool TryGetExportSymbol(string symbolName, out ulong offset)
+        {
+            try
+            {
+                if (_exportDirectory is null)
```
leculver, line 272: Nit:  The rest of ClrMD uses .HasValue and .Value instead of comparing to null.

src/Microsoft.Diagnostics.Runtime/src/Linux/ElfFile.cs
```diff
@@ -63,6 +63,29 @@ internal Reader VirtualAddressReader
...
+        /// Returns the address of a module export symbol if found
+        /// </summary>
+        /// <param name="symbolName">symbol name (without the module name prepended)</param>
+        /// <param name="offset">symbol offset returned</param>
+        /// <returns>true if found</returns>
+        public bool TryGetExportSymbol(string symbolName, out ulong offset)
+        {
+            try
+            {
+                if (DynamicSection is not null && DynamicSection.TryLookupSymbol(symbolName, out ElfSymbol? symbol) && symbol is not null)
```
leculver, line 76: Not something you need to fix in this checkin, but .TryLookupSymbol should not throw since the name starts with Try.  I need to remember to look at that later...

leculver (changes requested):
One minor change requested around reading ascii, one nitpick, and one note for future work.  Thanks!

src/Microsoft.Diagnostics.Runtime/src/Linux/ElfFile.cs
mikem8361: DynamicSection.TryLookupSymbol should throw any exceptions.  There is a try/catch in it for IOExceptions (read failures).  The reason this function (TryGetExportSymbol) has a try/catch is that getting the dynamic section (DynamicSection) can throw those exceptions.  Maybe this property should catch the exceptions and return null.

PR #956, opened 2021-10-01
Add initial Mach-O coredump support

leculver:
Still requires some cleanup.  This checkin is to get a build into the system.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/MacOS/Structs/MachHeader64.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/MacOS/Structs/MachHeader64.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
     [StructLayout(LayoutKind.Sequential, Pack = 1)]
     internal readonly struct MachHeader64
     {
+        public static unsafe uint Size => (uint)sizeof(MachHeader64);
         public const uint Magic64 = 0xfeedfacf;
 
         public uint Magic { get; }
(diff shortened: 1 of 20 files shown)


mikem8361:
It looks like the official CI build broke: https://dnceng.visualstudio.com/internal/_build/results?buildId=1396820&view=results

     4>D:\workspace\_work\1\s\src\Microsoft.Diagnostics.Runtime\src\MacOS\MachOCoreDump.cs(260,1): error RS0101: Avoid multiple blank lines [D:\workspace\_work\1\s\src\Microsoft.Diagnostics.Runtime\Microsoft.Diagnostics.Runtime.csproj]

src/Microsoft.Diagnostics.Runtime/src/MacOS/Structs/MachHeader64.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Runtime.MacOS
+{
+    [StructLayout(LayoutKind.Sequential, Pack = 1)]
+    internal readonly struct MachHeader64
+    {
+        public static unsafe uint Size => (uint)sizeof(MachHeader64);
+        public const uint Magic64 = 0xfeedfacf;
```
Therzok, line 13: MachO files could also be the `universal`/`fat` binary format, where the first bytes are defined the [fat_header](https://github.com/apple/darwin-xnu/blob/main/EXTERNAL_HEADERS/mach-o/fat.h#L48-L54), followed by per-CPU architecture headers.

PR #968, opened 2021-12-05
Fix leaking mach ports in MacOSProcessDataReader.Dispose()

k15tfu:


src/Microsoft.Diagnostics.Runtime/src/DataReaders/MacOS/MacOSProcessDataReader.cs
```diff
@@ -84,11 +84,15 @@ public void Dispose()
...
 
-        private void Dispose(bool _)
+        private void Dispose(bool disposing)
         {
+            _ = disposing;
             if (_disposed)
                 return;
 
+            foreach (uint threadAct in _threadActs.Values)
+                _ = Native.mach_port_deallocate(Native.mach_task_self(), threadAct);
```
Therzok, line 94: Hoist mach_task_self above the loop?

Therzok: Just came across this PR again, mach_task_self could actually be cached on the process reader instance.

leculver:
Thanks for the change, sorry it took so long for me to get around to this.

PR #984, opened 2022-03-23
Fix reading dumps from MacOS version 12 or greater (Monterey)

mikem8361:
Change the symbol export lookup to first search the exports and then all the symbols.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOModule.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOModule.cs
@@ -121,7 +121,7 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
             if (symTable is null)
                 return false;
 
-            for (uint i = 0; i < nsyms; i++)
+            for (uint i = 0; i < nsyms && start + i < symTable.Length; i++)
             {
                 string name = GetSymbolName(symTable[start + i], symbol.Length + 1);
                 if (name.Length > 0)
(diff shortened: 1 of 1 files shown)


src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOModule.cs
```diff
@@ -94,18 +94,36 @@ public bool TryLookupSymbol(string symbol, out ulong address)
...
+        private bool TryLookupSymbol(uint start, uint nsyms, string symbol, out ulong address)
+        {
+            address = 0;
 
             NList64[]? symTable = ReadSymbolTable();
             if (symTable is null)
                 return false;
 
-            for (int i = 0; i < _dysymtab.nextdefsym; i++)
+            for (uint i = 0; i < nsyms; i++)
```
leculver, line 124: This doesn't fully bounds check `symTable`.  Perhaps something like this, for defensive coding purposes?

```suggestion
            for (uint i = 0; i < nsyms && start + i < symTable.Length; i++)
```

leculver (changes requested):
One small nitpick, otherwise looks good to me!

src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOModule.cs
mikem8361: Good catch.  I pushed the changes.

PR #985, opened 2022-03-24
Fixed bug preventing the last sync block item to be considered for in…

Ephron-WL:
…clusion in one of the sync block lists.

If you run this method and count the number of items added to the three lists the total count is one less than the TotalSyncBlockCount.  That is a result of this bug.


leculver:
@mikem8361 Thoughts on this PR, wrt https://stackoverflow.com/questions/2203000/windbg-sos-explanation-of-syncblk-output?

mikem8361 (approved):

Issue #947, opened 2021-08-26
System.AccessViolationException when reading dump 


PhDuck:
```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Repeat 2 times:
--------------------------------
   at Microsoft.Diagnostics.Runtime.Windows.CacheNativeMethods+Memory.memcpy(IntPtr, UIntPtr, UIntPtr)
--------------------------------
   at Microsoft.Diagnostics.Runtime.Windows.ArrayPoolBasedCacheEntry.CopyDataFromPage(Microsoft.Diagnostics.Runtime.Windows.CachePage`1<Byte[]>, IntPtr, UInt32, UInt32)
   at Microsoft.Diagnostics.Runtime.Windows.CacheEntryBase`1[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].ReadPageDataFromOffset(Int32, UInt32, UInt32, IntPtr, System.Func`3<UIntPtr,UInt32
,UInt32>)
   at Microsoft.Diagnostics.Runtime.Windows.CacheEntryBase`1[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].GetDataForAddress(UInt64, UInt32, IntPtr, UInt32 ByRef)
   at Microsoft.Diagnostics.Runtime.Windows.CachedMemoryReader.TryReadMemory(UInt64, Int32, IntPtr)
   at Microsoft.Diagnostics.Runtime.Windows.CachedMemoryReader.Read(UInt64, System.Span`1<Byte>)
   at Microsoft.Diagnostics.Runtime.MemoryReader.MoveToPage(UInt64)
   at Microsoft.Diagnostics.Runtime.MemoryReader.ReadPtr(UInt64, UInt64 ByRef)
   at Microsoft.Diagnostics.Runtime.Implementation.ClrmdSegment+<EnumerateObjects>d__39.MoveNext()
   at System.Linq.Enumerable+SelectManySingleSelectorIterator`2[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[Microsoft.Diagnostics.Runtime.ClrObject, Microsoft.Diagnostics.Runtime, Version
[...]

PR #990, opened 2022-04-26
Mini Dump segment size might be bigger than Int32 can fit

catstrike:
Fix for #947.
Mini Dump can have large segments. Currently, unit is used to store the size of a segment and it might be not enough in some cases. In particular, I have a dump of IIS process that has a large (more than 8Gb) segment. When I'm trying to analyze it with ClrMD it crashes with System.AccessViolationException.


leculver:
LGTM, thanks!

Issue #977, opened 2022-02-04
Better error handling for Linux dumps loaded on Windows


noahfalk:
Currently if I load a linux core file when running CLRMD on Windows I get `FileNotFoundException 'Could not find matching DAC for this runtime'`. Could we improve this error to a NotSupportedException (assuming it isn't intended to be supported right now?)

Repro:
```C#
        static void Main(string[] args)
        {
            Environment.SetEnvironmentVariable("_NT_SYMBOL_PATH", "https://msdl.microsoft.com/download/symbols");
            DataTarget target = DataTarget.LoadDump("some_linux_dump");
            ClrInfo info = target.ClrVersions[0];
            ClrRuntime r = info.CreateRuntime();
        }
```

Expected result: A NotSupportedException (or working is cool too)
Actual result: FileNotFoundException

Debugging into it I see that CLRMD issues a symbol server request for mscordaccore like this:
`mscordaccore.dll\000000006f7000\mscordaccore.dll`
If the scenario is intended to work CLRMD would instead need to issue a request for:
`mscordaccore.dll\elf-buildid-coreclr-2cb234c544d6a97745e89f207485dc027a86b7a5\mscordaccore.dll`
The 20 byte hex value is the buildid for coreclr binary and I confirmed CLRMD did parse that correctly if you wanted to go this route.

When I manually downloaded the correct xplat DAC and pointed CLRMD at it my simple test for runtime.EnumerateObjects().Count() gave me a reasonable number so this might be a viable workaround for others as well.

PS. CLRMD remains one of the most reliable repos where I can clone, build and debug in literally seconds with zero hassle. It made it really convenient for me to investigate this. Thanks 👍

leculver:
@noahfalk Just back from vacation, working on this now.  Along with more OS X, Linux, and Windows -> Linux debugging improvements.

Issue #927, opened 2021-05-27
Non-writable cascading downstream stores in _NT_SYMBOL_PATH cause System.UnauthorizedAccessException


dekarma:
(Microsoft.Diagnostics.Runtime package 2.0.226801)

I ran into a problem when one of the downstream stores in _NT_SYMBOL_PATH is not writable.

For example: 

`_NT_SYMBOL_PATH=symsrv*symsrv.dll*c:\symbols*\\someserver\symbols*http://msdl.microsoft.com/download/symbols`

where `c:\symbols` is writable but `\\someserver\symbols` is **not**.

This is a supported symsrv configuration according to [this document](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/advanced-symsrv-use#cascading-downstream-stores):

> When SymSrv processes a path that contains cascading stores, it will skip any store that it cannot read or write to. So if a share goes down, it will copy the file to the store downstream from the missing store without any error. A nice side effect of this error is that the user can specify more than one master store that feeds a single stream of downstream stores as long as the master stores are not writable.

The exception on `ClrRuntime runtime = runtimeInfo.CreateRuntime();` is **System.UnauthorizedAccessException: Access to the path '\\someserver\symbols\mscordacwks_amd64_amd64_4.8.4311.00.dll\5fbdcb6dac2000' is denied.**

```
System.AggregateException: One or more errors occurred. ---> System.UnauthorizedAccessException: Access to the path '\\someserver\symbols\mscordacwks_amd64_amd64_4.8.4311.00.dll\5fbdcb6dac2000' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.Directory.InternalCreateDirectory(String fullPath, String path, Object dirSecurityObj, Boolean checkHost)
   at System.IO.Directory.InternalCreateDirectoryHelper(String path, Boolean checkHost)
   at Microsoft.Diagnostics.Runtime.Implementation.SymbolServerLocator.<GetPhysicalFileFromServerAsync>d__14.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
[...]

Issue #771, opened 2020-06-19
Add `DataTarget.SymbolPath` or `IBinaryLocator.SymbolPath`


leculver:
The symbol locator implementation was intentionally marked internal, but there still should be a way to set the symbol path other than setting _NT_SYMBOL_PATH.  I also need to go do symbol resolution on Linux as well, so I will likely roll that change into this too.

sukesh-ak:
Ideally DataTarget functions should have a overload with SymbolPath parameter.

DataTarget.AttachToProcess
DataTarget.CreateFromDbgEng
DataTarget.CreateSnapshotAndAttach
DataTarget.LoadDump

baal2000:
BTW [GettingStarted.md](https://github.com/microsoft/clrmd/blob/master/doc/GettingStarted.md#getting-the-dac-from-the-symbol-server) still mentions `DataTarget.SymbolLocator` (no longer present in ClrMd v.2)

PR #995, opened 2022-05-11
Implement Linux/OS X symbol server requests

leculver:
- Properly implement the symbol server protocol for requesting ELF and Mach-O files.
- Unfortunately this requires version bumping ClrMD to 2.1 because this is a breaking change.
- This means we now support cross-OS debugging for the scenarios that CLR Diagnostics supports.  E.G. it's possible to debug Linux coredumps on a Windows machine for recent builds of .Net Core which support it.
- Theoretical support for OS X -> Windows debugging exists since we plan to build that eventually, but currently the CLR Diagnostics team has not built the dac PE image that can debug OS X dumps.
- Fixed some Linux and OS X build issues.
- Do a better job parsing symbol server paths.
- Better understand the kinds of symbol servers out there and make less requests that will never be filled.

Fixes https://github.com/microsoft/clrmd/issues/977.
Fixes https://github.com/microsoft/clrmd/issues/927.
Closes https://github.com/microsoft/clrmd/issues/771.


src/Microsoft.Diagnostics.Runtime/src/Common/IFileLocator.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Locates an Elf binary.
+        /// </summary>
+        /// <param name="fileName">The file name or path of the binary to locate.</param>
+        /// <param name="archivedUnder">The file or keyword that this binary is archived under, <see langword="null"/> if its archived under its own properties.</param>
+        /// <param name="buildId">The buildId of the Elf image to locate or the buildId of the image specified by <paramref name="archivedUnder"/>.</param>
+        /// <param name="checkProperties">Whether or not to validate that the given file matches the build id.</param>
+        /// <returns>A full path on disk (local) of where the binary was copied to or <see langword="null"/> if it was not found.</returns>
+        string? FindElfImage(string fileName, string? archivedUnder, ImmutableArray<byte> buildId, bool checkProperties);
```
mikem8361, line 40: I'm not sure about exposing "archivedUnder" part of the index in this public API.  The "coreclr" is part of symbol server protocol.  I think of this more abstractly as either it is an "Identity" (the module itself), "Runtime" (the coreclr special keys) or "Symbol" index/keys.  A enum might be a better approach.

I guess related to this is that this interface couldn't be used to download symbols (PDB or dwarf for Linux/MacOS). I'm sure this is intentional and a ISymbolFileLocator could be added if clrmd ever needs to download them.

mikem8361 (approved):

src/Microsoft.Diagnostics.Runtime/src/Common/IFileLocator.cs
leculver: > I'm not sure about exposing "archivedUnder" part of the index in this public API. The "coreclr" is part of symbol server protocol.

I agree it's odd, it just felt like the least-odd option.  I know there's not currently anything other than `coreclr` as the value, but it still feels weird not having `archivedUnder` and just silently assuming it's `coreclr`.

If we don't include it as a parameter here then it becomes a "hidden" part of the API.  For example, we'd need something like `FindElfImageUnderClrProperties(string fileName, string? archivedUnder, ImmutableArray<byte> buildId, bool checkProperties);` to make it clear that we would be using "coreclr" there.  (Because FindElfImage with that overload already exists.)  Which also feels weird.

> I guess related to this is that this interface couldn't be used to download symbols (PDB or dwarf for Linux/MacOS). I'm sure this is intentional and a ISymbolFileLocator could be added if clrmd ever needs to download them.

Correct, we don't support downloading anything other than images, for now.

mikem8361: I was suggesting replacing the string archiveUnder with a more abstract enum not removing it or assuming it is coreclr. But I guess the only caller of this interface is clrmd itself.  There might be other implementations outside clrmd, but they can can do the same thing as clrmd's impl.

leculver: Ahh interesting, I'll take a look at doing that.

Issue #988, opened 2022-03-30
Unable to create DACLibrary for minidump from self contained app


pavel-faltynek:
When trying to create runtime from self contained app minidump, like:

```
            const string Dump = @"dump.13600.dmp";
            using var target = DataTarget.LoadDump(Dump);
            var runtime = target.ClrVersions.First().CreateRuntime();
```

Getting an error:
`Microsoft.Diagnostics.Runtime.ClrDiagnosticsException: 'Failure loading DAC: CreateDacInstance failed 0x80131c4f'`
The DAC dll is located, `CLRDataCreateInstance` called for IXCLRDataProcess interface and then failing.

- This applies for windows app built as `<SelfContained>` with given `<RuntimeIdentifier>` (so the DAC dll is part of the published stuff).
- This applies for mini dumps created by (the latest) `createdump.exe` (with no other than process id parameters), e.g.: the one located in `c:\Program Files (x86)\dotnet\shared\Microsoft.NETCore.App\6.0.3` for 32 bits.
- This does not apply for full dump (e.g. created via `taskmgr.exe` - either bitness).

The problematic dump can be loaded in WinDbg with no issues, `!sosstatus` reports expected DAC.
We would like to use also clrmd for such dump analysis.

leculver:
Hello!  I have taken a look at this and found that there is a problem with dac requests in single file mode, but the underlying issue you reported is actually a bug in dotnet-dump.  This tool does not put the enough information into the dump file for any tool to correctly determine that we are looking at a .Net process.  Since I've fixed what I can out of ClrMD and reported a bug to the diagnostics team, I'm closing this issue.

For further updates, please check this bug:  https://github.com/dotnet/diagnostics/issues/3065

Thanks!

PR #997, opened 2022-05-11
Dac single file request fixes

leculver:
We would request the wrong long-name dac in the single file case.

Contributes to https://github.com/microsoft/clrmd/issues/988.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Diagnostics.Runtime
         /// <summary>
         /// If the application is single-file, this contains the runtime, DAC and DBI index information
         /// </summary>
-        public RuntimeInfo? SingleFileRuntimeInfo { get; internal set; }
+        public ClrRuntimeInfo? SingleFileRuntimeInfo { get; internal set; }
 
         /// <summary>
         /// To string.
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
@@ -161,12 +161,12 @@ namespace Microsoft.Diagnostics.Runtime
             if (!_clrs.IsDefault)
                 return _clrs;
 
-            IExportReader? exportReader = DataReader as IExportReader;
             Architecture arch = DataReader.Architecture;
             RuntimeInfo? singleFileRuntimeInfo = null;
             ImmutableArray<ClrInfo>.Builder versions = ImmutableArray.CreateBuilder<ClrInfo>(2);
             foreach (ModuleInfo module in EnumerateModules())
             {
+                string? dacAgnosticName = null;
                 ImmutableArray<byte> runtimeBuildId = ImmutableArray<byte>.Empty;
                 int runtimeTimeStamp = 0;
                 int runtimeFileSize = 0;
@@ -179,17 +179,21 @@ namespace Microsoft.Diagnostics.Runtime
                 }
                 else
                 {
-                    if (exportReader is null || !exportReader.TryGetSymbolAddress(module.ImageBase, RuntimeInfo.SymbolValue, out ulong runtimeInfoAddress))
+                    if (DataReader is not IExportReader exportReader || !exportReader.TryGetSymbolAddress(module.ImageBase, RuntimeInfo.SymbolValue, out ulong runtimeInfoAddress))
                         continue;
 
                     if (!DataReader.Read(runtimeInfoAddress, out RuntimeInfo runtimeInfo))
                         continue;
 
+
                     unsafe
                     {
+                        // 
                         string signature = Encoding.ASCII.GetString(runtimeInfo.Signature, RuntimeInfo.SignatureValueLength);
                         if (signature != RuntimeInfo.SignatureValue || runtimeInfo.Version <= 0)
                             continue;
+
+                        using PEImage? clrPE = module.GetPEImage();
 
                         platform = DataReader.TargetPlatform;
                         flavor = ClrFlavor.Core;
@@ -201,12 +205,14 @@ namespace Microsoft.Diagnostics.Runtime
                             {
                                 runtimeTimeStamp = BitConverter.ToInt32(new ReadOnlySpan<byte>(runtimeInfo.RuntimeModuleIndex + sizeof(byte), sizeof(int)).ToArray(), 0);
                                 runtimeFileSize = BitConverter.ToInt32(new ReadOnlySpan<byte>(runtimeInfo.RuntimeModuleIndex + sizeof(byte) + sizeof(int), sizeof(int)).ToArray(), 0);
+                                dacAgnosticName = ClrInfoProvider.GetDacFileName(flavor, platform);
                             }
                         }
                         else
                         {
                             // The first byte of the module indexes is the length of the build id
                             runtimeBuildId = new ReadOnlySpan<byte>(runtimeInfo.RuntimeModuleIndex + sizeof(byte), runtimeInfo.RuntimeModuleIndex[0]).ToArray().ToImmutableArray();
+                            dacAgnosticName = ClrInfoProvider.GetDacFileName(flavor, platform);
                         }
                     }
                 }
@@ -239,8 +245,14 @@ namespace Microsoft.Diagnostics.Runtime
                 }
 
                 VersionInfo version = module.Version;
-                string dacAgnosticName = ClrInfoProvider.GetDacRequestFileName(flavor, arch, arch, version, platform);
+
                 string dacRegularName = ClrInfoProvider.GetDacRequestFileName(flavor, IntPtr.Size == 4 ? Architecture.X86 : Architecture.Amd64, arch, version, platform);
+
+                if (dacAgnosticName is null)
+                    dacAgnosticName = ClrInfoProvider.GetDacRequestFileName(flavor, arch, arch, version, platform);
+                else
+                    dacRegularName = dacAgnosticName;
+
 
                 DacInfo dacInfo = new DacInfo(dacLocation, dacRegularName, dacAgnosticName, arch, runtimeFileSize, runtimeTimeStamp, version, runtimeBuildId);
                 versions.Add(new ClrInfo(this, flavor, module, dacInfo) {
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
@@ -185,15 +185,11 @@ namespace Microsoft.Diagnostics.Runtime
                     if (!DataReader.Read(runtimeInfoAddress, out RuntimeInfo runtimeInfo))
                         continue;
 
-
                     unsafe
                     {
-                        // 
                         string signature = Encoding.ASCII.GetString(runtimeInfo.Signature, RuntimeInfo.SignatureValueLength);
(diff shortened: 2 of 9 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
```diff
@@ -46,7 +46,7 @@ internal ClrInfo(DataTarget dt, ClrFlavor flavor, ModuleInfo module, DacInfo dac
         /// <summary>
         /// If the application is single-file, this contains the runtime, DAC and DBI index information
         /// </summary>
-        public RuntimeInfo? SingleFileRuntimeInfo { get; internal set; }
+        public ClrRuntimeInfo? SingleFileRuntimeInfo { get; internal set; }
```
mikem8361, line 49: Isn't this a breaking change? SOS/dotnet-dump uses it.

mikem8361: Never mind. It won't be a breaking change.  I thought the property name changed for some reason but it is only the type.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
```diff
@@ -199,14 +199,16 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
...
-                                runtimeFileSize = BitConverter.ToInt32(new ReadOnlySpan<byte>(runtimeInfo.RuntimeModuleIndex + sizeof(byte) + sizeof(int), sizeof(int)).ToArray(), 0);
+                                runtimeTimeStamp = BitConverter.ToInt32(new ReadOnlySpan<byte>(runtimeInfo.DacModuleIndex + sizeof(byte), sizeof(int)).ToArray(), 0);
+                                runtimeFileSize = BitConverter.ToInt32(new ReadOnlySpan<byte>(runtimeInfo.DacModuleIndex + sizeof(byte) + sizeof(int), sizeof(int)).ToArray(), 0);
+                                dacAgnosticName = ClrInfoProvider.GetDacFileName(flavor, platform);
                             }
                         }
                         else
                         {
                             // The first byte of the module indexes is the length of the build id
                             runtimeBuildId = new ReadOnlySpan<byte>(runtimeInfo.RuntimeModuleIndex + sizeof(byte), runtimeInfo.RuntimeModuleIndex[0]).ToArray().ToImmutableArray();
```
mikem8361, line 210: Is this supposed to be the runtime's build id or the DAC's?  I'm confused now. It looks like you changed what runtimeTimeStamp/runtimeFileSize from the runtime's (coreclr's) to the DAC's. But for non-single-file apps, it is the timestamp/filesize or build id of coreclr.

When I added the single-file support here I may have misunderstood what the parameters to DacInfo constructor should be.

mikem8361:
I was wrong about the long named cross-architecture DAC's. We do still build them and publish them.  We (clrmd and dotnet-dump ) do need to be able to download them to load for example a Windows arm64 single-file dumps on a Windows x64 machine. That is currently supported for non-single-file apps.  So you are right the RuntimeInfo doesn't have enough info to build the name (the coreclr's Version).  The DAC/DBI runtime "identity" info in the RuntineInfo data structure is only for the native architecture so the runtime's Version field will be needed (hopefully in 7.0).  The runtime timestamp/file size + the new runtime Version will be needed to build the long named dac info in clrmd.

Hopefully that makes sense. We can talk more about it.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
leculver: Oddly enough, both are archived!  You can request `mscordaccore.dll` under either CLR or the Dac's own properties and it shows up.  I thought it would be more consistent and reasonable to request under the dac's properties, even if both currently work.

It's really the long name `mscordaccore_*_*_*.dll` which is supposed to be under clr'dll's properties, but we need the version to request that one.

leculver:
It makes sense.  I think just adding coreclr.dll's properties to the `RuntimeInfo` struct should be enough to make sure we have all the information to request any of the dacs...which might help if we accidently miss archiving one of them.  Otherwise what's checked in now should work fine.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
leculver: It **is** a breaking change though, you will be compile-time correct, but not run-time correct.  This is why this version will be shipped as ClrMD `2.1` and not ClrMD `2.0` (the daily builds haven't reflected that version update yet, I mean nuget packages the version will be revved with the other PR I have open for symbols).

I can change this back if it's too much of a hassle.  However, I feel `RuntimeInfo` is way too generic.  I kept hitting it while trying to use [RuntimeInformation](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation?view=net-6.0).  I should have caught that when it was introduced, but I'm fixing it now that we are moving from 2.0 -> 2.1.

mikem8361: I don't mind the change at all and it shouldn't break SOS when it gets pushed to the diagnostics repo.  We get clrmd through an DARC update and not from nuget.

src/Microsoft.Diagnostics.Runtime/src/DataTargets/DataTarget.cs
mikem8361: Yes, both are archived on purpose under the special clr keys.

mikem8361: Even though they indexed under both keys, it seems to me that the definition of DacInfo.IndexFileSize/DacInfo.IndexTimeStamp was coreclr's. It looks like that is true if it is a non-singlefile-app.

mikem8361: I guess this is a technicality if those properties are only used to download the DAC.

mikem8361: This did break SOS (I'm looking at the DARC update PR right now) because it assumes that DacInfo.IndexFileSize/DacInfo.IndexTimeStamp is the runtime index info. I guess SOS is made an incorrect assumption. Why did you change this again?

leculver: The dac should be indexed (and requested under these properties:

long - mscordaccore_amd64_amd64_1.2.3.4.dll [clr-timestamp] [clr-build-filesize]
short - mscordaccore.dll [dac-timestamp] [dac-filesize]

Because we are requesting the short named dac "mscordaccore.dll" in this scenario, we should be requesting it with its own properties, right?

mikem8361: Both the long and short named DAC is indexed with both the identity key and coreclr's.  SOS and dotnet-symbol has depended on that for a long time. But for the long named DAC it should be the coreclr index for cross-architecture DACs.  

long - same architecture: indexed under both keys
long - cross architecture (i.e. mscordaccore_amd64_arm64_1.2.3.4.dll): indexed only under the coreclr key
short - always indexed under both keys

/cc: @hoyosjs

leculver: We are talking about single-file scenario here?  Or a regular CLR debugging scenario?  The code in question is ONLY hit for single-file.  And we provide the short name and dac properties.  Take a look:

``` C#
        static void Main(string[] args)
        {
            DataTarget dt = DataTarget.LoadDump(@"D:\work\single_file_full.dmp");
            foreach (ClrInfo info in dt.ClrVersions)
            {
                Console.WriteLine($"CLR: {info.ModuleInfo.FileName} {info.Flavor} {info.Version}");
                Console.WriteLine($"Dac: {info.DacInfo.PlatformSpecificFileName} timestamp:{info.DacInfo.IndexTimeStamp:x} filesize:{info.DacInfo.IndexFileSize}");

                string result = dt.BinaryLocator?.FindBinary(info.DacInfo.PlatformSpecificFileName, info.DacInfo.IndexTimeStamp, info.DacInfo.IndexFileSize, checkProperties: false);

                Console.WriteLine(result);

                // We use this:
                // DataTarget.BinaryLocator?.FindBinary(DacInfo.PlatformSpecificFileName, DacInfo.IndexTimeStamp, DacInfo.IndexFileSize, checkProperties: false);
            }
        }
```

```
CLR: c:\git\clrmd\src\TestTargets\single-file\publish\single-file.exe Core 1.0.0.0
Dac: mscordaccore.dll timestamp:6216ca94 filesize:1331200
SUCCESS: C:\Users\leecu\AppData\Local\Temp\symbols\mscordaccore.dll\6216ca94145000\mscordaccore.dll
```

So as you can see, this is a single file .exe, and we are providing the **short named dac** along with the **dac's filesize and timestamp**, and it's successfully requested from the symbol server.

Can you please run the code above on the dump that's failing?  I'm happy to make changes here if going back to the previous behavior is correct but first we need to understand what exactly is happening.

leculver: Also, perhaps while making breaking changes, it might be a good idea to redesign this DacInfo property and simply enumerate all valid dacs (which there can be multiple of) but I want to get to the bottom of the failure you are seeing first.

mikem8361: I should have been clearer about the SOS failure. It is the clrstack -i command that downloads DBI that is failing. All the other DAC related tests are ok.  The reason for the failure is the assumption I made about the DacInfo: that is is the coreclr index info.  SOS uses the DAC timestamp/file size to build the DBI index which after your change to the actual DAC identity key breaks the download.  So you could correctly argue that SOS is using these values incorrectly.  I can fix SOS to build the DBI key from the ClrRuntimeInfo.DbiModuleIndex.

leculver: Ahhh ok!  That makes more sense!  Don't update SOS yet.  Let me look at the design here and provide you the right information.  We are already in the realm of breaking changes, so at this point let me build something that's actually the data you want in a sensible place.

leculver: I'll have something for you by end of day.

mikem8361: There is no hurry. You could do all the work in your other breaking PR #995. 

In the end, it should be clear what kind of index info it is (just like I suggested in 995): "identity" or "coreclr".  I am ok with the new assumption about DacInfo index info being the DAC index but it needs to be consistent across single-file and non-single-file apps.  SOS can always get the "coreclr" info from ClrInfo.ModuleInfo.

leculver: I'm starting to work on this now.

One issue right now is that `ClrInfo.ModuleInfo` in single-file mode just gives you information about the single-file exe.  It's too strange to change that behavior to be CLR's info.  It's best to just expose a new API that gives you all of the relevant information I think.  ClrMD was designed to be a runtime inspection tool, and giving deep information about all of the various pieces of CLR diagnostic artifacts is well within its purview. I've just never sat down to enumerate all of the various bits of information that's available about CLR diagnostic artifacts.  The end result will be cleaner than what we have now I think.

I hope to have something by tomorrow.  The change I intend to make is a bit bigger than just spot fixing this issue.

PR #1000, opened 2022-05-14
Improve non-single file perf

leculver:
Only check for the single file export if we have a high likelihood of succeeding.


leculver:
@mikem8361: Is there any way to know on Linux or OS X that a module in a coredump is the "main" module?

mikem8361:
Yes there is on Linux/MacOS except on Linux the way .NET builds the hosts "dotnet" or the singlefilehost sets the type (e_type) field in the ELF header to DYN (shared object file) instead of the correct "EXEC" (executable).  

On MacOS, the filetype in the header is set properly to execute (MH_EXECUTE = 2).

mikem8361 (approved):

PR #1002, opened 2022-05-16
Various fixes from feedback

leculver:


leculver:
@mikem8361 This change addresses all of your offline review feedback.

I went with `ModuleInfo.TryCreate` using an `IDataReader` + `imageBase` instead of Stream (note that name can be set to string.Empty since you don't care about ModuleInfo.FileName).  It's possible to add a Stream version in the future, but the Mach-O code currently doesn't work on a Stream, so if I wanted to add a Stream based overload to ModuleInfo I'd have to go do more implementation work.

src/Microsoft.Diagnostics.Runtime/src/Common/ModuleInfo.cs
```diff
@@ -24,6 +70,11 @@ public abstract class ModuleInfo
         /// </summary>
         public string FileName { get; }
 
+        /// <summary>
+        /// The size of this image (may be different from <see cref="IndexFileSize"/>).
+        /// </summary>
+        public virtual long Size => IndexFileSize;
```
mikem8361, line 76: nit: ImageSize to match "ImageBase"?

mikem8361 (approved):
It looks you covered all of my concerns.

PR #1008, opened 2022-05-26
Mark some implementation as internal

leculver:


leculver:
@mikem8361 This is the last of the breaking changes that are being made.  From digging through SOS code (and our chat a couple weeks ago) I do not expect this will affect you.

PR #1010, opened 2022-05-27
Target only netstandard2 and .Net 6, remove Utilities

leculver:
Now target netstandard2.0 and .Net 6.  I've held on to support for older CLRs for as long as is tenable, but I really need to set a new starting point for the library for future development.  ClrMD 2.1 will be that cutoff point.

Similarly I've removed the Utiltiies library.  The IDebugClient wrappers provided are not something the library uses anymore, and all of the functionality that the library contained has been merged into the main library...at least all of the functionality I wanted in ClrMD.

As mentioned in other checkins, I'm trying to reduce the scope of ClrMD down to just the important bits of code that relate to inspecting CLR.  That's roughly where we are with 2.1.


leculver:
@mikem8361 I'm dropping .Net Core 3.1 and Desktop 4.6.1.  I think SOS should still be covered by Netstandard2.0 (or net6).

Let me know if this causes too many issues.

PR #1025, opened 2022-06-15
Change HResult return to int

leculver:
On x86 Linux the HResult struct is not handled the same way as an int.  Make all calls respect this.


leculver:
@mikem8361 FYI, this is the change I made for ClrMD to stop using HResult directly.  I still use HResult as my own internal returns and I assign the `int` I get from APIs to HResult.  This work went pretty quick.  I just Quick Action -> Introduce Local, named the local "hr" then changed the type to HResult.

@t-mustafin This is the other half of the Linux x86 work.

mikem8361:
You filed https://github.com/dotnet/diagnostics/issues/3101 to track this in the diagnostics repo. Hopefully we can find time to do this in the next couple of weeks.

leculver:
No rush on my account, this is just an FYI.  I'm asking t-mustafin to test the ClrMD part of this work this week so if there's still more problems you will get a heads up before you start the SOS work.

PR #1033, opened 2022-07-02
Support reading the first 4K of a PEImage

cshung:
During the initialization of CLR debugging, we ran through this code path.

[CLRDataAccessCreateInstance](https://github.com/dotnet/runtime/blob/6b8d056a665fa20d321ea1d49a93a37a83960663/src/coreclr/debug/daccess/daccess.cpp#L7109) -> [ClrDataAccess::Initialize](https://github.com/dotnet/runtime/blob/6b8d056a665fa20d321ea1d49a93a37a83960663/src/coreclr/debug/daccess/daccess.cpp#L5491) -> [GetDacGlobalValues](https://github.com/dotnet/runtime/blob/6b8d056a665fa20d321ea1d49a93a37a83960663/src/coreclr/debug/daccess/daccess.cpp#L7008) -> [GetDacTableAddress](https://github.com/dotnet/runtime/blob/6b8d056a665fa20d321ea1d49a93a37a83960663/src/coreclr/debug/daccess/daccess.cpp#L6986) -> [TryGetSymbol](https://github.com/dotnet/runtime/blob/6b8d056a665fa20d321ea1d49a93a37a83960663/src/coreclr/debug/dbgutil/dbgutil.cpp#L446) -> [GetMachineAndDirectoryAddress](https://github.com/dotnet/runtime/blob/6b8d056a665fa20d321ea1d49a93a37a83960663/src/coreclr/debug/dbgutil/dbgutil.cpp#L25)

This will read some data from the first 4K byte of `coreclr.dll`. In case the image is not saved in the dump, we will resort to the `PEImage.Read`, therefore we need to support reading the first 4K.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
 
         private ImmutableArray<PdbInfo> _pdbs;
         private ResourceEntry? _resources;
-        
+
         private readonly ImageDataDirectory[] _directories = new ImageDataDirectory[ImageDataDirectoryCount];
         private readonly int _sectionCount;
         private ImageSectionHeader[]? _sections;
@@ -51,7 +51,7 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
         }
 
         public PEImage(ReadVirtualStream stream, bool leaveOpen, bool isVirtual)
-            :this((Stream)stream, leaveOpen, isVirtual)
+            : this((Stream)stream, leaveOpen, isVirtual)
         {
         }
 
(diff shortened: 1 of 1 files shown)


leculver:
I'm not sure this is the right fix.  Is there a repro I could take a look at?  Or a full callstack where we would read from the wrong address?

There's two functions here.  There's `Read` which converts the address to RVA, and there's `ReadFromOffset` which reads from the file offset and doesn't do an RVA translation.  This is likely a case where we are calling the wrong read method.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -219,6 +219,9 @@ public void Dispose()
         /// <returns>The position in the stream of the data, -1 if the virtual address doesn't map to any location of the PE image.</returns>
         public int RvaToOffset(int virtualAddress)
         {
+            if (virtualAddress < 4096)
```
leculver, line 222: Is 4096 the right magic number?  Does anything guarantee that sections of memory won't be laid out lower than that?  Or that the header will be always be shorter than 4096?

leculver (commented):
I'm going to merge this as-is based on your email, but I think I need to figure out what the right behavior is.  If 4096 is a magic constant that's respected by the compiler to never put pages before that mark and that our header never goes beyond that mark, then this is right but I'm not sure what the right cutoff is otherwise.

PR #1035, opened 2022-07-31
Fix a bug related to generation bounds for regions

cshung:
@Maoni0 reported that generational aware analysis in PerfView is not working for full core dumps. After debugging for a while, I have found a few bugs related to generation detection. This PR fixes some of them, the remaining ones are in PerfView and are fixed [there](https://github.com/microsoft/perfview/pull/1674).

With this change, we have these new invariants for regions, which used to hold for segments.

```
Gen2Start == Start
Gen2Start + Gen2Length == Gen1Start
Gen1Start + Gen1Length == Gen0Start
Gen1Start + Gen0Length == End
```

The old code is not *technically* wrong, the ranges are empty anyway, but this is more convenient for PerfView to consume, and we don't know if there are other tools replicated the way PerfView did things, so might as well get it fixed.


cshung:
@leculver, can you please take a look?

leculver (approved):

PR #1047, opened 2022-11-10
Support PE relocation

cshung:
When a `PEImage` is loaded because we are reading from a `DataTarget`, this change will perform the necessary relocation so that subsequent `PEImage.Read` will produce correct values as if it is loaded in memory in the process.

This is apparently incomplete (e.g. not handling all relocation types), but it is good enough in the sense that it can successfully initialize the DAC when the `coreclr.dll` image is not saved (e.g. created by `createdump.exe --with-heap` on Windows). 

Tested on x64 only.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
@@ -4,7 +4,9 @@
 
 using System;
 using System.Buffers;
+using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
@@ -30,9 +32,10 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
         private int ImageDataDirectoryOffset => DataDirectoryOffset + ImageDataDirectoryCount * sizeof(ImageDataDirectory);
 
         private readonly ulong _imageBase;
+        private readonly ulong _loadedImageBase;
+        private readonly int[]? _relocations;
 
         private readonly Stream _stream;
-        private readonly MemoryStream _memoryStream;
         private int _offset;
         private readonly bool _leaveOpen;
         private readonly bool _isVirtual;
@@ -48,18 +51,18 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
         private object? _metadata;
         private bool _disposed;
 
-        public PEImage(FileStream stream, bool leaveOpen = false)
-            : this(stream, leaveOpen, isVirtual: false)
+        public PEImage(FileStream stream, bool leaveOpen = false, ulong loadedImageBase = 0)
+            : this(stream, leaveOpen, isVirtual: false, loadedImageBase)
         {
         }
 
         public PEImage(ReadVirtualStream stream, bool leaveOpen, bool isVirtual)
-            : this((Stream)stream, leaveOpen, isVirtual)
+            : this((Stream)stream, leaveOpen, isVirtual, 0)
         {
         }
 
         public PEImage(ReaderStream stream, bool leaveOpen, bool isVirtual)
-            : this((Stream)stream, leaveOpen, isVirtual)
+            : this((Stream)stream, leaveOpen, isVirtual, 0)
         {
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -383,13 +442,25 @@ private ResourceEntry CreateResourceRoot()
             return t;
         }
 
+        private unsafe void Write<T>(T data) where T : unmanaged
+        {
+            int size = Unsafe.SizeOf<T>();
+            byte[] buffer = new byte[size];
+            Span<byte> span = new Span<byte>(&data, size);
```
Therzok, line 449: Would wrapping `buffer` into a Span, and using `CopyTo` generate code using intrinsics here? Is there any value in doing so?

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -76,6 +79,10 @@ private PEImage(Stream stream, bool leaveOpen, bool isVirtual)
             if (!stream.CanSeek)
                 throw new ArgumentException($"{nameof(stream)} is not seekable.");
 
+            _memoryStream = new MemoryStream();
```
Therzok, line 82: If the original stream is seekable, we could probably pre-allocate with stream.Length.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
```diff
@@ -113,6 +121,56 @@ private PEImage(Stream stream, bool leaveOpen, bool isVirtual)
                 for (int i = 0; i < _directories.Length; i++)
                     if (!TryRead(out _directories[i]))
                         break;
+
+            }
+        }
+
+        public void Relocate(ulong loadedImageBase)
```
leculver, line 128: I'm not a fan of Relocate being public.  If we went the MemoryStream route, Relocate should be private and it should either be called at the end of the constructor, or we should do it on-demand when we know we have to relocate the image.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
leculver, line 82: This will blow up memory usage.  We will now have an in memory copy of every PEImage we load/read from.  This also won't work well with ReadVirtualStream, which doesn't always know the true ending of a partial image in memory, especially in the minidump case.

leculver (changes requested):
I put some comments into the implementation but at its core I don't like the approach of making a copy of the PE Image in memory and editing it for a few reasons:

1. This will blow up the memory usage of ClrMD, especially when we have tons of modules.  The underlying _stream is almost always a FileStream or ReadVirtualStream which doesn't back any of the memory.
2. Most of ClrMD is done as lazily as possible, having to fetch a whole filestream into memory for every module we load is going to be expensive perf-wise.
3. We now keep a two copies of the module, _stream and _memoryStream.

What we need to do is build a mapping table out of the relocations to translate in the other direction.  Instead of laying the PE Image out in memory as the loader would do it, we need to use the reloc table to build a mapping of "read this address, oh wait, that's over here instead".

Thank you for the pull request and work here though!  If you have time to take another swing at this to build this as a translation table I'd be super appreciative.  If not, I think your `Relocate` function gives me enough to try to implement this via mapping table.  I didn't realize there were so few reloc opcodes we actually needed to deal with.  Let me know how you want to proceed.

src/Microsoft.Diagnostics.Runtime/src/Utilities/PEImage/PEImage.cs
cshung: > I put some comments into the implementation but at its core I don't like the approach of making a copy of the PE Image in memory and editing it for a few reasons:

@leculver, I will work on the translation table approach.

cshung:
In the new iteration I pushed a change that uses the translation table approach, it looked like a lot of code, but it really isn't. At a high level, here are the changes.

- I ditched the `_memoryStream` and all related changes
- I kept all the parsing logic in the `Relocate` method and folded that into the constructor to avoid having a separate public method.
- Instead of rewriting a memory stream, the code records all the relocation intervals into the `_relocations` array. The array is sorted in a way such that it is easy for `DoRead` to use. The precise definition of it is given as a code comment block in the `DoRead` method.
- Refactor the various `Read` method so that they all eventually funnel through the `DoRead` method, it is much easier to make sure we deal with relocation and `_offset` consistently that way.
- Deleted the unused private static `Read` methods, they are just unused.
- Implement `DoRead` such that it handles the relocation. The logic is a bit complicated there. The key complexity is that we could be partially reading a relocation either at the beginning or the end (or both), so we need some logic to make sure we extend the read operation to the underlying stream as well as make sure the method leaves the stream at the logical stop point and return the logical number of bytes read.

This change also implements the relocation needed for 32-bit entries and the `TODO` block is marked in the constructor is never actually hit while reading a minidump captured with `createdump.exe --withheap` for the latest runtime.

leculver (approved):
Looks great!  I will try to get a build out today or Monday.

Issue #962, opened 2021-11-08
Support NativeAOT for writing WinDbg extensions


kevingosse:
[NativeAOT](https://github.com/dotnet/runtimelab/tree/8e81d3a5bfd7639a197b51a1f65fcbba129d3b5f/samples/NativeLibrary) is slowly becoming mature, and is planned to be released with .NET 7. I believe this is a really convenient way to write WinDbg extensions, as it supports native exports without having to use ildasm-based hacks such as [UnmanagedExports](https://www.nuget.org/packages/UnmanagedExports).

I tried writing a proof of concept, and it works pretty well with a few changes to ClrMD: https://github.com/kevingosse/clrmd/tree/nativeaot

Mainly, adding `UnmanagedFunctionPointerAttribute` to some delegates, and exposing types that are useful to interact with WinDbg (for instance `DebugControl` to use `ControlledOutput`).

Is there any plan to actively support NativeAOT? If so, I'd be happy to contribute.

leculver:
This is something that in general I would like to move to and support as well as possible.  I think my north star here is: "Building WinDbg (or lldb, or other debugger) extensions with ClrMD should be very easy and well supported."

I struggle a bit with how much of this to put directly into ClrMD and how much to write as a public domain sample.  I've specifically been trying to slim ClrMD down into just the managed debugging parts and not be an "Everything, all in one" library with things like Elf file reading, PEImage reading, PDB reading, etc.  It's just me that maintains this library and I have a lot of other responsibilities at microsoft too...so I haven't scaled well to keep all of these side pieces up to date.

So, it makes me nervous adding direct DbgEng hooks back into the library, marked public that I personally don't regularly use...I've found it's a quick way to get code rot.

With that said, if this approach leads to very easy to build DbgEng plugins that may be worth the cost of maintenance.

I'll take a close look at nativeaot windbg plugins in the next few weeks (including your linked tree, thanks for that!).  If it makes sense to add to the library I will, and if it doesn't I think at minimum it's reasonable for me to provide a public domain sample of template code that gets someone directly up and running with a dbgeng plugin using nativeaot instead of ildasm based things.

Thanks!

kevingosse:
[...]

PR #1051, opened 2022-12-07
NativeAOT support

leculver:
I'm working on a lot of NativeAOT related changes and improvements.  This will include both making ClrMD compatible with NativeAOT, as well as updating the WinDbg plugin sample to be .Net Core using NativeAOT.

Resolves: https://github.com/microsoft/clrmd/issues/962


leculver:
@kevingosse Do you have any feedback about this pull request and the direction I'm going here?

This is a big PR, but I think looking at the [readme for the Utilities project](https://github.com/leculver/clrmd/tree/aot/src/Microsoft.Diagnostics.Runtime.Utilities), the [DbgEng Plugin Sample](https://github.com/leculver/clrmd/tree/aot/src/Samples/DbgEngExtension), and the [DbgEng Standalone App Sample](https://github.com/leculver/clrmd/blob/aot/src/Samples/DbgEngStandalone/Program.cs) fully captures what I'm trying to do here.  Reviewing those two files and one directory gives a good flavor of where I'm going here.

The # 1 concern I have here is that I haven't taken a look at the LLDB plugin/programming model.  I'd really like plugins to be shareable between LLDB and DbgEng.  For example, my [MHeap](https://github.com/leculver/clrmd/blob/aot/src/Samples/DbgEngExtension/MHeap.cs) plugin example does _nothing_ that is DbgEng specific.  I'd really like that code to be sharable between an lldb and dbgeng plugin...   Yet the design I've landed on wouldn't be suitable for that.

I'm planning to go spend another week or two digging deeply into LLDB and their plugin model to refactor/revise the design and implementation of plugins here.  Which is why I'm reaching out now and asking for feedback.

With all of that said.  This change updates ClrMD to be NativeAOT compliant.  You don't have to use my Microsoft.Diagnostics.Runtime.Utilities project at all.  Everyone is free to write their own wrappers and not use mine, which is why I pushed all of this code into another library.  (Note Utilities will be released in full from here on, it won't be marked prerelease...but it also won't contain a random set of code.  It's focused towards LLDB and WinDbg plugins, and maybe some really specific things like that which do not fit into ClrMD itself.)

@mikem8361 FYI.  Any feedback welcome.

leculver:
Having dug further into the lldb API, it looks like they don't offer any kind of C or easily wrappable interface definition.  It's all C++ API, which I don't think is a good idea to try to mimic here.  I think any lldb plugin is going to have to start with a C++ wrapper of some sort (CLI or otherwise), at which point we are already out of the business of putting this code into a single C# utilities library.  I don't think I'm headed down that road.

src/Microsoft.Diagnostics.Runtime.Utilities/DbgEng/DbgEngCommand.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        public (int HResult, string Output) RunCommandWithOutput(string command, DEBUG_OUTPUT mask = DEBUG_OUTPUT.NORMAL)
+        {
+            StringBuilder result = new();
+            int hr;
+            using (DbgEngOutputHolder outputHolder = new(DebugClient, mask))
+            {
+                outputHolder.OutputReceived += (text, flags) => result.Append(text);
+                hr = DebugControl.Execute(DEBUG_OUTCTL.THIS_CLIENT, "!address", DEBUG_EXECUTE.DEFAULT);
```
kevingosse, line 52: Shouldn't this run `command` instead of `!address`?

kevingosse:
> @kevingosse Do you have any feedback about this pull request and the direction I'm going here?

Unfortunately I don't think I'll have time to look into details in the coming days, but I've checked the `DbgEngCommand` API and it looks good.

> Having dug further into the lldb API, it looks like they don't offer any kind of C or easily wrappable interface definition. It's all C++ API, which I don't think is a good idea to try to mimic here. I think any lldb plugin is going to have to start with a C++ wrapper of some sort (CLI or otherwise), at which point we are already out of the business of putting this code into a single C# utilities library. I don't think I'm headed down that road.

Yeah, I've played with the LLDB API quite a bit and came to the same conclusion. It's technically possible to write a plugin entirely in C# (https://github.com/kevingosse/UpdateConference2022/tree/master/LldbExtension) but I don't think it's reasonable. A C++ wrapper is definitely the way to go.

src/Microsoft.Diagnostics.Runtime.Utilities/DbgEng/DbgEngCommand.cs
leculver: Yes, thank you.  =)

PR #1052, opened 2022-12-10
Fix MachO core dump generated on the Ventura (13.0.1)

mikem8361:
On Ventura, the dylinker header isn't on a segment boundary any more. It is usually 8K into segment. There were also problems with incorrectly calculating the LoadBias and the TryLookupSymbol result.

See: https://developer.apple.com/forums/thread/720064
Runtime issue: https://github.com/dotnet/runtime/issues/79355

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOCoreDump.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOCoreDump.cs
@@ -246,10 +246,17 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
             {
                 foreach (MachOSegment seg in _segments)
                 {
-                    MachHeader64 header = ReadMemory<MachHeader64>(seg.Address);
-                    if (header.Magic == MachHeader64.Magic64 && header.FileType == MachOFileType.Dylinker)
+                    for (ulong offset = 0; offset < seg.FileSize; offset += 0x1000)
                     {
-                        _dylinker = new MachOModule(this, seg.Address, "dylinker");
+                        MachHeader64 header = ReadMemory<MachHeader64>(seg.Address + offset);
+                        if (header.Magic == MachHeader64.Magic64 && header.FileType == MachOFileType.Dylinker)
+                        {
+                            _dylinker = new MachOModule(this, seg.Address + offset, "dylinker");
+                            break;
+                        }
+                    }
+                    if (_dylinker != null)
+                    {
                         break;
                     }
                 }
--- a/src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOCoreDump.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOCoreDump.cs
@@ -244,22 +244,7 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
 
             if (_dylinker == null)
             {
-                foreach (MachOSegment seg in _segments)
-                {
-                    for (ulong offset = 0; offset < seg.FileSize; offset += 0x1000)
-                    {
-                        MachHeader64 header = ReadMemory<MachHeader64>(seg.Address + offset);
-                        if (header.Magic == MachHeader64.Magic64 && header.FileType == MachOFileType.Dylinker)
-                        {
-                            _dylinker = new MachOModule(this, seg.Address + offset, "dylinker");
-                            break;
-                        }
-                    }
-                    if (_dylinker != null)
-                    {
-                        break;
-                    }
-                }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOCoreDump.cs
```diff
@@ -246,10 +246,17 @@ internal string ReadAscii(ulong address)
             {
                 foreach (MachOSegment seg in _segments)
                 {
-                    MachHeader64 header = ReadMemory<MachHeader64>(seg.Address);
-                    if (header.Magic == MachHeader64.Magic64 && header.FileType == MachOFileType.Dylinker)
+                    for (ulong offset = 0; offset < seg.FileSize; offset += 0x1000)
```
leculver, line 249: Does this mean we will walk every 4k segment of every region of memory in a dump?

So in a 4 gigabyte dump file we'll need to make 1,000,000 reads looking for the dylinker?

mikem8361: Yes, could get that bad but in practice, the dylinker header is in the first few segments.  I currently don't have a better way of doing this. I'm going to look at the latest lldb MachO code (it may take a while because it is hard to understand).  

One other possible is to modify createdump to put the dylinker header on a segment boundary but it was way to ugly and messy for the 6.0/7.0 service releases that are in progress to fix generating dumps on Ventura.

mikem8361: It looks like lldb does the same thing:

```
const size_t num_core_aranges = m_core_aranges.GetSize();
  for (size_t i = 0; i < num_core_aranges; ++i) {
    const VMRangeToFileOffset::Entry *entry = m_core_aranges.GetEntryAtIndex(i);
    lldb::addr_t section_vm_addr_start = entry->GetRangeBase();
    lldb::addr_t section_vm_addr_end = entry->GetRangeEnd();
    for (lldb::addr_t section_vm_addr = section_vm_addr_start;
         section_vm_addr < section_vm_addr_end; section_vm_addr += 0x1000) {
      addr_t dyld, kernel;
      if (CheckAddressForDyldOrKernel(section_vm_addr, dyld, kernel)) {
        if (dyld != LLDB_INVALID_ADDRESS)
          dylds_found.push_back(dyld);
        if (kernel != LLDB_INVALID_ADDRESS)
          kernels_found.push_back(dyld);
      }
    }
  }
```

mikem8361: We might be able to limit the check to rx segments but I would have to confirm with @hoyosjs testing on Ventura that the header was found in a rx segment. Since lldb doesn't filter the segments, it might be risky for our dump headers to do the same.

leculver: What if we check the first N sections of each segment, since it's almost always in the first few?  Something like this?
``` C#
if (_dylinker == null)
{
    _dylinker = FindDylinker(firstPass: true) ?? FindDylinker(firstPass: false);
}

private static MachOModule? FindDylinker(bool firstPass)
{
    const uint skip = 0x1000;
    const uint firstPassAttemptCount = 8;
    foreach (MachOSegment seg in _segments)
    {
        ulong end = seg.FileSize;
        ulong start = 0;

        if (firstPass)
            end = skip * firstPassAttemptCount;
        else
            start = skip * firstPassAttemptCount;

        for (ulong offset = start; offset < end; offset += skip)
        {
            MachHeader64 header = ReadMemory<MachHeader64>(seg.Address + offset);
            if (header.Magic == MachHeader64.Magic64 && header.FileType == MachOFileType.Dylinker)
            {
                return new MachOModule(this, seg.Address + offset, "dylinker");
            }
        }
    }
}
```

Note:  **I didn't test this code**.  I assume N=8 is reasonable?

This way the normal case is found quickly, but a bad dump (with no dylinker) still checks everything.

leculver:
If we can find a way to mitigate the impact of this for the normal case scenario I think that'd be best.  I have a suggestion above that should reduce the impact.

src/Microsoft.Diagnostics.Runtime/src/MacOS/MachOCoreDump.cs
mikem8361: I'll work on this today.

leculver:
LGTM!

I'm working on a sizable change to support NativeAOT.  I probably will not release this as a build until after that's completed, probably early next week.

PR #1057, opened 2022-12-20
Fix parsing /proc/PID/maps entries with (deleted) tag

janvorli:
The /proc/PID/maps can contain entries for file mapping / shared memory mapping that contain `(deleted)` 
after the filename. The W^X double mapping use shared memory mapping backed by the pagefile where 
the entries look like
`7fb6d722b000-7fb6d722c000 rw-s 0043c000 00:01 47618 /memfd:doublemapper (deleted)`
ClrMD was rejecting such entries as invalid in `LinuxLiveDataReader.LoadMemoryMaps.`

This change fixes it.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
@@ -309,6 +309,10 @@ namespace Microsoft.Diagnostics.Runtime.Utilities
                 {
                     path = parts[5].StartsWith("[") ? string.Empty : parts[5];
                 }
+                else if (parts.Length == 7 && parts[6] == "(deleted)")
+                {
+                    path = string.Empty;
+                }
                 else
                 {
                     DebugOnly.Fail("Unknown data format");


janvorli:
cc: @adamsitnik

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
```diff
@@ -309,6 +309,10 @@ private ImmutableArray<MemoryMapEntry>.Builder LoadMemoryMaps()
                 {
                     path = parts[5].StartsWith("[") ? string.Empty : parts[5];
                 }
+                else if (parts.Length == 7 && parts[6] == "(deleted)")
```
adamsitnik, line 312: the `string.Split` used above sets the count as 6:

https://github.com/microsoft/clrmd/blob/c936dd96eb996b4489995c46df7fd7ad7c1da3e2/src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs#L303

so `parts.Length` will never be more than 6. We need to set it to `7`

adamsitnik (commented):
@janvorli big thanks for investigating the issue and fixing it!

src/Microsoft.Diagnostics.Runtime/src/Linux/LinuxLiveDataReader.cs
janvorli: Hmm, actually, then the ClrMD was not rejecting these entries, as it would only parse the first 6 ones. @leculver does the ClrMD use the extracted path for any kind of access to the underlying files? If not, then we could close this PR. If yes, then it seems it would still make sense to report the path as empty, as the physical underlying file doesn't exist in the filesystem.

leculver: There's a lot of machinery in ClrMD.  Without debugging it, I believe we _do_ try to use that path, but all of that code should be around a try/catch (or a "File.Exists").  I'm not sure what the underlying issue is that you are trying to fix so I'm really not sure whether this would affect it or not.

janvorli: The point was that the issue I was attempting to fix by this change was in fact caused by a different problem. It originally seemed that the issue happens because we are unable to parse some of the memory blocks here, but after Adam's comment, I've realized that we were still parsing them ok, as the 7-th space separated portion of the line was never parsed. 
What I was trying to say was that if there is the `(deleted)` after a filename or the `/memfd:doublemapper`, which represents mapping that's backed by memory and not a file, then the filename portion of the maps line doesn't represent any existing file in the filesystem. So I was thinking that my change (after fixing it based on Adam's feedback) might still make sense in general, as it would clear the path instead of returning a path of a non-existent file.

leculver: I had some time to dig into the code this morning.  I think the right approach here is to simply simply remove the ` (deleted)` string from the filename.

We eventually do expose this data out to the user via `DataTarget.EnumerateModules().FileName`.  We definitely _don't_ want to write out `/foo/bar (deleted)` as the filename, but ClrMD works fine if the file isn't on disk there anymore.  In fact, that's the "normal" case for dump debugging.

It's true that the user could end up enumerating modules in the process and finding a `.FileName` that doesn't exist on disk, but that's a standard problem of dealing with files...someone might come along and delete them between the time at which someone calls `File.Exists` and `File.Open`.

As for `/memfd:doublemapper`, I'm not sure what those are.  Are the files mapped this way ever PE or ELF images?  If so we can still enumerate them as normal and again erase `/memfd:doublemapper` off of it.  If the memory mapped this way is _never_ elf or PE images we can just omit them from the enumeration because our `ModuleInfo` class doesn't make sense to expose them if they aren't.

What do you think?

janvorli: The `/memfd:doublemapper` are mappings created using file descriptor obtained from `memfd_create` - basically anonymous files. They represent executable memory allocated for jitted code, various stubs etc. when W^X is enabled (which is by default in .NET 7, can be enabled in .NET 6).
Based on what you've said, I think that ignoring the `(deleted)` and reporting filenames without it makes sense. And in fact, it is what happens without my change. The `(deleted)` is separated from the filename by space, so with the existing code, it just works. So I am going to close this PR as not needed.

Btw, @adamsitnik mentioned today that the current way of parsing the maps lines is broken for the case of filenames containing spaces. That might be worth fixing eventually.

PR #1070, opened 2023-02-14
Fix NuGet.config

leculver:


leculver:
@alexperovich Awesome!  Seems to be working now!

(I still need to fix this PR, but at least it's working!)

PR #1072, opened 2023-02-19
Add a way to enumerate native heaps

leculver:
Added ClrRuntime.EnumerateClrNativeHeaps.  This function will enumerate the native heaps/memory that CLR has allocated.  This is intended to be used in functions similar to SOS's !eeheap.

Added ISOSDacInterface13 to implement this and related functions.

Worked around an issue in .Net 7.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
@@ -763,7 +763,16 @@ namespace Microsoft.Diagnostics.Runtime.Builders
 
                 hr = _sos13.TraverseLoaderHeap(address, kind, callback);
             }
-            else if (_clrInfo.Flavor == ClrFlavor.Desktop || (_clrInfo.Flavor == ClrFlavor.Core && _clrInfo.Version.Major != 7))
+            else if (_clrInfo.Flavor == ClrFlavor.Core && _clrInfo.Version.Major == 7)
+            {
+                // See note below, .Net 7 inverts the logic that everything else uses.
+
+                if (kind == SOSDac13.LoaderHeapKind.LoaderHeapKindNormal)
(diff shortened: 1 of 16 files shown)


src/Microsoft.Diagnostics.Runtime/src/Builders/RuntimeBuilder.cs
```diff
@@ -655,6 +657,140 @@ private IEnumerable<ClrHandle> EnumerateHandleTable(ClrRuntime runtime, HandleDa
...
+            List<ClrNativeHeapInfo> heaps = new();
+
+            // Enumerate JIT code heaps.
+            foreach (JitManagerInfo jitMgr in _sos.GetJitManagers())
+            {
+                foreach (var mem in _sos.GetCodeHeapList(jitMgr.Address))
+                {
+                    if (mem.Type == CodeHeapType.Loader)
+                    {
+                        if (TraverseLoaderHeap(mem.Address, SOSDac13.LoaderHeapKind.LoaderHeapKindExplicitControl, (address, size, _) => heaps.Add(new ClrNativeHeapInfo(address, SanitizeSize(size), NativeHeapKind.LoaderCodeHeap))))
```
Therzok, line 671: Wouldn't it be more efficient to create an enumerable out of the traversing code? That would avoid eagerly evaluating the data and holding onto it for the entirety of the enumeration.

leculver:
Can you give me an example code pattern you are referring to?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

leculver:
Oh you mean making the underlying dac function be an enumerable instead of a callback?  Yeah, in general that would be ideal.  There's a few reasons why I didn't though:

1.  We still have to maintain the original function, so it's better to share the implementation at all of the entry points.
2. There's not a great iterator pattern in the dac to crib from.
3. This is a pretty rarely used function, so it wasn't worth agonizing over.  =)

PR #1077, opened 2023-02-25
Use the RuntimeVersion in the new RuntimeInfo 2 for single file apps

mikem8361:
Checks for the new version 2 of the RuntimeInfo added in .NET 8 containing the runtime version. 

I'm not sure if this is the best approach but I didn't want to read the whole new ClrRuntimeInfo2 without checking the version just in case on < 8 runtimes it is on a page boundary.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
+++ b/src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
@@ -97,8 +97,12 @@ namespace Microsoft.Diagnostics.Runtime
                     }
                 }
 
-                // We don't actually know the version of single file CLR since the version of the module would be the user's app version
                 Version = new Version();
+                if (info.IsClrRuntimeInfo2)
+                {
+                    ClrRuntimeInfo2 info2 = DataTarget.DataReader.Read<ClrRuntimeInfo2>(runtimeInfo);
+                    Version = info2.RuntimeVersion;
+                }
             }
             else
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Runtime/src/DataTargets/ClrInfo.cs
```diff
@@ -97,8 +97,12 @@ internal ClrInfo(DataTarget dt, ClrFlavor flavor, ModuleInfo module, ulong runti
                     }
                 }
 
-                // We don't actually know the version of single file CLR since the version of the module would be the user's app version
                 Version = new Version();
+                if (info.IsClrRuntimeInfo2)
```
leculver, line 101: Instead of reading twice, how about adding something like this to `ClrRuntimeInfo`?  (I did not test this code, please run it through your change...especially making sure the alignment is right.)

```C#

        public static unsafe bool TryReadClrRuntimeInfo(IDataReader reader, ulong address, out ClrRuntimeInfo header, out Version? version)
        {
            header = default;
            version = null;

            int size = sizeof(ClrRuntimeInfo);
            fixed (ClrRuntimeInfo* pHeader = &header)
                if (reader.Read(address, new Span<byte>(pHeader, size)) != size || !header.IsValid)
                    return false;

            if (header._version >= 2)
            {
                Span<int> intVersion = stackalloc int[4];
                Span<byte> buffer = MemoryMarshal.Cast<int, byte>(intVersion);

                if (reader.Read(address + (uint)size, buffer) == buffer.Length)
                    version = new Version(intVersion[0], intVersion[1], intVersion[2], intVersion[3]);
            }

            return true;
        }
```

leculver (changes requested):
How about something like the suggested code below?  (Then replace the original Read<ClrRuntimeInfo> with this.)

I didn't test that suggestion, so if it doesn't work as expected I'm happy to merge your original change.

PR #1097, opened 2023-03-09
Fix NotImplementedExceptions on common types.

nschuessler:
Fixes some easy NotImplementedExceptions.


nschuessler:
@leculver Here is a small fix.

nschuessler:
@leculver Here is a small fix.

leculver:
Much appreciated!  Sorry I missed these in the last round of changes.

PR #1108, opened 2023-03-13
Add style rules and analyzers + fixes

hoyosjs:
- Add base infrasctructure for analyzers and style
- Standardize on modern .NET Foundation license
- Don't initialize to defaults.
- Always add access modifiers and order them
- Fix file endings
- Order using statemenets and remove unnecessary ones
- Use explicit type and prefer target-type new
- Discard unused outs
- Use default when applicable
- Fix docs
- Make methods static if no state is used
- Seal internal/private underived classes
- Disable warnings that are false positives
- Fix trailing whitespace
- REVIEW: Don't use StringBuilder in PInvoke
- REVIEW: Implement IEquatable if overrides equals.
- Use ordinal comparison for strings
- Use static initializers over cctor
- REVIEW: Delete unused constants in sigpaser
- Remove unnecessary init
- Use ConfigureAwait within library boundaries
- Remove unused sync field in DataTarget
- REVIEW: make IDebugClientVtable readonly
- Simplify statements with pattern matching


leculver:
This is awesome, thank you for doing this!

PR #1126, opened 2023-03-27
Ensure public interfaces have all the public APIs.

nschuessler:
This should ensure that the interfaces for `IClrHeap`, `IClrSubHeap`, and `IClrSegment` match the public APIs of the corresponding classes.  Some new interfaces were required.


leculver (approved):
LGTM

PR #1127, opened 2023-03-31
FinalizerRoots and Generations

leculver:
- Fix an issue with recent runtimes not properly reporting Finalizer roots.
- Change ClrSegment.GetGeneration to return an enum of what kind of generation the object belongs to.
- Re-order `GCSegmentKind` to match `Generation` ordering.


leculver:
@mikem8361 FYI, I think this change might be incompatible with dotnet/diagnostics/main.  I'll use the auto-generated PR for ClrMD dependency update to fix it on Monday.

mikem8361:
Sounds fine with me.

PR #1139, opened 2023-04-12
Some interface clean up changes. Add back old IsFinalizer and IsGc apis.

nschuessler:
- Some new public apis not on interfaces.
- Add `IsFinalizer` and `IsGc` back to `ClrThread` and `IClrThread`


leculver:
Awesome, thank you!  I had not gotten to this yet.

I'm going to change HillClimbingLogEntry to be fully constructible and not need an interface, otherwise LGTM.

leculver:
I'll push updates to this PR.

src/Microsoft.Diagnostics.Runtime/ClrThread.cs
```diff
@@ -30,6 +30,8 @@ internal ClrThread(IClrThreadHelpers helpers, ClrRuntime runtime, ClrAppDomain?
             LockCount = data.LockCount;
             State = (ClrThreadState)data.State;
             _exceptionHandle = data.LastThrownObjectHandle;
+            this.IsFinalizer = isFinalizer;
+            this.IsGc = isGc;
```
leculver, line 34: One nitpick, we don't use `this.` in any place here.  All members/properties have to be unambiguous.  Members start with `_` statics with `s_`, locals with lowercase, properties with uppercase.

PR #1142, opened 2023-04-15
Remove ObjectSet

leculver:
Use HashSet<ulong> instead.


leculver:
I cut ObjectSet for a few reasons:

1.  It has bad performance characteristics.
2.  I don't use it anywhere in ClrMD.
3. I think folks could probably come up with better data structure if they are really using ClrMD under tight memory constraints.
4. I ran out of time to improve it, and if I'm going to cut it I can really only do so in a major release...which won't happen again for the foreseeable future.

The longer explanation:

I originally wrote ObjectSet ~10 years ago when HashSet<ulong> would throw exceptions when you tried to add 20+ million objects to it.  We wouldn't be OOM, but we'd hit other implementation limitations.  Those no longer apply, in both desktop CLR and .Net Core we can add objects to a set until we run out of space.

Recently, I started updating SOS to use ClrMD instead of its own logic.  (We have a problem of duplicating complicating diagnostic logic all over the place.)  In doing so, I found that ObjectSet was actually a huge performance bottleneck when performing !gcroot.  I always assumed that ObjectSet was trading space for performance, but this was _particularly_ bad.

I'm winding down a long sprint of work (https://github.com/dotnet/diagnostics/issues/3736) and I only have a short period of time to complete it before I go on leave for 3 months and didn't think I'd have a ton of time to go investigate better performance options.  I can only really remove something like ObjectSet in a major release, so I chose to cut it now rather than leave it in the library, unused, and be unable to cut it later.

Lastly, since I don't use it anywhere and I try to keep ClrMD slimmed down to just what's needed, it was a something I removed to not have to continue maintaining.

If there is significant pushback I can see if I can find time to re-add it in an optimized way.  I'm not sure I have a ton of time between now and when I leave though.

ayende:
The primary reason that I'm interested is the comment on `HashSet<ulong>`. I needed to use that recently, remembered the `ObjectSet` comment and tracked it down.

Thanks.

Issue #1079, opened 2023-03-01
ClrMD is moving to dotnet/diagnostics


leculver:
The ClrMD repo is moving to be part of https://github.com/dotnet/diagnostics.  I plan to complete the move some time in April.

This move is being made purely to reduce my development overhead.  I spend a LOT of time behind the scenes just maintaining this repo and fixing weird issues with our CI system.  Moving the code into dotnet/diagnostics will let me spend more time actually working on ClrMD instead of just keeping the lights on.

If you have any questions about this, please let me know.  Here's some quick answers to things that may come up:

1.  The license will not change (still MIT).
2. I will still be working on the project.  If anything it gives me more time to work in diagnostics.
3. I will be manually porting any actionable issues from this repo to there.
4. New bugs and pull requests go to the diagnostics repo.

Please let me know if anyone has any questions.  Thanks!

leculver:
We are abandoning this effort.  ClrMD will remain here.

PR #1149, opened 2023-04-24
Add SUPPORT.md

paulomorgado:
Add SUPPORT.md file with support policy.


leculver:
ClrMD is moving to dotnet/diagnostics, see #1079.  We will be following their support policy and release schedule, so I don't want to merge this PR just for the few more weeks this repo is in operation.

PR #1150, opened 2023-05-02
Add TryReadStringField api similar to TryReadObjectField

nschuessler:
If you don't know the field name, then the `Try` version won't cause `ArgumentExceptions`.


leculver (approved):
LGTM!

PR #1156, opened 2023-05-11
Fix module size return

leculver:


src/Microsoft.Diagnostics.Runtime/ClrModule.cs
```diff
@@ -253,7 +253,7 @@ private ulong GetSize()
         {
             ulong size = ExtendedData.Size;
             if (size != 0)
-                return 0;
+                return size;
```
mikem8361, line 256: Wouldn't be cleaner to say `if (size == 0)` the whole block and return only once?

PR #1168, opened 2023-07-04
Prevent exception when reading roots from the bad stack

catstrike:
I have a memory dump in which the following operation finds 2 frames:
```
ClrStackFrame? frame = stack.SingleOrDefault(f => f.StackPointer == stackRef.Source || f.StackPointer == stackRef.StackPointer && f.InstructionPointer == stackRef.Source);
```

The dump is apparently special but contains useful information. I believe that would be nice to still be able to list the roots from the stack.


leculver:
Sorry for the delay, I have been on leave.

The change here is reasonable, as we want to make sure that we don't simply throw an exception and give up, so I'll merge this as-is.

The interesting thing though is that you still _shouldn't_ be getting that exception unless there's a bug somewhere else in our diagnostic stack.  Whenever we unwind a stack, you shouldn't get the same IP and SP back for two different frames...  If you do, it means that the stack unwind isn't making progress, and possibly is infinitely looping.

Let me know if the scenario you encountered an exception here is reproduceable or common.

Thanks!

leculver (approved):

Issue #897, opened 2021-02-26
DacNameParser.Parse(name) throws System.IndexOutOfRangeException


catstrike:
I'm enumerating all the types in a dump with ClrMD 2.0
For one of the types, DAC returns the following name:
```StructureMap.Pipeline.ExpressedInstance`3[[StructureMap.Pipeline.LambdaInstance`2, EnforceHttpsModule],T,TPluginType]```

DacNameParser can't parse it and throws System.IndexOutOfRangeException.

```
Unhandled Exception: System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at System.String.get_Chars(Int32 index)
   at Microsoft.Diagnostics.Runtime.Builders.DACNameParser.DetermineNextStateAndPos(String name, Int32 curPos)in C:\Developer\projects\clrmd2\src\Microsoft.Diagnostics.Runtime\src\Common\DACNameParser.cs:line 630
   at Microsoft.Diagnostics.Runtime.Builders.DACNameParser.Parse(String name) in C:\Developer\projects\clrmd2\src\Microsoft.Diagnostics.Runtime\src\Common\DACNameParser.cs:line 165
   at Microsoft.Diagnostics.Runtime.Builders.RuntimeBuilder.Microsoft.Diagnostics.Runtime.Implementation.ITypeHelpers.GetTypeName(UInt64 mt, String& name) in C:\Developer\projects\clrmd2\src\Microsoft.Diagnostics.Runtime\src\Builders\RuntimeBuilder.cs:line 1465
   at Microsoft.Diagnostics.Runtime.Implementation.ClrmdType.get_Name() in C:\Developer\projects\clrmd2\src\Microsoft.Diagnostics.Runtime\src\Implementation\ClrmdType.cs:line 61
   at Microsoft.Diagnostics.Runtime.Implementation.ClrmdType.DebugOnlyLoadLazyValues() in C:\Developer\projects\clrmd2\src\Microsoft.Diagnostics.Runtime\src\Implementation\ClrmdType.cs:line 120
   at Microsoft.Diagnostics.Runtime.Implementation.ClrmdType..ctor(ClrHeap heap, ClrType baseType, ClrModule module, ITypeData data, String name) in C:\Developer\projects\clrmd2\src\Microsoft.Diagnostics.Runtime\src\Implementation\ClrmdType.cs:line 114
[...]

PR #1172, opened 2023-07-26
Fix bugs in DAC names that either contain non-closed generics or non-assembly qualified names

ryanmolden:
Fix DAC name parser to deal with non-closed generic types and non-assembly qualified names (see bug #897).

Add lots of comments to parser as its complex due to having to deal with tragedies like the type name in the ParseGenericWithComplexGenericArgs2 test.

Replace uses of char literals in numerous places with their named constants I was using elsewhere.

Correct numerous comment spelling errors (thanks VS spell checker).

Add numerous tests both for the case in 897, some variants of it, and another case I encountered internally.

Add support for non-traditional generics as produced by FSharp and various obfuscators.


ryanmolden:
Actually, wait a bit on this, found a few more cases, gonna run some more dumps locally and make sure I can't find any more.

leculver (approved):
LGTM, I made a minor fix to reduce allocations.
This worked well on the test dumps I threw at it, and passes all tests.

PR #1175, opened 2023-08-15
Update the ThreadPool command to handle both Portable and Windows thread pool

eduardo-vp:


leculver:
Hey!

I'm in the process of migrating the codebase here over to dotnet/diagnostics.  The code will remain the same, but the relative path of the files and location of the repo will change.  This will make migrating your change annoying (but not impossible).

Are you ready to wrap up this change?  It's no problem if you aren't yet ready.

If you need more time, you may need to generate a patch for the diff (maybe try `git format-patch`), edit the patch to apply to the new paths in the dotnet/diagnostics repo.  I'm happy to help if that's what we need to do.

Thanks!

leculver (approved):
LGTM, I have two minor suggestions.

kouvel (approved):

PR #1199, opened 2023-10-04
Add ModuleInfo.TryCreateResourceRoot

mikem8361:
This allows the SOS infra to leverage the clrmd internal PEImage's resource code for use in its IDataReader impl.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ModuleInfo.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ModuleInfo.cs
@@ -82,6 +82,24 @@ namespace Microsoft.Diagnostics.Runtime
             return result;
         }
 
+        /// <summary>
+        /// Attempts to create a root resource node for the PE module.
+        /// </summary>
+        /// <param name="reader">The data reader to create this module from.</param>
+        /// <param name="imageBase">The base address of this module.</param>
+        /// <param name="imageSize">The module image size.</param>
+        /// <param name="isFileLayout">if true, the PE has file layout, false loaded layout</param>
+        /// <returns>IResourceNode instance</returns>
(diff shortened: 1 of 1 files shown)


src/Microsoft.Diagnostics.Runtime/ModuleInfo.cs
```diff
@@ -82,6 +82,24 @@ public abstract class ModuleInfo
...
 
+        /// <summary>
+        /// Attempts to create a root resource node for the PE module.
+        /// </summary>
+        /// <param name="reader">The data reader to create this module from.</param>
+        /// <param name="imageBase">The base address of this module.</param>
+        /// <param name="imageSize">The module image size.</param>
+        /// <param name="isFileLayout">if true, the PE has file layout, false loaded layout</param>
+        /// <returns>IResourceNode instance</returns>
+        public static IResourceNode? TryCreateResourceRoot(IDataReader reader, ulong imageBase, ulong imageSize, bool isFileLayout)
```
leculver, line 93: Does this need to be static?  Or should it be instance?  Are you intending to use this to read resources on things that aren't strictly wrapped by ModuleInfo?

mikem8361: I want it to be static because I want to read resources from modules that don't have a ModuleInfo (like when SOS infra is hosted under windbg or lldb).

mikem8361: Here is an example of how I'm going to use it: https://github.com/mikem8361/diagnostics/blob/42b6643ed799c1259aedd023c74380b2c5fa3abc/src/Microsoft.Diagnostics.DebugServices.Implementation/DataReader.cs#L208

leculver: Sounds good.

PR #1203, opened 2023-10-12
Show IO completion thread pool info when the info is available and the portable thread pool is being used

kouvel:
Updated to fill in the IO completion thread pool info when it exists.


leculver (approved):
LGTM.  Feel free to merge when you are ready.

PR #1206, opened 2023-10-16
Get method's hot and cold region info based on the MethodDesc's native address

PhDuck:
The current implementation uses the CodeHeader based on the slot address, which does not provide the hot and cold region info for the method.

Instead use the MethodDesc's native address.


leculver:
LGTM, thanks!

PR #1208, opened 2023-10-24
[RISC-V] Minor fixes.

viewizard:
Minor fixes and improvements for RISC-V related code.

CC @clamp03 @wscho77 @HJLeee @JongHeonChoi @t-mustafin @gbalykov

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/DataReaders/Core/CoreDumpReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DataReaders/Core/CoreDumpReader.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Diagnostics.Runtime
                 ElfMachine.EM_386 => (4, Architecture.X86),
                 ElfMachine.EM_AARCH64 => (8, Architecture.Arm64),
                 ElfMachine.EM_ARM => (4, Architecture.Arm),
-                ElfMachine.EM_RISCV => _core.ElfFile.Header is ElfHeader64 ?
+                ElfMachine.EM_RISCV => _core.ElfFile.Header.Is64Bit ?
                                             (8, (Architecture)9 /* Architecture.RiscV64 */) :
                                             throw new NotImplementedException($"Support for RISC-V 32bit not yet implemented."),
                 _ => throw new NotImplementedException($"Support for {architecture} not yet implemented."),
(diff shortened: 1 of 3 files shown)


gbalykov (approved):

src/Microsoft.Diagnostics.Runtime/DataReaders/Core/CoreDumpReader.cs
```diff
@@ -32,7 +32,7 @@ public CoredumpReader(string path, Stream stream, bool leaveOpen)
                 ElfMachine.EM_386 => (4, Architecture.X86),
                 ElfMachine.EM_AARCH64 => (8, Architecture.Arm64),
                 ElfMachine.EM_ARM => (4, Architecture.Arm),
-                ElfMachine.EM_RISCV => _core.ElfFile.Header is ElfHeader64 ?
+                ElfMachine.EM_RISCV => _core.ElfFile.Header.Is64Bit ?
```
HJLeee, line 35: Does this change make a difference?

leculver: (I leave that to you all for the real answer, but I'm always happy to make a check into a property request instead of a cast-check in clr.  Much faster.)

viewizard: @HJLeee this is absolutely same logic, but faster, as @leculver also noticed.
`Is64Bit` set in ElfHeader64/ElfHeader32 for proper value:
https://github.com/microsoft/clrmd/blob/8ee682ab6a50704b4f398cf683032f92963b4d71/src/Microsoft.Diagnostics.Runtime/Linux/Structs/ElfHeader32.cs#L27
https://github.com/microsoft/clrmd/blob/8ee682ab6a50704b4f398cf683032f92963b4d71/src/Microsoft.Diagnostics.Runtime/Linux/Structs/ElfHeader64.cs#L27
no reason for cast-check all the time. Just missed this property at first time changes. -_-

HJLeee: Got it. Thx. :)

PR #1237, opened 2024-02-18
Remove boxing when calling generic ReadField method

anderspedersen:
Remove boxing when calling ClrObject.ReadField<T> and ClrValueType.ReadField<T>.

Cast to object seems to be leftover from ClrMd v1, where ClrInstanceField.Read was not generic and returned an object.

Btw. I am unable to run the tests, I get an InvalidOperationException stating: "You must first generate test binaries and crash dumps using by running: dotnet build C:\projects\clrmd\src\TestTargets\TestTargets.csproj". But the csproj-file doesn't exist, so the command does nothing.

I am pretty sure my change is correct, but it would be nice with some documentation on how to build everything locally, so I can also run the tests :)


anderspedersen:

leculver:
Looks good, thanks!

PR #1239, opened 2024-02-22
Make dependencies to System.Collection.Immutable and System.Runtime.CompilerService.Unsafe conditional

iskiselev:
System.Collection.Immutable and System.Runtime.CompilerService.Unsafe both are included in .NET 6.0, so there is no need to reference their packages.
With it, dependent project will also not include that unneeded dependencies.


leculver:
LGTM, thanks!

PR #1240, opened 2024-02-28
Fix the way of getting ClrMethod from method handle or instruction pointer

hez2010:
The existing implementation is not able to get the method from an instantiated method handle or instruction pointer, as `type.Methods` only contains typical uninstantiated MethodDescs.
Change the way how we get ClrMethod to support getting instantiated method as well.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrType.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrType.cs
@@ -227,7 +227,7 @@ namespace Microsoft.Diagnostics.Runtime
             }
         }
 
-        public ClrMethod GetOrCreateMethod(MethodInfo methodInfo)
+        internal ClrMethod GetOrCreateMethod(MethodInfo methodInfo)
         {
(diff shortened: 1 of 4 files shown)


leculver:
I was just debugging this and created a PR before I saw yours, so you beat me to it.  :)

I'll merge yours instead, thanks for the submission!  I'll push a build tomorrow morning to NuGet

src/Microsoft.Diagnostics.Runtime/ClrType.cs
```diff
@@ -225,6 +227,19 @@ public virtual ImmutableArray<ClrThreadStaticField> ThreadStaticFields
             }
         }
 
+        public ClrMethod GetOrCreateMethod(MethodInfo methodInfo)
```
leculver, line 230:
```suggestion
        internal ClrMethod GetOrCreateMethod(MethodInfo methodInfo)
```

leculver (approved):
This should probably be internal to not create new surface area.

PR #1251, opened 2024-04-08
Wrapped additional IDebug methods

WillVan:
I needed read some additional information out of my memroy dumps, tried to follow the existing code as closely as possible.
 
* Upped IDebugSystemObjects to v4 and IDebugAdvanced to v2.

* IDebugAdvanced wrapped Request method.

* IDebugControl wrapped GetNumberProcessors, GetCurrentTimeDate, GetCurrentSystemUpTime

* IDebugSystemObjects wrapped GetCurrentProcessPeb, GetCurrentProcessUpTime, GetCurrentProcessExecutableNameWide


WillVan:

src/Microsoft.Diagnostics.Runtime.Utilities/DbgEng/Structs/MINIDUMP_STREAM_TYPE .cs
```diff
@@ -0,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Diagnostics.Runtime.Utilities.DbgEng.Structs
+{
+    public enum MINIDUMP_STREAM_TYPE : uint
+    {
+        UnusedStream = 0u,
```
leculver, line 8: nit:  The `u` is not needed on these and inconsistently used.

src/Microsoft.Diagnostics.Runtime.Utilities/DbgEng/Wrappers/IDebugAdvancedWrapper.cs
```diff
@@ -17,6 +19,42 @@ int IDebugAdvanced.GetThreadContext(Span<byte> buffer)
                 return vtable->GetThreadContext(self, ptr, buffer.Length);
         }
 
+        int IDebugAdvanced.GetCommentWide(out string? comment)
+        {
+            GetVTable(this, out nint self, out IDebugAdvancedVtable* vtable);
+            comment = null;
+            char[] buffer = ArrayPool<char>.Shared.Rent(1048576/2);
```
leculver, line 26: What is this number?  Is it meaningful?  Either way it should be pulled into a constant if this code is left as-is.

Usually, these methods follow a common patern where you can call "Request" and set `&minidumpStream` parameter to `null`.  This tells you the size.  Then allocate a buffer and pass the correct size.  Does that work for this method?

src/Microsoft.Diagnostics.Runtime.Utilities/DbgEng/Wrappers/IDebugAdvancedWrapper.cs
```diff
@@ -17,6 +19,42 @@ int IDebugAdvanced.GetThreadContext(Span<byte> buffer)
...
+                {
+                    DEBUG_READ_USER_MINIDUMP_STREAM minidumpStream = default;
+                    minidumpStream.StreamType = MINIDUMP_STREAM_TYPE.CommentStreamW;
+                    minidumpStream.Flags = 0u;
+                    minidumpStream.Offset = 0u;
+                    minidumpStream.Buffer = new IntPtr(ptr);
+                    minidumpStream.BufferSize = (uint)(buffer.Length * sizeof(char));
+                    minidumpStream.BufferUsed = 0u;
+                    HResult hr = vtable->Request(self, (uint)DEBUG_REQUEST.READ_USER_MINIDUMP_STREAM, &minidumpStream, sizeof(DEBUG_READ_USER_MINIDUMP_STREAM), null, 0, null);
+                    if (hr < 0)
```
leculver, line 39: This should follow the standard .Net coding convention.  If any part of an if needs braces, all of them should have braces.  Additionally, `else` should not be on a different line than `if`.  e.g.:

``` C#
if (hr < 0)
{
    return hr;
}
else if (minidumpStream.BufferUsed > 1)
{
    comment = new string(ptr, 0, ((int)minidumpStream.BufferUsed - 2) / 2);
    return hr;
}
else
{
    comment = string.Empty;
}
```

leculver (changes requested):
Thanks for submitting this change!  Functionally it looks ok, but can you please take a look at the issues here?

src/Microsoft.Diagnostics.Runtime.Utilities/DbgEng/Wrappers/IDebugAdvancedWrapper.cs
WillVan: Done

leculver (approved):

src/Microsoft.Diagnostics.Runtime.Utilities/DbgEng/Wrappers/IDebugAdvancedWrapper.cs
WillVan: Number added to a constant, no specific significance - 1 mb in bytes.
I cannot seem to get the request method to return the size.
int size = 0;
HResult hr = vtable->Request(self, (uint)DEBUG_REQUEST.READ_USER_MINIDUMP_STREAM, null, 0, null, 0, &size);
returns E_INVALIDARG.
I assume that for this case the request method needs the steam type.

leculver:
Thanks for the code!  I need to do a new release at some point this month, so this should be on nuget before 5/1.

WillVan:
Thank you!

PR #1253, opened 2024-05-03
Implement /proc/pid fallback on Linux

kevingosse:
I ran into a situation where createdump was able to read the memory of the crashing process but ClrMD failed with an `UnauthorizedAccessException`. After digging deeper, it looks like createdump fallbacks to reading `/proc/<pid>/mem` when `process_vm_readv` returns EPERM:

https://github.com/dotnet/runtime/blob/15e98e58e6023be3121dcb0ed2e701d1d5cd098e/src/coreclr/debug/createdump/crashinfounix.cpp#L514-L531

I don't actually know _why_ `process_vm_readv` fails (it's in an ARM64 CI agent running docker), but the fallback fixes it.


leculver:
Apologies for the delay here.  I have been focused on other work, then was on vacation.

leculver (approved):

PR #1259, opened 2024-06-05
Add support for Azure based symbol servers

leculver:
- Added support for Azure based symbol servers.
- Version bump to 4.0 since this is a very disruptive change.


src/Microsoft.Diagnostics.Runtime/CustomDataTarget.cs
```diff
@@ -36,11 +37,14 @@ public class CustomDataTarget : IDisposable
...
 
+        public TokenCredential? SymbolTokenCredential { get; set; }
+
         /// <summary>
         /// Constructor.
         /// </summary>
         /// <param name="reader">A non-null IDataReader.</param>
-        public CustomDataTarget(IDataReader reader)
+        /// <param name="symbolCredential">The TokenCredential to use for any Azure based symbol servers (set to null if not using one).</param>
+        public CustomDataTarget(IDataReader reader, TokenCredential? symbolCredential)
```
mikem8361, line 47: I think you forgot to set the SymbolTokenCredential property with the symbolCredential parameter here.  It would be nice if you set this new parameter to null so it doesn't break SOS.

leculver: I did, thank you.  Fixed here, including setting the default to null:  https://github.com/microsoft/clrmd/pull/1261

I'm racing as fast as possible to get these changes in, so I missed that.  :(

PR #1265, opened 2024-06-15
Bump to macos-latest build image

akoeplinger:
AzDO will remove macos-11 in about two weeks: https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#recent-updates

Bump to macos-latest to match the internal build.

/cc @leculver


leculver:
Thanks!

PR #1267, opened 2024-06-19
Add SpecialDiagInfo block short cut for runtime module enumeration

mikem8361:
Need the lowest common denominator version for System.Collections.Immutable

Add ForceCompleteRuntimeEnumeration CustomTarget property

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DataTarget.cs
@@ -196,7 +196,7 @@ namespace Microsoft.Diagnostics.Runtime
             if (_clrs.IsDefault)
             {
                 IEnumerable<ModuleInfo> modules = EnumerateModules();
-                IEnumerable<ClrInfo>? clrs = null;
+                IEnumerable<ClrInfo>? clrs = Array.Empty<ClrInfo>();
 
                 // First try the SpecialDiagInfo block short cut to find a supported runtime
                 if (!_target.ForceCompleteRuntimeEnumeration && DataReader.TargetPlatform != OSPlatform.Windows)
@@ -223,12 +223,15 @@ namespace Microsoft.Diagnostics.Runtime
                 // We order this so .Net Core comes first, so if there's multiple CLRs we prefer
                 // to debug .Net Core (assuming the user is just debugging one of them)
 
-                clrs ??= from module in modules
-                       let clrInfos = s_clrInfoProviders.Select(provider => provider.ProvideClrInfoForModule(this, module)).Where(clrInfo => clrInfo != null)
-                       let clrInfo = clrInfos.LastOrDefault()
-                       where clrInfo != null
-                       orderby clrInfo.Flavor descending, clrInfo.Version
-                       select clrInfo;
+                if (!clrs.Any())
+                {
+                    clrs = from module in modules
+                           let clrInfos = s_clrInfoProviders.Select(provider => provider.ProvideClrInfoForModule(this, module)).Where(clrInfo => clrInfo != null)
+                           let clrInfo = clrInfos.LastOrDefault()
+                           where clrInfo != null
+                           orderby clrInfo.Flavor descending, clrInfo.Version
+                           select clrInfo;
+                }
 
                 _clrs = clrs.ToImmutableArray();
             }
--- a/src/Microsoft.Diagnostics.Runtime/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DataTarget.cs
@@ -196,7 +196,7 @@ namespace Microsoft.Diagnostics.Runtime
             if (_clrs.IsDefault)
             {
                 IEnumerable<ModuleInfo> modules = EnumerateModules();
-                IEnumerable<ClrInfo>? clrs = Array.Empty<ClrInfo>();
+                List<ClrInfo>? clrs = null;
 
                 // First try the SpecialDiagInfo block short cut to find a supported runtime
                 if (!_target.ForceCompleteRuntimeEnumeration && DataReader.TargetPlatform != OSPlatform.Windows)
@@ -210,10 +210,10 @@ namespace Microsoft.Diagnostics.Runtime
                         if (found > 0)
                         {
                             // Run the module through the clr info providers to see if it is an app model/runtime supported by CLRMD
-                            clrs = from clrInfo in s_clrInfoProviders.Select(provider => provider.ProvideClrInfoForModule(this, _modules![found])).Where(clrInfo => clrInfo != null)
-                                   where clrInfo != null
-                                   orderby clrInfo.Flavor descending, clrInfo.Version
-                                   select clrInfo;
+                            clrs = new(
+                                from clrInfo in s_clrInfoProviders.Select(provider => provider.ProvideClrInfoForModule(this, _modules![found])).Where(clrInfo => clrInfo != null)
+                                orderby clrInfo.Flavor descending, clrInfo.Version
+                                select clrInfo);
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
```diff
@@ -34,7 +34,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0'">
-    <PackageReference Include="System.Collections.Immutable" Version="8.0.0" />
+    <PackageReference Include="System.Collections.Immutable" Version="6.0.0" />
```
leculver, line 37: Why is this downgraded?  With 4.0 I have been pushing versions up to their newest release, as long as those releases still support 6.0 (which is the most recent in support).

src/Microsoft.Diagnostics.Runtime/DataTarget.cs
```diff
@@ -195,15 +195,40 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
...
                 // We order this so .Net Core comes first, so if there's multiple CLRs we prefer
                 // to debug .Net Core (assuming the user is just debugging one of them)
 
-                IEnumerable<ClrInfo> clrs = from module in EnumerateModules()
-                                            let clrInfoEnumerable = s_clrInfoProviders.Select(provider => provider.ProvideClrInfoForModule(this, module)).Where(clrInfo => clrInfo != null)
-                                            let clrInfo = clrInfoEnumerable.LastOrDefault()
-                                            where clrInfo != null
-                                            orderby clrInfo.Flavor descending, clrInfo.Version
-                                            select clrInfo;
+                clrs ??= from module in modules
```
leculver, line 226: Does desktop CLR have the special diagnostic info block?  It doesn't, right?

That means that if one coreclr is in the process with a special diagnostic block, we will skip enumerating all modules which will find the desktop runtime?  Or am I missing something?

src/Microsoft.Diagnostics.Runtime/DataTarget.cs
```diff
@@ -195,15 +195,40 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
...
+                    if (SpecialDiagInfo.TryReadSpecialDiagInfo(DataReader, out SpecialDiagInfo info) && info.RuntimeBaseAddress != 0)
+                    {
+                        Debug.Assert(_modules != null);
+                        
+                        // The above EnumerateModules() call populated (the sorted) _modules array
+                        int found = _modules!.Search(info.RuntimeBaseAddress, (m, k) => m.ImageBase.CompareTo(k));
+                        if (found > 0)
+                        {
+                            // Run the module through the clr info providers to see if it is an app model/runtime supported by CLRMD
+                            clrs = from clrInfo in s_clrInfoProviders.Select(provider => provider.ProvideClrInfoForModule(this, _modules![found])).Where(clrInfo => clrInfo != null)
```
leculver, line 213: This sets the `clrs` variable to be non-null, skipping the below.  Aside from the question I have below, note that s_clrInfoProviders could all return null.  Meaning we will have a non null `clrs` with an empty enumeration, so it will skip the full enumeration below, even though this method failed.

I guess this couldn't happen in practice, but maybe it would be better to change 'clrs' to a List<> and build it to check what we have instead of leaving `clrs` as an enumeration in this new logic.

leculver (changes requested):
A few questions.  The actual code to read the diagnostic block looks fine to me.

src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
mikem8361: I could not get SOS to run with this change (and SOS's reference) with 8.0.0 hosted under 6.0. I'll try again to make sure I built everything correctly.

src/Microsoft.Diagnostics.Runtime/DataTarget.cs
mikem8361: No, only Linux and MacOS .NET Core have this block.  The idea is to only return the "crashing" runtime.  If a user wants all runtimes, then the new force flag should be set.  Under SOS this happens in the !runtimes command.

src/Microsoft.Diagnostics.Runtime/DataTarget.cs
mikem8361: Yes, it should for an empty list instead of using null.  I'll push a change.

src/Microsoft.Diagnostics.Runtime/DataTarget.cs
mikem8361: Only Linux and MacOS .NET core dumps created with createdump have this block. Windows, live-debugging don't.

src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
mikem8361: Using 8.0 to host under windbg works but 6.0 and 7.0 don't.

leculver: Ok, we can downgrade for now, but we can't remain pinned on 6.0.0 forever.  We'll have to figure out why `8.0.0` says it supports 6.0 runtimes (see https://www.nuget.org/packages/System.Collections.Immutable/) but doens't work when loaded there.

src/Microsoft.Diagnostics.Runtime/DataTarget.cs
leculver: This isn't quite what I had in mind.  There are lots of scenarios where we don't have a "crashing" runtime.  The change as it stands leads to inconsistent behavior in many scenarios.

I think the right approach is:

1.  Look only for coreclr/clr in the process (under the correct name for all architectures).  That scan is fast, so there's no reason to not look for those runtimes.
2. Disable the SingleFileClrInfoProvider unless the user sets "Force scanning all runtimes."
3. When we have the extra debug info block, use that to supplement 1 and 2.

This means we would always find crashing single-file apps.  We would not find non-crashing single-file apps unless the user requested to scan everything.  More importantly, we will _always_ normal "coreclr" and "clr" runtimes in the process, preserving that behavior.

I am out of office for a while at a team event today.  Let me take a closer look when I'm back tomorrow morning and maybe submit changes to your branch.

mikem8361: For all the scenarios that don't have a "crashing" runtime, it is exactly like it was before. It doesn't make sense to me to enumerate any modules when the special diag block exists. Yes, it does change the behavior but the force flag restore the behavior.

I think we need to talk about this some more before you make any changes.

src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj
mikem8361: There might something I can do in SOS's hosting code that builds the "TPA". I'll have to investigate.

src/Microsoft.Diagnostics.Runtime/DataTarget.cs
```diff
@@ -195,15 +195,43 @@ private ImmutableArray<ClrInfo> GetOrCreateClrVersions()
...
                 // We order this so .Net Core comes first, so if there's multiple CLRs we prefer
                 // to debug .Net Core (assuming the user is just debugging one of them)
 
-                IEnumerable<ClrInfo> clrs = from module in EnumerateModules()
-                                            let clrInfoEnumerable = s_clrInfoProviders.Select(provider => provider.ProvideClrInfoForModule(this, module)).Where(clrInfo => clrInfo != null)
-                                            let clrInfo = clrInfoEnumerable.LastOrDefault()
-                                            where clrInfo != null
-                                            orderby clrInfo.Flavor descending, clrInfo.Version
-                                            select clrInfo;
+                if (!clrs.Any())
```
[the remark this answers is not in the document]
mikem8361: I don't understand how.  _clrs starts in the "Default" case and even if we end up with an empty "clrs" when ToImmutableArray() is done, the resulting _clrs is not "Default" anymore. 

Or am I really missing something here?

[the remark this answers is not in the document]
mikem8361: I understand now what you mean about the enumeration being run twice.  The first is the clrs.Any() and the second time is the clrs.ToImmutableArray(). Switch to List.

leculver (approved):

PR #1268, opened 2024-06-20
Symbol cache cleanup/fixes

leculver:
Fix bug where we did not properly respect the cache set in the symbol path.


ryanmolden:
Looks good to me, though I am not sure of all the insane formats that symbol server strings can take on.

leculver:
We support what we can and try to ignore the rest...

PR #1270, opened 2024-06-22
MachOModule.cs add bound check

Foxtr0t1337:
`ReadSymbolTable()` will use `DataReader.Read()`
If `_symtab.NSyms` is `0` but it passed previous checks, `Read()` will assert.
Didn't remove the assert but add this check.
I don't know why but in a game some modules will cause the assert.


leculver (approved):
LGTM, thanks!

PR #1273, opened 2024-06-26
Version bump utilities dependency

leculver:
Update utilities dependency version.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Utilities/Microsoft.Diagnostics.Runtime.Utilities.csproj
+++ b/src/Microsoft.Diagnostics.Runtime.Utilities/Microsoft.Diagnostics.Runtime.Utilities.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Diagnostics.Runtime" Version="2.3.405304" />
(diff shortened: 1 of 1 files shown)


src/Microsoft.Diagnostics.Runtime.Utilities/Microsoft.Diagnostics.Runtime.Utilities.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Diagnostics.Runtime" Version="2.3.405304" />
+    <PackageReference Include="Microsoft.Diagnostics.Runtime" Version="4.0.0-beta.24320.3" />
```
mikem8361, line 11: Why don't you use a ProjectReference here?

leculver: It's a separate library with a separate nuget package and dependencies.  It also (theoretically) can version and move independently of the original library, though they've been roughly in sync in the past.  This lets me put experimental/useful tooling there without polluting the main ClrMD library.  These being loosely coupled has been a useful property.

Issue #1279, opened 2024-07-17
_isVirtual has to be set as true in the context of DataTarget.LoadDump.


stjeong:
### Description

At .\Microsoft.Diagnostics.Runtime\DataReaders\Core\CoreDumpReader.cs,

CreateModuleInfo always create PEModuleInfo instance with "virtualHint == false",

```
private ModuleInfo CreateModuleInfo(ElfLoadedImage image)
{
    using ElfFile? file = image.Open();

    string path = image.FileName.Replace(" (deleted)", "");
    if (file is not null)
    {
        long size = image.Size > long.MaxValue ? long.MaxValue : unchecked((long)image.Size);
        return new ElfModuleInfo(this, file, image.BaseAddress, size, path);
    }

    return new PEModuleInfo(this, image.BaseAddress, path, false); // <--- HERE!
}
`````

When dll image in memory dump file has 'export data directory', this code lead to almost infinite loop because of invalid ImageExportDirectory instance given by TryGetExportSymbol method.

```
// .\Microsoft.Diagnostics.Runtime\Utilities\PEImage\PEImage.cs
public bool TryGetExportSymbol(string symbolName, out ulong offset)
{
    int nameIndex = 0;
    try
    {
        ImageDataDirectory exportTableDirectory = ExportDirectory;
        if (exportTableDirectory.VirtualAddress != 0 && exportTableDirectory.Size != 0)
        {
            if (TryRead(RvaToOffset(exportTableDirectory.VirtualAddress), out ImageExportDirectory exportDirectory)) // <--- HERE!

     // ...omitted for brevity...
```

At the code of "RvaToOffset" method,

```
public int RvaToOffset(int virtualAddress)
{
    if (virtualAddress < 4096)
        return virtualAddress;

    if (_isVirtual) // <--- HERE!
        return virtualAddress;

    ImageSectionHeader[] sections = ReadSections(); // <---- Should not run this code in the context of memory dump
    for (int i = 0; i < sections.Length; i++)
    {
        ref ImageSectionHeader section = ref sections[i];
        if (section.VirtualAddress <= virtualAddress && virtualAddress < section.VirtualAddress + section.VirtualSize)
            return (int)(section.PointerToRawData + ((uint)virtualAddress - section.VirtualAddress));
    }

    return -1;
}
```

Because _isVirtual is false, it doesn't run second if's "return virtualAddress", instead it calculates the value as RVA for PE offset, like loading from FILE directly.

[...]

PR #1280, opened 2024-07-18
Changed the boolean parameter in the PEModuleInfo constructor from false to true

stjeong:
Changed the boolean parameter in the PEModuleInfo constructor from false to true within CoreDumpReader.cs (https://github.com/microsoft/clrmd/issues/1279)


leculver (approved):
LGTM, thanks!

PR #1285, opened 2024-09-12
Fix a couple of issues for statics in sos14

kouvel:
- Followup to https://github.com/microsoft/clrmd/pull/1284, passed in the method table address to the statics APIs instead of the module address
- Treated `ClrElementType.Pointer` as primitive, as for thread-statics they seem to be stored in the non-GC section


leculver (approved):
LGTM, thanks!

PR #1292, opened 2024-10-24
[LoongArch64] Fix the issue of incorrect printing information in the gcroot command

lawn123:
Adjust the offset values of SP and IP as well as the contextSize value in the EnumerateSackTrace function.


lawn123:
@shushanhf @mikem8361 Could you please review this PR？

mikem8361 (approved):

PR #1299, opened 2024-12-09
Change netstandard2 -> net47

leculver:
Move to a more explicit supported runtime.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/LegacyAttributes.cs
+++ b/src/Microsoft.Diagnostics.Runtime/LegacyAttributes.cs
@@ -3,7 +3,7 @@
 
 using System.ComponentModel;
 
-#if NETSTANDARD2_0
+#if NET47_OR_GREATER
 namespace System.Diagnostics.CodeAnalysis
 {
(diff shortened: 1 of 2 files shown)


mikem8361:
SOS still needs netstandard2.0! Until SOS builds/installs/uses 2 different versions (one for .NET 6 or more and one for Framework) we build everything as netstandard2.0 so it can be hosted under Framework (and it is in Watson).

timcassell:
> SOS still needs netstandard2.0! Until SOS builds/installs/uses 2 different versions (one for .NET 6 or more and one for Framework) we build everything as netstandard2.0 so it can be hosted under Framework (and it is in Watson).

Same with BenchmarkDotNet. It actually used to do separate framework and core targets, but unified to netstandard2.0 to fix some issues a while back.

leculver:
Feedback heard!

Keeping netstandard 2 really makes maintaining this project tough.  Using netstandard pulls in a lot of dependencies that throw our security checks for a loop, even if they aren't used.  It's a constant drain on resources keeping this project afloat instead of working on other things.

In the future, I would like to drop netstandard2 support, but for now I'll close this PR.

PR #1311, opened 2025-03-28
Add cDac runtime enumeration support

mikem8361:
Adds a new DebugLibaryKind.CDac enum used when the data contract export is found on a runtime module.

Currently the file path to the cdac is the same directory as the clrmd module.

Currently leaves it to the client (SOS in this case) to decide when it wants to instantiate the cdac. The standalone hosting support doesn't currently pick the cdac.

Breaking changes:

Add "bool verifySignature" to IClrInfoProvider.GetDacServices and remove the CustomDataTarget.DacSignatureVerificationEnabled property.

Adds data target support for ICLRContractLocator for when clrmd loads the cdac.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrInfo.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrInfo.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Diagnostics.Runtime
         /// </summary>
         /// <param name="dacPath">A full path to the matching DAC dll for this process.</param>
         /// <returns>The runtime associated with this CLR.</returns>
-        public ClrRuntime CreateRuntime(string dacPath)=> CreateRuntime(dacPath, ignoreMismatch: false, verifySignature: false);
+        public ClrRuntime CreateRuntime(string dacPath) => CreateRuntime(dacPath, ignoreMismatch: false, verifySignature: false);
 
         /// <summary>
         /// Creates a runtime from the given DAC file on disk.
--- a/src/Microsoft.Diagnostics.Runtime/ClrInfo.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrInfo.cs
@@ -61,6 +61,11 @@ namespace Microsoft.Diagnostics.Runtime
         public ModuleInfo ModuleInfo { get; }
 
         /// <summary>
+        /// The CDAC contract export address
+        /// </summary>
+        public ulong ContractDescriptor { get; set; }
+
+        /// <summary>
         /// The timestamp under which this CLR is is archived (0 if this module is indexed under
         /// a BuildId instead).  Note that this may be a different value from ModuleInfo.IndexTimeStamp.
         /// In a single-file scenario, the ModuleInfo will be the info of the program's main executable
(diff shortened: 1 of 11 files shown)


src/Microsoft.Diagnostics.Runtime/ClrInfo.cs
```diff
@@ -88,49 +93,49 @@ internal ClrInfo(DataTarget dt, ModuleInfo module, Version clrVersion, IClrInfoP
...
-        {
-            if (string.IsNullOrEmpty(dacPath))
-                throw new ArgumentNullException(nameof(dacPath));
-
-            if (!File.Exists(dacPath))
-                throw new FileNotFoundException(dacPath);
-
-            return CreateRuntimeWorker(dacPath, ignoreMismatch: false);
-        }
+        public ClrRuntime CreateRuntime(string dacPath)=> CreateRuntime(dacPath, ignoreMismatch: false, verifySignature: false);
```
max-charlamb, line 108: nit
```suggestion
        public ClrRuntime CreateRuntime(string dacPath) => CreateRuntime(dacPath, ignoreMismatch: false, verifySignature: false);
```

src/Microsoft.Diagnostics.Runtime/Implementation/DotNetClrInfoProvider.cs
```diff
@@ -319,6 +327,28 @@ protected ClrInfo CreateClrInfo(DataTarget dataTarget, ModuleInfo module, ulong
...
+            if (flavor == ClrFlavor.Core)
+            {
+                contractDescriptor = module.GetExportSymbolAddress(c_cdacContractDescriptorExport);
+                if (contractDescriptor != 0)
+                {
+                    string? cdacName = GetCDacFileName(currentPlatform);
+                    if (cdacName is not null)
+                    {
+                        // The CDAC is located in the same directory as CLRMD
+                        cdacName = Path.Combine(s_defaultAssembliesPath, cdacName);
```
max-charlamb, line 339: What directory is CLRMD located in? This is looking at where SOS is installed for the cDAC?

max-charlamb (commented):
LGTM

src/Microsoft.Diagnostics.Runtime/Implementation/DotNetClrInfoProvider.cs
mikem8361: CLRMD is located in the same directory as the rest of SOS.  I thought about making this path more flexible and settable as property on the CustomTarget instance or something.

leculver (changes requested):
Marking as Request Changes until I understand the change to `DacSignatureVerificationEnabled`.

src/Microsoft.Diagnostics.Runtime/CustomDataTarget.cs
```diff
@@ -47,11 +47,6 @@ public class CustomDataTarget : IDisposable
         /// </summary>
         public bool ForceCompleteRuntimeEnumeration { get; set; }
 
-        /// <summary>
-        /// If true, enforces the proper DAC certificate signing
-        /// </summary>
-        public bool DacSignatureVerificationEnabled { get; set; }
```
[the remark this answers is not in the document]
mikem8361: It has moved from the CustomDataTarget to a parameter on ClrInfo.CreateRuntime verifySignature.  The reason is that SOS (and probably other clients), don't really know if they want to verify the signature until the runtime is created if the CDAC is being loaded.  It isn't going to be signed with the special cert (just the regular one) so it can't be verified. And currently SOS doesn't know if it is loading the cDac or regular DAC until later.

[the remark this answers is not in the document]
leculver: Ah ok that makes sense.

PR #1317, opened 2025-05-08
Dispose dump file stream when leaveOpen is false

zachcmadsen:
I was writing a test that gave an invalid dump to DataTarget.LoadDump, and I noticed that the file stream wasn't closed. It looks like the condition has been wrong since it was originally added


mikem8361:
I need to make sure this doesn't cause any problems in SOS tests before I approve.  It shouldn't but you never know.

mikem8361 (approved):

PR #1324, opened 2025-07-31
Update/Modernize code

leculver:
- Update projects to .Net 8.
- Remove obsolete serialization overloads in exception class.
- Don't use `field` as a variable name as this is a new keyword in a preview version of C#.
- Fixed an issue where we called the wrong MemoryMarshal overload.


marklio (approved):
LGTM

PR #1366, opened 2026-02-16
Fix crash analyzing system dumps (gcore/coredumpctl)

leculver:
We shouldn't hard-crash when analyzing incorrectly taken crash dumps.  Contributes to https://github.com/dotnet/diagnostics/issues/5632.

- Add IsCreatedByDotNetRuntime to IDumpInfoProvider to detect non-createdump dumps and improve error messages
- Fix off-by-one in SpecialDiagInfo module lookup (found > 0 -> found >= 0)
- Guard against null Assembly address in GetModuleInfo

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/DataReaders/Minidump/MinidumpReader.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DataReaders/Minidump/MinidumpReader.cs
@@ -59,6 +59,8 @@ namespace Microsoft.Diagnostics.Runtime
 
         public bool IsMiniOrTriage => _minidump.IsMiniDump;
 
+        public bool IsCreatedByDotNetRuntime => true;
+
         public void Dispose()
         {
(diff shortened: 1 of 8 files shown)


src/Microsoft.Diagnostics.Runtime/DataReaders/Minidump/MinidumpReader.cs
```diff
@@ -57,6 +57,8 @@ public MinidumpReader(string displayName, Stream stream, CacheOptions cacheOptio
 
         public bool IsMiniOrTriage => _minidump.IsMiniDump;
 
+        public bool IsCreatedByDotNetRuntime => true;
```
hoyosjs: I've seen this - I've asked the question to a few dex folks but there's largely been no way to see if aux providers were loaded. Maybe we can try probing a canonical structure in a follow up change as a heuristic.

PR #1375, opened 2026-02-18
Build test .dmps on demand

leculver:
- Changed the test infrastructure to build .dmp files on demand instead of requiring a pre-step of building testtasks -> testtargets.
- We used to test framework on windows, .net core on linux.  Now we test both on windows (.net core still on linux).
- Now test x86/x64 architectures and not just x64.


marklio (approved):
LGTM. I'm leaning on the fact this is a test-heavy PR and so is largely "self-testing"

PR #1380, opened 2026-02-20
Validate addresses before calling DAC to avoid crashes on older runtimes

leculver:
Add zero-address guards to GetAppDomainName, TraverseModuleMap, TraverseStubHeap, GetFieldInfo, and GetMethodTableData. These prevent passing null/zero to DAC methods that crash without null checks on older runtimes (dotnet/runtime#124640).


timcassell:
There were also crashes due to -1 addresses in older runtimes (net7 and earlier). In BDN we just filter the entire invalid address range. Not sure if that strategy makes sense for every API, but maybe worth considering. 

https://github.com/dotnet/BenchmarkDotNet/blob/f54f764266722c6eba4e6f6ded8c808a4012a34e/src/BenchmarkDotNet/Disassemblers/ClrMdDisassembler.cs#L19-L44

PR #1385, opened 2026-02-23
Add DataTargetOptions, remove CustomDataReader

leculver:
Breaking change for v4.

Removed `CustomDataReader`.  All of that has been moved to DataTargetOptions.  Now we simply have a constructor for DataTarget which takes in an IDataReader and a DataTargetOptions.  Much cleaner and simpler.  Static methods stille xist to load dumps and attach to processes, since those data readers are marked internal (we don't want folks taking direct dependencies on their implementation).

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/DumpGenerator.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/DumpGenerator.cs
@@ -53,6 +53,12 @@ namespace Microsoft.Diagnostics.Runtime.Tests
                 {
                     GenerateFrameworkDump(exePath, dumpPath, gcMode, full);
                 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Runtime/MacOS/MachOCoreDump.cs
```diff
@@ -344,7 +344,10 @@ private bool FindSegmentContaining(ulong address, out MachOSegment seg)
         public void Dispose()
         {
             if (!_leaveOpen)
+            {
+                _leaveOpen = false; // allow multiple calls to Dispose without throwing
```
hoyosjs: If you entered, leave open is false. Does something else reset it

leculver: Oh, thanks :)

PR #1412, opened 2026-04-03
Fix GetTypeByName to find cached generic instantiations

### LINKED ISSUE #1396: LinkedListNode.item property returns UnknowType
I've tried latest version: 2384507 - Improve error messages for uninitialized/corrupt dumps (#1395)

and see that for LinkedList nodes it does not work correctly although version 4.0.0-beta.24360.3 does

leculver:
GetTypeByName only searched TypeDef maps, missing generic instantiations (e.g. LinkedListNode<System.String>) that were already constructed during heap enumeration. Add TryGetCachedTypeByName to ClrTypeFactory that searches all cached types by name, and use it as a fallback in ClrHeap.GetTypeByName.

This improves field type resolution for generic types where:
- The type argument is itself a generic instantiation
- The concrete generic type needs to be found by constructed name
- The field signature's GenericInstantiation or Var elements need resolved concrete types that aren't in any module's TypeDef map

Add regression tests verifying all LinkedListNode fields have non-null types and that GetTypeByName finds constructed generic instantiations.

Fixes #1396 .

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/GenericFieldTests.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/GenericFieldTests.cs
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#nullable enable
+
 using System.Linq;
 using Xunit;
 
@@ -30,7 +32,7 @@ namespace Microsoft.Diagnostics.Runtime.Tests
 
             Assert.True(node.IsValid, "Could not find LinkedListNode<string> on the heap.");
 
-            ClrInstanceField itemField = node.Type!.GetFieldByName("item");
+            ClrInstanceField? itemField = node.Type!.GetFieldByName("item");
             Assert.NotNull(itemField);
 
             // The field type should be System.String, not T or System.__Canon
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Runtime.Tests/src/GenericFieldTests.cs
```diff
@@ -100,6 +102,52 @@ public void Fields_ShouldNotExposeCanonTypes()
...
+        /// just TypeDef maps.
+        /// </summary>
+        [Fact]
+        public void LinkedListNode_AllFields_HaveNonNullType()
+        {
+            ClrHeap heap = _connection.Runtime.Heap;
+            ClrObject node = heap.EnumerateObjects()
+                .FirstOrDefault(o => o.Type?.Name?.Contains("LinkedListNode<System.String>") == true);
+
+            Assert.True(node.IsValid, "Could not find LinkedListNode<string> on the heap.");
```
brianrob, line 119: Not sure if this is relevant, but would it make sense to include the `IsValid` check in the lambda in addition to the assert here?  Or is this a failure if you ever find an invalid one?

brianrob (approved):
One question below, but approving in case it's not relevant.

src/Microsoft.Diagnostics.Runtime.Tests/src/GenericFieldTests.cs
leculver: ClrObject is actually a struct, so FirstOrDefault can return an empty ClrObject where isvalid is false.  This is basically a nullcheck.

PR #1419, opened 2026-04-14
Look up TEB from OS thread ID instead of DacpThreadData.Teb

max-charlamb:
https://github.com/dotnet/runtime/pull/126902 removes the Windows only TEB field from the coreclr Thread concept, DAC, and cDAC. This field only exposed data on Windows which was already queryable using other means.

DacRuntime.EnumerateThreads() was reading StackBase and StackLimit by dereferencing the TEB address from DacpThreadData.Teb. 

Instead, look up the TEB via IThreadReader.GetThreadTeb(osThreadId) which all ClrMD data reader implementations already support. This retrieves the TEB from the debugger/dump directly using the OS thread ID, independent of the DAC. The ClrThreadInfo.Teb field is also populated from this source.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacRuntime.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacRuntime.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using Microsoft.Diagnostics.Runtime.AbstractDac;
 using Microsoft.Diagnostics.Runtime.DacInterface;
+using Microsoft.Diagnostics.Runtime.DataReaders.Implementation;
 using Microsoft.Diagnostics.Runtime.Implementation;
 
 namespace Microsoft.Diagnostics.Runtime.DacImplementation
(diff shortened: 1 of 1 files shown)


leculver (changes requested):
This might be right for newer .Net versions, but remember ClrMD also can debug DesktopCLR, and IThreadReader may not be implemented everywhere.

Instead of this change, prefer `IThreadReader.GetThreadTeb`.  If the data reader isn't an IThreadReader or if `GetThreadTeb` returns 0, set `teb = threadData.Teb` (which will no-op on newer dotnets but is correct for DesktopCLR).  Then, `if (teb != 0)` fill in stackBase/stackLimit.

max-charlamb:
> This might be right for newer .Net versions, but remember ClrMD also can debug DesktopCLR, and IThreadReader may not be implemented everywhere.
> 
> Instead of this change, prefer `IThreadReader.GetThreadTeb`. If the data reader isn't an IThreadReader or if `GetThreadTeb` returns 0, set `teb = threadData.Teb` (which will no-op on newer dotnets but is correct for DesktopCLR). Then, `if (teb != 0)` fill in stackBase/stackLimit.

Makes sense. I updated the PR. Thank you for the feedback.

leculver (approved):

PR #1421, opened 2026-04-15
Fix cross-bitness dump reading: replace IntPtr.Size with target PointerSize

max-charlamb:
## Summary

Replace ~50 uses of `IntPtr.Size` (host pointer size) with the target process's `PointerSize` across 14 files so that dumps from a different architecture can be read correctly (e.g. reading ARM64 dumps from an x64 host, or x86 dumps from an x64 host).

## Motivation

Cross-platform diagnostic scenarios (e.g. [this CI pipeline](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1377349&view=results)) fail when reading dumps from a target with a different pointer size than the host. The root cause is that ClrMD uses `IntPtr.Size` in many places to compute object layout offsets, field addresses, GC descriptor sizes, and heap alignment — all of which must use the **target's** pointer size, not the host's.

### Files modified (14):

| File | What changed |
|------|-------------|
| `ClrArray.cs` | Array length offset, multi-dim detection, rank calc, element addressing (6 locations) |
| `ClrObject.cs` | Boxed value header offset |
| `ClrInstanceField.cs` | Field address calculation |
| `ClrField.cs` | `GetSize` accepts optional `pointerSize` parameter |
| `ClrException.cs` | All 4 offset methods + stack trace parsing |
| `ClrHeap.cs` | String offsets, alignment, min object size, GC ref scanning, finalize queue walking, `MemoryCache.ReadPointer` (12+ locations) |
| `ClrDacType.cs` | GC descriptor reading, array element type detection, base array offset |
| `ClrTypeFactory.cs` | String type `StaticSize` |
| `GCDesc.cs` | New constructor overload with explicit `pointerSize`; all internal methods converted |
| `DacHeap.cs` | Frozen segment committed start |
| `DacThreadHelpers.cs` | IP/SP reads use target pointer size |
| `CommonMemoryReader.cs` | **Bug fix** (see below) |
| `SpanExtensions.cs` | New 3-arg `AsPointer` overload (`pointerSize`, `offset`) |
| `SigParser.cs` | `_pointerSize` field for `PeekElemTypeSize` |

### Bug fix: CommonMemoryReader.ReadPointer

`ReadPointer` was calling `AsPointer(pointerSize)` which resolved to the 2-arg overload `AsPointer(span, int offset)`, treating the pointer size as an offset. Fixed to `AsPointer(pointerSize, 0)` to use the correct 3-arg overload.

## Testing

- **x64 tests**: 379 pass, 1 fail (`GCRootTests.FindAllPaths` — pre-existing on baseline), 4 skip
- **x86 tests**: 257 pass, 123 fail, 4 skip — identical to baseline
- **Cross-bitness smoke test**: Successfully loaded 21 ARM64 Windows dumps from x64 host — pointer size correctly detected as 8, architecture as Arm64, modules enumerated, CLR versions found
- **No regressions**: Verified by stashing changes and running baseline — identical results

## What was NOT changed

Files using `IntPtr.Size` for **host-side** purposes (correctly):
- `CacheOptions.cs`, `DacDataTargetCOM.cs`, `WindowsProcessDataReader.cs`, `VTableBuilder.cs`, `ComCallableIUnknown.cs`, cache entry classes

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrInstanceField.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrInstanceField.cs
@@ -38,6 +38,21 @@ namespace Microsoft.Diagnostics.Runtime
         }
 
         /// <summary>
+        /// Reads the value of a pointer-sized field using the target's pointer size.
+        /// </summary>
+        internal ulong ReadPointer(ulong objRef, bool interior)
+        {
+            ulong address = GetAddress(objRef, interior);
+            if (address == 0)
+                return 0;
+
+            if (!ContainingType.Module.DataReader.ReadPointer(address, out ulong value))
+                return 0;
+
+            return value;
+        }
+
+        /// <summary>
         /// Reads the value of an object field.
         /// </summary>
         /// <param name="objRef">The object to read the instance field from.</param>
--- a/src/Microsoft.Diagnostics.Runtime/ClrInstanceField.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrInstanceField.cs
@@ -110,7 +110,7 @@ namespace Microsoft.Diagnostics.Runtime
             if (interior)
                 return objRef + (ulong)Offset;
 
-            return objRef + (ulong)(Offset + IntPtr.Size);
+            return objRef + (ulong)(Offset + ContainingType.Module.DataReader.PointerSize);
         }
     }
 }
--- a/src/Microsoft.Diagnostics.Runtime/ClrObject.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrObject.cs
@@ -492,9 +492,17 @@ namespace Microsoft.Diagnostics.Runtime
 
             ulong mt;
             if (field.ElementType == ClrElementType.NativeInt)
-                mt = (ulong)ReadField<nint>("m_handle");
+            {
+                mt = field.ReadPointer(Address, interior: false);
+            }
             else
-                mt = (ulong)ReadValueTypeField("m_handle").ReadField<nint>("m_ptr");
+            {
+                ClrValueType handle = ReadValueTypeField("m_handle");
+                ClrInstanceField? ptrField = handle.Type?.GetFieldByName("m_ptr");
+                if (ptrField is null)
+                    return null;
+                mt = ptrField.ReadPointer(handle.Address, interior: true);
+            }
 
             return type.Heap.Runtime.GetTypeByMethodTable(mt);
         }
--- a/src/Microsoft.Diagnostics.Runtime/ClrObject.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrObject.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Diagnostics.Runtime
             if (helpers is null)
                 return default;
 
-            return DataReader.Read<T>(Address + (ulong)IntPtr.Size);
+            return DataReader.Read<T>(Address + (ulong)DataReader.PointerSize);
         }
 
         public bool IsException => Type != null && Type.IsException;
--- a/src/Microsoft.Diagnostics.Runtime/DacInterface/MetadataImport.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacInterface/MetadataImport.cs
@@ -233,14 +233,6 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             return true;
         }
 
-        public SigParser GetSigFromToken(int token)
-        {
-            HResult hr = VTable.GetSigFromToken(Self, token, out IntPtr sig, out int len);
-            if (hr)
-                return new SigParser(sig, len);
-
-            return default;
-        }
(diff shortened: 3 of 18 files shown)


src/Microsoft.Diagnostics.Runtime/Extensions/SpanExtensions.cs
```diff
@@ -45,19 +45,28 @@ public static unsafe string GetString(this Encoding encoding, ReadOnlySpan<byte>
             }
         }
 #endif
-        public static unsafe ulong AsPointer(this Span<byte> span) => AsPointer(span, 0);
+        public static unsafe ulong AsPointer(this Span<byte> span) => AsPointer(span, IntPtr.Size, 0);
```
leculver: This adds some IntPtr.Size calls.  I assume these needs to be replaced as well?

There's likely also pointer math somewhere buried throughout the code.  We've never had a cross 32 <-> 64 dac before (but it's cool we are headed there!), and as a result there's probably more than just nuint, nint, IntPtr.Size.

leculver:
Look at these as well:
https://github.com/microsoft/clrmd/blob/cb7cd841e91d72c358a8a0caec3d8db41a02dfd5/src/Microsoft.Diagnostics.Runtime/ClrDelegate.cs#L74

https://github.com/microsoft/clrmd/blob/cb7cd841e91d72c358a8a0caec3d8db41a02dfd5/src/Microsoft.Diagnostics.Runtime/ClrEnum.cs#L98-L99

https://github.com/microsoft/clrmd/blob/cb7cd841e91d72c358a8a0caec3d8db41a02dfd5/src/Microsoft.Diagnostics.Runtime/ClrField.cs#L258

Maybe AsPointer overloads should all take in a pointer size and not have any IntPtr.Size itself.

It's ok to make breaking changes here if you have to.  We haven't yet shipped clrmd 4 yet.

src/Microsoft.Diagnostics.Runtime/Utilities/SigParser/SigParser.cs
```diff
@@ -10,15 +10,21 @@ public unsafe struct SigParser
...
         public SigParser(SigParser rhs)
         {
             _sig = rhs._sig;
             _len = rhs._len;
             _offs = rhs._offs;
+            _pointerSize = rhs._pointerSize;
         }
 
-        public SigParser(IntPtr sig, int len)
+        public SigParser(IntPtr sig, int len) : this(sig, len, IntPtr.Size)
```
leculver: We likely have to just remove this overload and plumb the pointers size throughout.

src/Microsoft.Diagnostics.Runtime/ClrInstanceField.cs
```diff
@@ -37,6 +37,21 @@ public T Read<T>(ulong objRef, bool interior) where T : unmanaged
             return value;
         }
 
+        /// <summary>
+        /// Reads the value of a pointer-sized field using the target's pointer size.
+        /// </summary>
+        public ulong ReadPointer(ulong objRef, bool interior)
```
leculver: This should probably be marked `internal`.

src/Microsoft.Diagnostics.Runtime/ClrObject.cs
```diff
@@ -492,9 +492,15 @@ public bool TryReadField<T>(string fieldName, out T result)
 
             ulong mt;
             if (field.ElementType == ClrElementType.NativeInt)
-                mt = (ulong)ReadField<nint>("m_handle");
```
leculver, line 495: nit: Please add `{` and `}` to this clause too.

.gitignore
```diff
@@ -358,3 +358,4 @@ MigrationBackup/
 .ionide/
 
 
+CROSS_BITNESS_PLAN.md
```
leculver: Is this needed?

src/Microsoft.Diagnostics.Runtime/DacInterface/MetadataImport.cs
```diff
@@ -17,14 +17,18 @@ internal sealed unsafe class MetadataImport : CallableCOMWrapper
...
+        private readonly int _pointerSize;
+
+        public MetadataImport(RefCountedFreeLibrary? library, IntPtr pUnknown, int pointerSize = 8)
             : base(library, IID_IMetaDataImport, pUnknown)
         {
+            _pointerSize = pointerSize;
         }
 
-        public MetadataImport(DacLibrary library, IntPtr pUnknown)
+        public MetadataImport(DacLibrary library, IntPtr pUnknown, int pointerSize = 8)
```
leculver: Actually, I think this is all dead code.  The only place _pointerSize/IntPtr.Size was used is `PeekElemTypeSize`.  Can you remove _pointerSize entirely from MetadataImport and just delete that method?

leculver (commented):
A few remaining issues, but overall looks great.

leculver (approved):

PR #1423, opened 2026-04-18
Fix cross-bitness address truncation in ClrDataAddress conversions

max-charlamb:
## Summary

Refactor `ClrDataAddress` to correctly handle sign-extended addresses from the DAC and eliminate cross-platform ABI ambiguity at the COM interop boundary. This prevents silent address truncation when inspecting 32-bit targets and ensures correct marshalling on all platforms.

## Problem

`ClrDataAddress` used implicit operators that converted through `nint`/`nuint`:
- `operator ulong`: `(nuint)Value` — truncates 64-bit addresses to 32 bits on x86 hosts
- `operator ClrDataAddress(ulong)`: `(nint)value` — same truncation on input

This corrupts **all** addresses when a 32-bit host reads a 64-bit dump, causing:
- `GetExportSymbolAddress()` returning 0
- `PEImage` being null
- `Could not find DotNetRuntimeContractDescriptor` errors

Additionally, the DAC's sign-extended 32-bit addresses (e.g., `0xFFFFFFFF_80123456`) caused `OverflowException` in checked arithmetic contexts in `DacDataTargetCOM.ReadVirtual`.

### 1. Sign Extension
The legacy DAC sign-extends 32-bit target addresses on the wire:
- `0x80123456` becomes `0xFFFFFFFF_80123456` (CLRDATA_ADDRESS)
- `FromAddress(addr, target)` applies sign extension: `(ulong)(long)(int)(uint)addr`
- `ToAddress(target)` strips it: `(ulong)(uint)_value`

### 2. ABI Boundary — Managed-Only Type
`ClrDataAddress` **never crosses the ABI boundary by value**. Vtable slots use plain `ulong` (annotated `/*ClrDataAddress*/`) because single-field struct passing differs across platform ABIs. Managed wrappers convert via:
- `addr.ToInteropAddress()` — unwrap to raw ulong (no transform)
- `ClrDataAddress.FromInteropAddress(raw)` — wrap from raw ulong (no transform)

Pointer forms (`ref`/`out`/`ClrDataAddress*`) are safe as-is and preserved in vtable signatures.

### 3. Confinement
`ClrDataAddress` exists **only** in `DacInterface/` (vtable wrappers, struct definitions) and `DacImplementation/` (conversion boundary). It never leaks into `AbstractDac/`, `Implementation/`, or the public API. The rest of ClrMD uses plain `ulong` for all addresses.

### 4. Data Flow
```
DacImplementation (ulong)
    | FromAddress(addr, _target)
    v
SosDac wrapper (ClrDataAddress)
    | .ToInteropAddress()
    v
Vtable call (ulong on the wire) --> Native DAC
    |
Returns via out struct fields (ClrDataAddress)
    | .ToAddress(_target)
    v
DacImplementation (ulong, clean target address)
    |
Public API (ulong: ClrObject.Address, ClrType.MethodTable, etc.)
```

### ClrDataAddress API
- **Removed** implicit operators between `ClrDataAddress` and `ulong`
- **Added** `ToAddress(TargetProperties)` / `FromAddress(ulong, TargetProperties)` — sign-extension-aware conversions
- **Added** `ToInteropAddress()` / `FromInteropAddress(ulong)` — raw wire-value helpers for the ABI boundary
- **Added** `ClrDataAddress.Null` property — replaces `default` for clarity
- **Added** `IsNull`, `IEquatable<ClrDataAddress>`, hex `DebuggerDisplay`
- **Added** `TargetProperties` struct — holds pointer size, threaded through DAC layer
- **Added** comprehensive design documentation as block comment

### Vtable Signatures (all SosDac files)
Flipped every by-value `ClrDataAddress` parameter to `ulong /*ClrDataAddress*/`:
- `SosDac.cs` — 48 vtable slots, 47 wrapper call sites
- `SosDac6.cs`, `SosDac8.cs`, `SosDac13.cs`, `SosDac14.cs`, `SOSDac13Old.cs`
- `ClrDataProcess.cs` — including `CLRDATA_ENUM` handle corrections
- `DacDataTargetCOM.cs` — address-typed slots annotated

Verified slot-by-slot against `sospriv.idl` in dotnet/runtime.

### DacImplementation Conversion Sites
- All callers use `ClrDataAddress.FromAddress(ulong, _target)` to wrap addresses for DAC calls
- All DAC return values / struct fields use `.ToAddress(_target)` to unwrap
- Replaced all `default` ClrDataAddress usages with `ClrDataAddress.Null`
- Removed all 28 redundant `TargetProperties target = _target` local copies

### Inbound DAC Callback Fix
- `DacDataTargetCOM.ReadVirtual`: wrap `(long)address` in `unchecked(...)` to prevent `OverflowException` on sign-extended 32-bit addresses

## Testing

- Build: 0 errors, 0 warnings on both `netstandard2.0` and `net10.0`
- x64 tests: 308 pass / 138 fail — matches main baseline (pre-existing `SingleFileBasicArrayTests` failures unrelated)
- x86: known `OverflowException` in `ClrDataAddress.ToAddress` for checked context — pre-existing, fix pending (`unchecked((uint)_value)`)

## Follow-up from #1421

This PR addresses the broader `ClrDataAddress` design issue discovered after #1421 (IntPtr.Size fixes) was merged.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
--- a/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
@@ -23,7 +23,7 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
 
         public IEnumerable<ClrRcwCleanupData> EnumerateRcwCleanupData()
         {
-            return _sos.EnumerateRCWCleanup(default).Select(r => new ClrRcwCleanupData(r.Rcw.ToAddress(_target), r.Context.ToAddress(_target), r.Thread.ToAddress(_target), r.IsFreeThreaded));
+            return _sos.EnumerateRCWCleanup(ClrDataAddress.Null).Select(r => new ClrRcwCleanupData(r.Rcw.ToAddress(_target), r.Context.ToAddress(_target), r.Thread.ToAddress(_target), r.IsFreeThreaded));
         }
 
         public bool GetCcwInfo(ulong obj, out CcwInfo info)
--- a/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
@@ -13,34 +13,37 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
     internal class DacComHelpers : IAbstractComHelpers
     {
         private readonly SOSDac _sos;
+        private readonly TargetProperties _target;
 
-        public DacComHelpers(SOSDac sos)
+        public DacComHelpers(SOSDac sos, TargetProperties target)
         {
             _sos = sos;
+            _target = target;
         }
 
         public IEnumerable<ClrRcwCleanupData> EnumerateRcwCleanupData()
         {
-            return _sos.EnumerateRCWCleanup(0).Select(r => new ClrRcwCleanupData(r.Rcw, r.Context, r.Thread, r.IsFreeThreaded));
+            return _sos.EnumerateRCWCleanup(default).Select(r => new ClrRcwCleanupData(r.Rcw.ToAddress(_target), r.Context.ToAddress(_target), r.Thread.ToAddress(_target), r.IsFreeThreaded));
         }
 
         public bool GetCcwInfo(ulong obj, out CcwInfo info)
         {
             info = default;
-            if (_sos.GetObjectData(obj, out ObjectData objData) &&
-                objData.CCW != 0 &&
+            if (_sos.GetObjectData(ClrDataAddress.FromAddress(obj, _target), out ObjectData objData) &&
+                !objData.CCW.IsNull &&
                 _sos.GetCCWData(objData.CCW, out CcwData ccwData))
             {
+                ulong ccwAddr = objData.CCW.ToAddress(_target);
                 COMInterfacePointerData[]? pointers = _sos.GetCCWInterfaces(objData.CCW, ccwData.InterfaceCount);
                 info = new()
                 {
-                    Address = objData.CCW,
-                    IUnknown = ccwData.OuterIUnknown,
-                    Object = ccwData.ManagedObject,
-                    Handle = ccwData.Handle,
+                    Address = ccwAddr,
+                    IUnknown = ccwData.OuterIUnknown.ToAddress(_target),
+                    Object = ccwData.ManagedObject.ToAddress(_target),
+                    Handle = ccwData.Handle.ToAddress(_target),
                     RefCount = ccwData.RefCount,
                     JupiterRefCount = ccwData.JupiterRefCount,
-                    Interfaces = pointers?.Select(r => new ComInterfaceEntry() { InterfacePointer = r.InterfacePointer, MethodTable = r.MethodTable }).ToArray() ?? Array.Empty<ComInterfaceEntry>()
+                    Interfaces = pointers?.Select(r => new ComInterfaceEntry() { InterfacePointer = r.InterfacePointer.ToAddress(_target), MethodTable = r.MethodTable.ToAddress(_target) }).ToArray() ?? Array.Empty<ComInterfaceEntry>()
                 };
             }
 
@@ -50,21 +53,22 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
         public bool GetRcwInfo(ulong obj, out RcwInfo info)
         {
             info = default;
-            if (_sos.GetObjectData(obj, out ObjectData objData) &&
-                objData.RCW != 0 &&
+            if (_sos.GetObjectData(ClrDataAddress.FromAddress(obj, _target), out ObjectData objData) &&
+                !objData.RCW.IsNull &&
                 _sos.GetRCWData(objData.RCW, out RcwData rcwData))
             {
+                ulong rcwAddr = objData.RCW.ToAddress(_target);
                 COMInterfacePointerData[]? pointers = _sos.GetRCWInterfaces(objData.RCW, rcwData.InterfaceCount);
                 info = new()
                 {
-                    Address = objData.RCW,
-                    IUnknown = rcwData.IUnknownPointer,
-                    Object = rcwData.ManagedObject,
+                    Address = rcwAddr,
+                    IUnknown = rcwData.IUnknownPointer.ToAddress(_target),
+                    Object = rcwData.ManagedObject.ToAddress(_target),
                     RefCount = rcwData.RefCount,
-                    CreatorThread = rcwData.CreatorThread,
+                    CreatorThread = rcwData.CreatorThread.ToAddress(_target),
                     IsDisconnected = rcwData.IsDisconnected != 0,
-                    VTablePointer = rcwData.VTablePointer,
-                    Interfaces = pointers?.Select(r => new ComInterfaceEntry() { InterfacePointer = r.InterfacePointer, MethodTable = r.MethodTable }).ToArray() ?? Array.Empty<ComInterfaceEntry>()
+                    VTablePointer = rcwData.VTablePointer.ToAddress(_target),
+                    Interfaces = pointers?.Select(r => new ComInterfaceEntry() { InterfacePointer = r.InterfacePointer.ToAddress(_target), MethodTable = r.MethodTable.ToAddress(_target) }).ToArray() ?? Array.Empty<ComInterfaceEntry>()
                 };
             }
 
--- a/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
         public bool GetCcwInfo(ulong obj, out CcwInfo info)
         {
             info = default;
-            if (_sos.GetObjectData(ClrDataAddress.FromAddress(obj, _target), out ObjectData objData) &&
+            if (_sos.GetObjectData(ClrDataAddress.FromTargetAddress(obj, _target), out ObjectData objData) &&
                 !objData.CCW.IsNull &&
                 _sos.GetCCWData(objData.CCW, out CcwData ccwData))
             {
@@ -53,7 +53,7 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
         public bool GetRcwInfo(ulong obj, out RcwInfo info)
         {
             info = default;
-            if (_sos.GetObjectData(ClrDataAddress.FromAddress(obj, _target), out ObjectData objData) &&
+            if (_sos.GetObjectData(ClrDataAddress.FromTargetAddress(obj, _target), out ObjectData objData) &&
                 !objData.RCW.IsNull &&
                 _sos.GetRCWData(objData.RCW, out RcwData rcwData))
             {
--- a/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
@@ -33,11 +33,10 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
                 !objData.CCW.IsNull &&
                 _sos.GetCCWData(objData.CCW, out CcwData ccwData))
             {
-                ulong ccwAddr = objData.CCW.ToAddress(_target);
                 COMInterfacePointerData[]? pointers = _sos.GetCCWInterfaces(objData.CCW, ccwData.InterfaceCount);
                 info = new()
                 {
-                    Address = ccwAddr,
+                    Address = objData.CCW.ToAddress(_target),
                     IUnknown = ccwData.OuterIUnknown.ToAddress(_target),
                     Object = ccwData.ManagedObject.ToAddress(_target),
                     Handle = ccwData.Handle.ToAddress(_target),
@@ -57,11 +56,10 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
                 !objData.RCW.IsNull &&
                 _sos.GetRCWData(objData.RCW, out RcwData rcwData))
             {
-                ulong rcwAddr = objData.RCW.ToAddress(_target);
                 COMInterfacePointerData[]? pointers = _sos.GetRCWInterfaces(objData.RCW, rcwData.InterfaceCount);
                 info = new()
                 {
-                    Address = rcwAddr,
+                    Address = objData.RCW.ToAddress(_target),
                     IUnknown = rcwData.IUnknownPointer.ToAddress(_target),
                     Object = rcwData.ManagedObject.ToAddress(_target),
                     RefCount = rcwData.RefCount,
--- a/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacMethodLocator.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacMethodLocator.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Microsoft.Diagnostics.Runtime.AbstractDac;
--- a/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacMethodLocator.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacImplementation/DacMethodLocator.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Microsoft.Diagnostics.Runtime.AbstractDac;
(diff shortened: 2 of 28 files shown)


max-charlamb:
Follow-up to #1421 — we missed this case. `MinidumpMemoryDescriptor` uses `ClrDataAddress` for its `StartAddress` and `DataSize64` fields, which truncates 64-bit addresses on 32-bit hosts via `nint`/`nuint` casts. This wasn't caught in #1421 because those fields aren't `IntPtr.Size`-dependent — the truncation happens inside `ClrDataAddress`'s implicit conversions.

src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
```diff
@@ -13,34 +13,39 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
...
+        public DacComHelpers(SOSDac sos, int pointerSize)
         {
             _sos = sos;
+            _pointerSize = pointerSize;
         }
 
         public IEnumerable<ClrRcwCleanupData> EnumerateRcwCleanupData()
         {
-            return _sos.EnumerateRCWCleanup(0).Select(r => new ClrRcwCleanupData(r.Rcw, r.Context, r.Thread, r.IsFreeThreaded));
+            int pointerSize = _pointerSize;
```
max-charlamb: we don't need this local field if _pointerSize is a class field. This can be applied for the whole file.

leculver:
Let's separate out the `MinidumpMemoryDescriptor` into its own change.

This is a pretty messy change for something I don't really care if it works or not (32bit process debugging an x64 process).  I don't mind supporting it if it's something you all want, but I want to make sure it doesn't overcomplicate the code.

leculver:
This will be incredibly tricky to get right.

I think I'd like to see a more explicit converter class which we create once when the DacLibrary.  This converter can hang off of DacLibrary but should be passed to ClrDataProcess and all SosDac instances.  This way we have an encapsulated/explicit way of handling this conversion throughout.

**Danger**:  `structs` and CLRDATA_ADDRESS are not interchangable on 32bit linux ABI (x86 sysv and arm32, if I remember correctly).  These are passed by hidden pointer instead of in a register.  So if ClrDataAddress is ever returned from a function, you cannot use `ClrDataAddress` at that boundary point.  This also applies to ClrDataAddress as a parameter (every dac method) which is why it's defined as ulong in vtables.

**Danger**:  ClrDataAddress value parameters (which is basically every dac method) also expects a sign extended value at the boundary.  The current code works (probably by accident), so we have to be careful not to break it.  For example:

```C++
 inline TADDR CLRDATA_ADDRESS_TO_TADDR(CLRDATA_ADDRESS cdAddr)
 {
 #ifndef HOST_64BIT
     INT64 iSignedAddr = (INT64)cdAddr;
     if (iSignedAddr > INT_MAX || iSignedAddr < INT_MIN)
         DacError(E_INVALIDARG);
 #endif
     return (TADDR)cdAddr;
 }
```
 Walk 0x80000000 (a valid 3-GB-range 32-bit target address) through this:

   - Sign-extended input 0xFFFFFFFF80000000 -> signed -2147483648 -> in [INT_MIN, INT_MAX] -> truncates to 0x80000000. Works.
   - Un-sign-extended input 0x0000000080000000 -> signed +2147483648 -> > INT_MAX -> DacError(E_INVALIDARG). Fails.

Below is a rough sketch of what I'm thinking, though this is untested.  You can modify this design if you want, but the key points:

1. `ClrDataAddress` has no way to get the value out of it.  It's opaque except to `ClrDataAddress.Converter`.
2. An instance of the converter should be built once and passed around.
3. All structs which have a ClrDataAddress should define it as such.
4. Current inputs to legacy dac functions are all CLRDATA_ADDRESS and expect the sign extended value.
5. Values returned from the dac are ClrDataAddress and convert with ToAddress.
6. Values given to the dac need to be defined as ClrDataAddress somewhere to prevent mistakes, but pass through ToInteropAddress when passed to the raw method on the vtable.

(Or feel free to come up with something that works better.  Just keep in mind that touching this fragile system is a hot stove of pain.  You are more than welcome to tackle this, we just need to make sure it's right with the legacy dac and test it as best we can.)

```C#
/// <summary>
/// A wrapper around the raw 64-bit value produced by the DAC to represent an address. Due to historic/legacy
/// reasons, some DACs (legacy DACs) produce a sign-extended 64-bit value even when the target process is 32-bit.
/// This struct encapsulates that value and provides a converter to handle both legacy and modern DACs correctly.
/// </summary>
/// <param name="_value">The raw 64 bits produced by the DAC.  May be sign-extended from a  32-bit target address
/// when the backing DAC is a legacy DAC.</param>
[StructLayout(LayoutKind.Sequential)]
[DebuggerDisplay("{Value,h}")]
internal readonly struct ClrDataAddress(ulong _value) : IEquatable<ClrDataAddress>
{
    public bool IsNull => _value == 0;
    public bool Equals(ClrDataAddress other) => _value == other._value; 
    public override bool Equals(object? obj) => obj is ClrDataAddress cda && Equals(cda);
    public override int GetHashCode() => _value.GetHashCode();
    public static bool operator ==(ClrDataAddress left, ClrDataAddress right) => left.Equals(right);
    public static bool operator !=(ClrDataAddress left, ClrDataAddress right) => !left.Equals(right);
    public override string ToString() => $"0x{_value:x}";

    /// <summary>
    /// Constructs a ClrDataAddress.Converter for the given DAC configuration.  
    /// </summary>
    /// <param name="_signExtend">Should be false if cDac, true otherwise.</param>
    /// <param name="_pointerSize">The pointer size of the target process.</param>
    public sealed class Converter(bool _signExtend, int _pointerSize)
    {
        /// <summary>
        /// The target pointer size this converter was built for, in bytes.
        /// </summary>
        public int PointerSize => _pointerSize;

        /// <summary>
        /// True if this converter un-sign-extends legacy DAC output.  Exposed
        /// primarily for diagnostics/asserts; callers should normally just call
        /// <see cref="ToAddress"/> / <see cref="FromAddress"/>.
        /// </summary>
        public bool IsLegacy => _signExtend;

        /// <summary>
        /// Returns the un-sign-extended value of a ClrDataAddress to pass to a dac function.
        /// </summary>
        public ulong ToInteropAddress(ClrDataAddress cda) => cda._value;

        /// <summary>
        /// Converts a DAC-produced address to the real target address that
        /// the rest of ClrMD should see.
        /// </summary>
        public ulong ToAddress(ClrDataAddress cda)
        {
            if (!_signExtend)
                return cda.Value;

            return _pointerSize == 4 ? (uint)cda.Value : cda.Value;
        }

        /// <summary>
        /// Converts a canonical target address to the bit pattern the DAC
        /// expects back on its interface.
        /// </summary>
        public ClrDataAddress FromAddress(ulong address)
        {
            if (!_signExtend)
                return new ClrDataAddress(address);
                
            return _pointerSize == 4
                ? new ClrDataAddress(unchecked((ulong)(long)(int)address))
                : new ClrDataAddress(address);
        }
    }
}
```

max-charlamb:
> Let's separate out the `MinidumpMemoryDescriptor` into its own change.
> 
> This is a pretty messy change for something I don't really care if it works or not (32bit process debugging an x64 process). I don't mind supporting it if it's something you all want, but I want to make sure it doesn't overcomplicate the code.

Agreed that this is messy and that 32 bit processes debugging a 64 bit process is not an important scenario. The larger reason I want to fix this is supporting 64 bit processes debugging a 32 bit dump. Today it mostly works, but only when the 32 bit address is not in the top half of the address (and the top half of bits are filled with `1`s). The 32 bit host targeting a 64 bit process serves as a proxy for the more common scenario.

>Danger: structs and CLRDATA_ADDRESS are not interchangable on 32bit linux ABI (x86 sysv and arm32, if I remember correctly). These are passed by hidden pointer instead of in a register. So if ClrDataAddress is ever returned from a function, you cannot use ClrDataAddress at that boundary point. This also applies to ClrDataAddress as a parameter (every dac method) which is why it's defined as ulong in vtables.

We handle this in the cDAC by using custom source generated marshallers to marshal the `ClrDataAddress` object as a `ulong`. I can see if there is an equivalent that we can use that works on netstandard.

>Danger: ClrDataAddress value parameters (which is basically every dac method) also expects a sign extended value at the boundary. The current code works (probably by accident), so we have to be careful not to break it. 

The cDAC maintains the sign-extension of the legacy DAC APIs which use ClrDataAddress, so there shouldn't be any behavioral differences between legacy and 'modern' ISOSDacInterface APIs.

I appreciate the feedback. I'll rework this to try and make it cleaner using the converter approach. Thanks.

max-charlamb:
**Update:** Rebased onto latest `origin/main` (picks up #1424 high-bit test harness). Dropped one commit (`Fix 64-bit address truncation in MinidumpMemoryDescriptor`) as `patch already upstream`.

**New in this push:**
- `DacDataTargetCOM.ReadVirtual` (net6.0+): wrap `(long)address` in `unchecked` to avoid `OverflowException` in checked mode when the DAC passes a sign-extended 32-bit `CLRDATA_ADDRESS`. Surfaced by running `HeapTests.HighBit_HeapHasObjectsAbove2Gb` under x86.

**Validation:**
- Clean build (0 warnings, 0 errors) on `netstandard2.0` + `net10.0`.
- `HeapTests.HighBit_HeapHasObjectsAbove2Gb` **passes** under x86 HighBitHost.

src/Microsoft.Diagnostics.Runtime/DacInterface/ClrDataAddress.cs
```diff
@@ -1,42 +1,57 @@
...
-    /// Unfortunately this can cause issues when inspecting 32bit processes, since
-    /// if the highest bit is set the value will be sign-extended.  This struct is
-    /// meant to
+    /// When inspecting 32-bit processes the DAC sign-extends addresses with bit 31
+    /// set, so conversions to/from ulong must be done with the target pointer size
+    /// to correctly mask or sign-extend the value.
     /// </summary>
-    [DebuggerDisplay("{AsUInt64()}")]
-    internal readonly struct ClrDataAddress
+    [DebuggerDisplay("{Value}")]
```
leculver: Can you make this `Value,h`?  The signed decimal value won't be much help.

src/Microsoft.Diagnostics.Runtime/DacInterface/ClrDataAddress.cs
```diff
@@ -1,42 +1,57 @@
...
 
         /// <summary>
-        /// Returns the value of this address and un-sign extends the value if appropriate.
+        /// Creates a ClrDataAddress from a ulong, sign-extending for 32-bit targets.
         /// </summary>
-        /// <returns>The value of this address and un-sign extends the value if appropriate.</returns>
-        private ulong AsUInt64() => unchecked((nuint)Value);
+        /// <param name="address">The address value.</param>
+        /// <param name="pointerSize">The target process pointer size (4 or 8).</param>
+        public static ClrDataAddress FromAddress(ulong address, int pointerSize) => pointerSize == 4
```
leculver: Can you rename this `FromSignExtendedAddress`?

src/Microsoft.Diagnostics.Runtime/DacInterface/SosDac.cs
leculver: This one was missed, syncBlockCleanupPonter is a ClrDataAddress.

src/Microsoft.Diagnostics.Runtime/DacInterface/SosDac.cs
```diff
@@ -55,7 +55,7 @@ public RejitData[] GetRejitData(ulong md, ulong ip = 0)
             return Array.Empty<RejitData>();
         }
 
-        public HResult GetMethodDescData(ulong md, ulong ip, out MethodDescData data)
+        public HResult GetMethodDescData(ClrDataAddress md, ulong ip, out MethodDescData data)
```
leculver: `ulong ip` should be `ClrDataAddress`.

src/Microsoft.Diagnostics.Runtime/DacInterface/SosDac.cs
```diff
@@ -41,7 +41,7 @@ public SOSDac(DacLibrary lib, CallableCOMWrapper toClone) : base(toClone)
             _library = lib;
         }
 
-        public RejitData[] GetRejitData(ulong md, ulong ip = 0)
+        public RejitData[] GetRejitData(ClrDataAddress md, ulong ip = 0)
```
leculver: `ulong ip` should be `ClrDataAddress`.

src/Microsoft.Diagnostics.Runtime/DacInterface/SosDac.cs
leculver, line 791: Can you check all "struct *VTable" entries in ClrMD for any instance of `ulong`?  Those all will need to be reviewed for correctness.  That review would have caught `GetMethodDescData` above, for example.

src/Microsoft.Diagnostics.Runtime/DacInterface/SosDac8.cs
leculver: GetGenerationTableSrv/GetFinalizationFillPointersSvr were missed.

leculver (commented):
Just some pre-review things I found.

leculver:
Oh, I think SOSDac12.cs was entirely missed?

src/Microsoft.Diagnostics.Runtime/DacImplementation/DacMethodLocator.cs
```diff
@@ -1,55 +1,61 @@
...
+                    md = codeHeaderData.MethodDesc.ToAddress(_target);
             }
 
             return md;
         }
 
         public bool GetMethodInfo(ulong methodDesc, out MethodInfo methodInfo)
         {
-            if (!_sos.GetMethodDescData(methodDesc, 0, out MethodDescData mdd))
+            if (!_sos.GetMethodDescData(ClrDataAddress.FromAddress(methodDesc, _target), default, out MethodDescData mdd))
```
max-charlamb: Lets add a `ClrDataAddress.Null` and use that over default. I think it makes it more clear what is happening. We should replace any usage of default for ClrDataAddress with `ClrDataAddress.Null`.

src/Microsoft.Diagnostics.Runtime/DacInterface/SOSDac12.cs
```diff
@@ -15,22 +15,28 @@ internal sealed unsafe class SosDac12 : CallableCOMWrapper
     {
         internal static readonly Guid IID_ISOSDac12 = new("1b93bacc-8ca4-432d-943a-3e6e7ec0b0a3");
 
+        private readonly DacLibrary _library;
+
         public SosDac12(DacLibrary library, IntPtr ptr)
             : base(library?.OwningLibrary, IID_ISOSDac12, ptr)
         {
+            _library = library ?? throw new ArgumentNullException(nameof(library));
```
max-charlamb: Does this need to be a nullable check? The library doesn't seem to be nullable.

src/Microsoft.Diagnostics.Runtime/DacInterface/SOSDac12.cs
```diff
@@ -15,22 +15,28 @@ internal sealed unsafe class SosDac12 : CallableCOMWrapper
     {
         internal static readonly Guid IID_ISOSDac12 = new("1b93bacc-8ca4-432d-943a-3e6e7ec0b0a3");
 
+        private readonly DacLibrary _library;
```
max-charlamb: Lets store just the TargetProperties, not the DacLibrary to reduce the length of calls.

src/Microsoft.Diagnostics.Runtime/DacInterface/SOSDac12.cs
leculver: No, that's dead code from when this class was marked `public`.

leculver: Rather, you may see code like that around (and copilot might be copying it) but some of that is leftover from dead code, which might still be around.

src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
```diff
@@ -13,34 +13,37 @@ namespace Microsoft.Diagnostics.Runtime.DacImplementation
...
         public bool GetCcwInfo(ulong obj, out CcwInfo info)
         {
             info = default;
-            if (_sos.GetObjectData(obj, out ObjectData objData) &&
-                objData.CCW != 0 &&
+            if (_sos.GetObjectData(ClrDataAddress.FromTargetAddress(obj, _target), out ObjectData objData) &&
+                !objData.CCW.IsNull &&
                 _sos.GetCCWData(objData.CCW, out CcwData ccwData))
             {
+                ulong ccwAddr = objData.CCW.ToAddress(_target);
```
leculver: nit: Any reason this isn't inlined like the rest below?

src/Microsoft.Diagnostics.Runtime/DacImplementation/DacComHelpers.cs
```diff
@@ -50,21 +53,22 @@ public bool GetCcwInfo(ulong obj, out CcwInfo info)
         public bool GetRcwInfo(ulong obj, out RcwInfo info)
         {
             info = default;
-            if (_sos.GetObjectData(obj, out ObjectData objData) &&
-                objData.RCW != 0 &&
+            if (_sos.GetObjectData(ClrDataAddress.FromTargetAddress(obj, _target), out ObjectData objData) &&
+                !objData.RCW.IsNull &&
                 _sos.GetRCWData(objData.RCW, out RcwData rcwData))
             {
+                ulong rcwAddr = objData.RCW.ToAddress(_target);
```
leculver: nit: Any reason this isn't inlined like the rest below?

src/Microsoft.Diagnostics.Runtime/DacInterface/ClrDataAddress.cs
```diff
@@ -1,42 +1,158 @@
...
 
         /// <summary>
-        /// Creates an instance of ClrDataAddress.
+        /// Converts this <see cref="ClrDataAddress"/> returned from the DAC to a target-process
+        /// pointer, un-sign-extending the high 32 bits on 32-bit targets.
         /// </summary>
-        /// <param name="value"></param>
-        public ClrDataAddress(long value) => Value = value;
+        public ulong ToAddress(TargetProperties target) =>
+            target.PointerSize == 4 ? (ulong)(uint)_value : _value;
```
leculver: This should be unchecked:

 public ulong ToAddress(TargetProperties target) =>
     target.PointerSize == 4 ? unchecked((ulong)(uint)_value) : _value;

src/Microsoft.Diagnostics.Runtime/DacInterface/SosDac8.cs
```diff
@@ -15,22 +15,25 @@ internal sealed unsafe class SOSDac8 : CallableCOMWrapper
     {
         internal static readonly Guid IID_ISOSDac8 = new("c12f35a9-e55c-4520-a894-b3dc5165dfce");
 
+        private readonly DacLibrary _library;
+
         public SOSDac8(DacLibrary library, IntPtr ptr)
             : base(library?.OwningLibrary, IID_ISOSDac8, ptr)
         {
+            _library = library ?? throw new ArgumentNullException(nameof(library));
```
leculver: nit:  Don't need the ?? check.

leculver (commented):
One issue and three nitpicks.

It will likely be a day or two before I hit approve, I have to cleanly get the 4.0 change through.

leculver (dismissed):
LGTM. Thanks for the hard work on this one!

leculver (approved):

PR #1425, opened 2026-04-22
Fix GCRoot _seen set poisoning across FindPathFrom calls

leculver:
Bug: When WalkObject adds all children of an object to _seen before walking them, and then DFS returns early on a successful match, the unwalked children remain in _seen. Future FindPathFrom calls skip those objects, missing valid paths. This caused flaky FindAllPaths failures depending on heap layout.

Fix: After a successful search, drain remaining unprocessed children from the stack's ReferenceList buffers via Next() and remove them from _seen. This preserves the cross-call optimization for fully-walked dead-end subtrees (they stay in _seen correctly) while removing only the problematic entries.

Also fixes tests for .NET 10 compatibility:
- Test target now creates explicit GCHandle.Alloc roots so EnumerateRootPaths finds multiple independent paths regardless of runtime version (on .NET 10, statics live in a single Object[] handle and stack roots may not include local variables).
- ContainsPathsToTarget searches for source anywhere in the chain, not just the first element, since TheRoot may be nested under an Object[] root.
- GCRoot tests that call EnumerateRoots use [WindowsOrNet11Fact]: they run on Windows unconditionally, and on Linux/macOS only on .NET 11+ (the DAC SIGSEGV in EnumerateStackRoots on .NET 10 is fixed in .NET 11).

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime.Tests/src/Attributes/WindowsOrNet11FactAttribute.cs
+++ b/src/Microsoft.Diagnostics.Runtime.Tests/src/Attributes/WindowsOrNet11FactAttribute.cs
@@ -1,0 +1,25 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+using Xunit;
+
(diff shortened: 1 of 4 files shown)


brianrob (approved):
LGTM

PR #1431, opened 2026-04-28
Add lock free memory-mapped file data reader

leculver:
Add a single-threaded lock-free IDataReader built on MemoryMappedFile.  When setting DataTargetOptions.UseLockFreeMemoryMapReader = true, we will read data faster, but ClrRuntime.IsThreadSafe will return false, and it will no longer be thread safe.


src/Microsoft.Diagnostics.Runtime/DataReaders/Minidump/MinidumpReader.cs
```diff
@@ -104,5 +105,14 @@ public bool GetThreadContext(uint threadID, uint contextFlags, Span<byte> contex
...
         public bool ReadPointer(ulong address, out ulong value) => MemoryReader.ReadPointer(address, out value);
         public ulong ReadPointer(ulong address) => MemoryReader.ReadPointer(address);
+
+        public IReadOnlyList<DumpMemorySegment> EnumerateMemorySegments()
+        {
+            ImmutableArray<MinidumpSegment> segments = _minidump.Segments;
+            DumpMemorySegment[] result = new DumpMemorySegment[segments.Length];
+            for (int i = 0; i < segments.Length; i++)
+                result[i] = new DumpMemorySegment(segments[i].VirtualAddress, segments[i].FileOffset, segments[i].Size);
+            return result;
```
max-charlamb, line 115: Any reason not to sort here? It seems like this would be sorted, but there is a comment in the MachO reader about sorting again to be defensive.

max-charlamb (approved):

src/Microsoft.Diagnostics.Runtime/DataReaders/Minidump/MinidumpReader.cs
leculver: The segments are actually already sorted.  We also sort on the other end of this.

Issue #1334, opened 2025-10-22
`DataTarget.CreateSnapshotAndAttach` fails to disassemble generic type in Windows/CoreCLR


timcassell:
While working on https://github.com/dotnet/BenchmarkDotNet/pull/2843, I found that using `DataTarget.CreateSnapshotAndAttach` in the same process fails to disassemble properly only on Windows and CoreCLR (.Net 8 in my case). It works fine in .Net Framework, and it also works fine in Linux. Thankfully it works with `AttachToProcess` in Windows, so I can use that as a workaround, but I figured I should raise this issue.

I tried `3.1.512801` and `4.0.0-beta.24360.3`, same behavior on both.

The test that failed:

```cs
public class Generic<T> where T : new()
{
    [Benchmark]
    public T Create() => new T();
}

[Theory]
[MemberData(nameof(GetAllJits), DisableDiscoveryEnumeration = true)]
[Trait(Constants.Category, Constants.BackwardCompatibilityCategory)]
public void CanDisassembleGenericTypes(Jit jit, Platform platform, IToolchain toolchain)
{
    if (OsDetector.IsMacOS()) return; // currently not supported

    var disassemblyDiagnoser = new DisassemblyDiagnoser(
        new DisassemblyDiagnoserConfig(printSource: true, maxDepth: 3));

    CanExecute<Generic<int>>(CreateConfig(jit, platform, toolchain, disassemblyDiagnoser, RunStrategy.Monitoring));

    var result = disassemblyDiagnoser.Results.Values.Single();

    Assert.Empty(result.Errors);
    Assert.Contains(result.Methods, method => method.Maps.Any(map => map.SourceCodes.OfType<Asm>().Any()));
}
```

https://github.com/dotnet/BenchmarkDotNet/blob/1c3faa21ab9aac8aab277e1fd19e151768c57cba/tests/BenchmarkDotNet.IntegrationTests/DisassemblyDiagnoserTests.cs#L136-L159

leculver:
I think the `CreateSnapshotAndAttach` portion of this bug is a red herring.  I dug into what BenchmarkDotNet is doing, and it's calling FindTypeByName with a generic class...which never would have worked.  This is the code in question:

```C#
var typeWithBenchmark = state.Runtime.EnumerateModules()
    .Select(module => module.GetTypeByName(args.TypeName))
    .WhereNotNull().First();
```

[...]

PR #1436, opened 2026-04-30
Throw InvalidOperationException when AttachToProcess targets the current process

leculver:
Self-attaching with AttachToProcess leads to crashes (SIGSEGV) because the DAC reads process memory that is concurrently mutating. The stack walker encounters stale or null frame pointers, causing null dereferences in libmscordaccore.so.

Users who need to inspect the current process should use CreateSnapshotAndAttach which takes a consistent snapshot first.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/DataTarget.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DataTarget.cs
@@ -374,6 +374,16 @@ namespace Microsoft.Diagnostics.Runtime
         /// <returns>A <see cref="DataTarget"/> instance.</returns>
         public static DataTarget AttachToProcess(int processId, bool suspend, DataTargetOptions? options = null)
         {
+#if NET6_0_OR_GREATER
+            int currentPid = Environment.ProcessId;
+#else
+            int currentPid;
+            using (Process p = Process.GetCurrentProcess())
+                currentPid = p.Id;
+#endif
+            if (processId == currentPid)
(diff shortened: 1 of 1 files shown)


filzrev:
> Self-attaching with AttachToProcess leads to crashes (SIGSEGV) 

Is this restriction also applied to Windows platform?

**Background**
Currently BenchmarkDotNet using `AttachToProcess` on same process when running on .NET Core.
It's used to avoid issue #1334.
After this PR is merged, this workaround code is no longer works.

https://github.com/dotnet/BenchmarkDotNet/blob/ac0b009bc95159c04957837225f7fd7651a82ef9/src/BenchmarkDotNet/Disassemblers/ClrMdDisassembler.cs#L44-L50

leculver:
That really isn't safe to do.  Any change to the runtime is a landmine waiting to go off which can lead to access violations (sigsegv) and other issues. On windows, we will silently swallow access violations, but that doesn't mean it's good to generate them.

Let me see if I can figure out why snapshots don't work in that scenario.  I'll hold off merging this for now.

leculver:
@filzrev Ok, I think I have a fix for issue #1334, see the comment in that bug.  I _will_ be merging this change to disable self-attach, but it won't be in next ClrMD nuget package update, but rather the one after that.  That will give you a version to test on which both has the fix for that issue and doesn't disallow self-attach.

max-charlamb (approved):

PR #1438, opened 2026-04-30
Fix #1334: GetTypeByName resolves closed generics and reflection FullNames

### LINKED ISSUE #1334: `DataTarget.CreateSnapshotAndAttach` fails to disassemble generic type in Windows/CoreCLR

leculver:
ClrModule.GetTypeByName(name) was unable to find closed generic instantiations such as Generic<int>, because the per-module overload only walks the module's TypeDef->MethodTable map and closed generic MTs are not present in any module's metadata (only the open generic typedef is). Additionally, callers naturally reach for Type.FullName which uses the assembly-qualified format 'Foo`1[[System.Int32, mscorlib, ...]]' rather than ClrType.Name's 'Foo<System.Int32>' format, causing string-equality comparisons to fail even when the type is otherwise findable.

This commit:

- Adds a closed-generic fallback to ClrHeap.GetTypeByName(ClrModule, string) that walks the heap (or already-cached types) for a constructed instantiation in the requested module when the TypeDef-map scan misses.
- Adds an analogous fallback to ClrHeap.GetTypeByName(string).
- Adds TryNormalizeReflectionFullName() which converts assembly-qualified reflection-style names to ClrType.Name format (recursively, supporting nested generics).
- Gates both fallbacks with LooksLikeClosedGenericName() so the (more expensive) heap walk only runs when the input plausibly refers to a generic; non-generic and previously-found types take the original fast path with no behavior change.
- Adds EnumerateCachedTypes() to ClrTypeFactory for the cache-only fast path before falling back to a full heap walk.

This unblocks BenchmarkDotNet's ClrMdDisassembler self-attach scenario (see dotnet/BenchmarkDotNet ClrMdDisassembler.cs:81), which calls module.GetTypeByName(args.BenchmarkInstance.GetType().FullName!) and was silently throwing InvalidOperationException via the subsequent .First() on an empty sequence.

Tests: adds 5 new tests under GenericFieldTests covering the FullName-input case and the normalizer's recursive parsing. All 506 tests pass.

Fixes: #1334

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
@@ -1223,6 +1223,20 @@ namespace Microsoft.Diagnostics.Runtime
             // in any module's TypeDef map but were constructed during heap enumeration.
             result ??= _typeFactory.TryGetCachedTypeByName(name);
 
+            // Closed-generic fallback: if the caller passed a reflection-style name like
+            // "Generic`1[[System.Int32, mscorlib, ...]]" (i.e. Type.FullName), normalize
+            // to ClrType.Name format ("Generic<System.Int32>") and try again.
+            if (result is null && LooksLikeClosedGenericName(name))
+            {
+                string? normalized = TryNormalizeReflectionFullName(name);
+                if (normalized != null && normalized != name)
+                {
+                    result = _typeFactory.TryGetCachedTypeByName(normalized);
+                    if (result is null)
+                        result = FindClosedGenericInAnyModule(normalized);
+                }
+            }
+
             return result;
         }
 
@@ -1234,7 +1248,19 @@ namespace Microsoft.Diagnostics.Runtime
             if (name.Length == 0)
                 throw new ArgumentException($"{nameof(name)} cannot be empty");
 
-            return FindTypeName(module.Address, module.EnumerateTypeDefToMethodTableMap(), name);
+            ClrType? result = FindTypeName(module.Address, module.EnumerateTypeDefToMethodTableMap(), name);
+
+            // Closed-generic fallback: closed generic instantiations are not in any module's
+            // TypeDef map (only the open generic typedef is).  If the caller asked for a name
+            // that looks like a closed generic, normalize it and walk the heap for a matching
+            // constructed instantiation in this module.
+            if (result is null && LooksLikeClosedGenericName(name))
+            {
+                string? normalized = TryNormalizeReflectionFullName(name) ?? name;
+                result = FindClosedGenericInModule(module.Address, normalized);
+            }
+
+            return result;
         }
 
         private ClrType? FindTypeName(ulong moduleAddress, IEnumerable<(ulong MethodTable, int Token)> map, string name)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Runtime/ClrHeap.cs
```diff
@@ -1262,6 +1288,150 @@ private IEnumerable<ClrRoot> EnumerateFinalizers(IEnumerable<MemoryRange> memory
...
+        }
+
+        /// <summary>
+        /// Converts a reflection-style assembly-qualified generic name like
+        /// "Foo`1[[System.Int32, System.Private.CoreLib, Version=...]]" into the
+        /// simplified ClrType.Name format "Foo&lt;System.Int32&gt;".  Returns null if the
+        /// input is not in a recognizable reflection generic form (in which case the
+        /// caller should treat the input as already simplified).
+        /// </summary>
+        internal static string? TryNormalizeReflectionFullName(string name)
```
max-charlamb, line 1310: It looks like this method doesn't properly handle types like:
```
typeof(List<int>[]).FullName
// System.Collections.Generic.List`1[[System.Int32, System.Private.CoreLib, Version=10.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]][]
```

The trailing `[]` is lost.

leculver: Good catch, that's now fixed.

max-charlamb (approved):

PR #1440, opened 2026-04-30
Add IThreadInfoReader, IProcessInfoProvider, IMemoryRegionReader

leculver:
Expose minidump/coredump auxiliary data via three new opt-in 'hidden' interfaces in the DataReaders.Implementation namespace:

- IThreadInfoReader: per-thread OS data (CPU times, stack range, priority, suspend count, create/exit times) parsed from minidump ThreadInfoListStream / ThreadListStream / ThreadExListStream and ELF prstatus notes.
- IProcessInfoProvider: process-wide metadata (PID, image path, OS version, processor count, dump timestamp, aggregate user/kernel times) from minidump MiscInfoStream/SystemInfoStream and ELF prpsinfo / prstatus aggregation.
- IMemoryRegionReader: full process VAD/VMA map. Minidump path parses MemoryInfoListStream into cross-platform State/Type/Protect enums; coredump path maps every PT_LOAD program header, classifying as Image when the range falls inside a known ElfLoadedImage.

MinidumpReader and CoredumpReader implement all three. LockFreeMmf- DataReader passes them through to the wrapped inner reader.

Adds 20 tests under DumpAuxiliaryDataReaderTests covering both minidump and coredump paths (the same LoadFullDump test target produces a coredump on Linux), plus a passthrough test for the LockFreeMmf wrapper. Full suite: 519 passed / 5 skipped.

Also fixes a pre-existing bug in Minidump.ReadThreadData where the foreach iterated the unfiltered _directories list, causing the ThreadInfoListStream branch to be unreachable.


leculver:
@max-charlamb This adds some "hidden" interfaces to ClrMD similar to how we do this for SOS.  These interfaces allow some more-direct access to data that's in the Minidump/ElfCoreFile/MachoCoreDump without having to expose those classes publicly.  I don't want to expose the underlying dump readers as public, because then I'd have to maintain them. :)

max-charlamb (approved):

PR #1449, opened 2026-05-08
Add stresslog support

leculver:
This pull request adds comprehensive infrastructure for stress log integration testing in the diagnostics runtime tests. It introduces new synthetic test helpers for stress log data, a suite of integration tests to validate stress log parsing across different runtime configurations, and extends dump generation utilities to allow custom environment variables, enabling more flexible test scenarios.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/StressLog/Internal/StressLogLayout.cs
+++ b/src/Microsoft.Diagnostics.Runtime/StressLog/Internal/StressLogLayout.cs
@@ -3,134 +3,398 @@
 
 using System;
 using System.Buffers.Binary;
+using Microsoft.Diagnostics.Runtime;
 
 namespace Microsoft.Diagnostics.Runtime.StressLogs.Internal
 {
     /// <summary>
-    /// Field offsets within the on-target structures we read. Only 64-bit
-    /// targets are supported; 32-bit targets are rejected during open.
+    /// Per-target description of the on-disk struct layouts that make up a
+    /// runtime stress log. A single instance captures pointer size, OS, and
+    /// runtime variant (Core vs .NET Framework V1) so that consumers can
+    /// read fields by symbolic name regardless of the target.
     /// </summary>
     /// <remarks>
-    /// The runtime's <c>StressLog</c> class layout is documented at
-    /// <c>src/coreclr/inc/stresslog.h</c> and mirrored in this repo's
-    /// adjacent dotnet/diagnostics tree. These offsets must move in
-    /// lockstep with that header.
+    /// <para>
+    /// The bit-field decoders (<see cref="DecodeMessageV3"/>,
+    /// <see cref="DecodeMessageV4"/>, <see cref="DecodeMessageFxV1"/>) are
+    /// independent of pointer size and remain static methods on this type.
+    /// </para>
+    /// <para>
+    /// All pointer-sized field reads MUST go through
+    /// <see cref="ReadTargetPointer"/>. Reads done with
+    /// <c>BinaryPrimitives.ReadInt32LittleEndian</c> would sign-extend a
+    /// high-bit (≥ 0x80000000) 32-bit pointer, producing a malformed
+    /// 64-bit address; reads done with <c>ReadUInt64LittleEndian</c> on a
+    /// 32-bit target would consume four bytes of an unrelated adjacent
+    /// field. Centralizing in a single helper avoids both classes of bug.
+    /// </para>
     /// </remarks>
-    internal static class StressLogLayout
+    internal sealed class StressLogLayout
     {
-        // -----------------------------------------------------------------
-        // In-process StressLog (64-bit target)
-        // -----------------------------------------------------------------
+        /// <summary>Pointer size on the target, in bytes (4 or 8).</summary>
+        public int PointerSize { get; }
 
-        public const int InProc_FacilitiesToLog   = 0;     // uint
-        public const int InProc_LevelToLog        = 4;     // uint
-        public const int InProc_MaxSizePerThread  = 8;     // uint
-        public const int InProc_MaxSizeTotal      = 12;    // uint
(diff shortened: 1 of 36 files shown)


src/Microsoft.Diagnostics.Runtime/StressLog/StressLog.cs
```diff
@@ -0,0 +1,523 @@
...
+
+        /// <summary>
+        /// Try to open the stress log for the given <paramref name="runtime"/>.
+        /// Looks up the stress log address through the runtime's DAC and
+        /// reads from the runtime's data target. Returns <see langword="false"/>
+        /// if the runtime does not have stress logging enabled, the DAC does
+        /// not support <c>GetStressLogAddress</c>, or the resulting log fails
+        /// validation.
+        /// </summary>
+        internal static bool TryOpen(ClrRuntime runtime, out StressLog? stressLog)
```
brianrob: Looks like this doesn't have any callers?

src/Microsoft.Diagnostics.Runtime/StressLog/StressLog.cs
```diff
@@ -0,0 +1,523 @@
...
+            => TryOpen(runtime, out stressLog, out _);
+
+        /// <summary>
+        /// Try to open the stress log for the given <paramref name="runtime"/>,
+        /// returning a human-readable explanation in
+        /// <paramref name="failureReason"/> when the open fails.
+        /// </summary>
+        internal static bool TryOpen(ClrRuntime runtime, out StressLog? stressLog, out string? failureReason)
```
brianrob, line 112: There are a bunch of `TryOpen` overloads that all seem pretty close and are all internal.  Do you need all of these?

src/Microsoft.Diagnostics.Runtime/StressLog/StressLog.cs
```diff
@@ -0,0 +1,523 @@
...
+
+                // Yield in newest-first order via repeated linear max scan.
+                // Iterator count is bounded by MaxThreads; the linear scan is
+                // O(threads) per message, which is fine for typical numbers.
+                while (iterators.Count > 0)
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+
+                    int newest = 0;
+                    for (int i = 1; i < iterators.Count; i++)
```
brianrob, line 377: If the perf of this is currently fine, then this is fine.  Otherwise, you might consider a minheap here to avoid the O(n^2) cost.

src/Microsoft.Diagnostics.Runtime/StressLog/Internal/StressLogLayout.cs
```diff
@@ -0,0 +1,488 @@
...
+        /// Layout for a 64-bit Core target. Matches modern CoreCLR.
+        /// </summary>
+        public static StressLogLayout CoreX64 { get; } = new StressLogLayout(pointerSize: 8, isFramework: false);
+
+        /// <summary>
+        /// Layout for a 64-bit .NET Framework V1 target. Same chunk and
+        /// module-table sizes as Core x64; ThreadStressLog and StressMsg
+        /// layouts differ.
+        /// </summary>
+        public static StressLogLayout FrameworkX64 { get; } = new StressLogLayout(pointerSize: 8, isFramework: true);
```
brianrob: FrameworkWinX64?

src/Microsoft.Diagnostics.Runtime/StressLog/Internal/StressLogLayout.cs
```diff
@@ -0,0 +1,488 @@
...
+        /// preserving pointer size.
+        /// </summary>
+        public StressLogLayout WithFramework() => PointerSize switch
+        {
+            8 => FrameworkX64,
+            4 => FrameworkWinX86,
+            _ => throw new InvalidOperationException(),
+        };
+
+        private StressLogLayout(int pointerSize, bool isFramework)
```
brianrob, line 230: Not required, but does it make sense to generate structs for each of the layouts that you can just read directly into, rather than having a Layout that you then read individual field offsets from?

src/Microsoft.Diagnostics.Runtime/StressLog/Internal/StressLogSanitizer.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (dst.Length < srcCount)
+                throw new ArgumentException("destination buffer is too small", nameof(dst));
+
+            for (int i = 0; i < srcCount; i++)
+            {
+                ushort u = (ushort)(src[i * 2] | (src[i * 2 + 1] << 8));
+                if (u == 0)
+                    return i;
+
+                dst[i] = u <= 0x7F ? SanitizeByte((byte)u) : Replacement;
```
brianrob, line 83: This comparison doesn't seem to match the comparison in `SanitizeByte`.  Is that intentional?

brianrob (dismissed):
Overall, LGTM.  A few questions - you don't need to consider any of these blocking unless you want to.

src/Microsoft.Diagnostics.Runtime/StressLog/StressLog.cs
leculver: Thanks, these TryOpen were public but I swapped up the interface to this code and forgot to clean this up.

src/Microsoft.Diagnostics.Runtime/StressLog/StressLog.cs
leculver: Perf is ok here, and we are bound by MaxThreads to not blow up costs.

src/Microsoft.Diagnostics.Runtime/StressLog/Internal/StressLogSanitizer.cs
leculver: This is intentional, I updated the comments.

brianrob (approved):

Issue #420, opened 2019-12-12
IDataReader implementations need to be thread safe


leculver:
Need to make CoreDumpReader and LinuxLiveDataReader thread safe.

leculver:
Also need to validate that PR https://github.com/microsoft/clrmd/commit/54b64b91554873fff671156cf7cd0b5b34bed1c2 is thread safe.

NextTurn:
> Also need to validate that PR [54b64b9](https://github.com/microsoft/clrmd/commit/54b64b91554873fff671156cf7cd0b5b34bed1c2) is thread safe.

There is an internal lock in S.R.PE.

https://github.com/dotnet/runtime/blob/33bedaf3bcc95d91dde5f09251a5972fbac5f05e/src/libraries/System.Reflection.Metadata/src/System/Reflection/Internal/MemoryBlocks/StreamMemoryBlockProvider.cs#L44
https://github.com/dotnet/runtime/blob/33bedaf3bcc95d91dde5f09251a5972fbac5f05e/src/libraries/System.Reflection.Metadata/src/System/Reflection/Internal/MemoryBlocks/StreamMemoryBlockProvider.cs#L121

leculver:
I will likely not tackle thread-safety of Linux data readers before RTM.  Moving this to "future".

PR #1471, opened 2026-06-03
Make dump reader caches safe for concurrent reads

leculver:
## Summary

Makes the file-backed dump reader cache layer safe for concurrent readers without changing DAC behavior.

This PR fixes managed cache publication in the dump-reader/cache-entry layer:

- publishes CoreDumpReader and ELF-derived lazy state with atomic/locked first-publisher patterns
- makes Windows cached reader pages immutable once published
- returns only ArrayPool buffers that lost publication and were never visible to readers
- avoids returning published evicted buffers to ArrayPool while in-flight readers might still hold references

Part of microsoft/clrmd#420.

## Validation

- Built as part of the split branch verification: `dotnet build src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj --framework net10.0` -- passed with 0 warnings and 0 errors.
- This split branch was recombined with the other reduced branches and compared against `lockfree-mmf-threadsafe`; all non-stress paths matched the original combined branch.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/Linux/ElfCoreFile.cs
+++ b/src/Microsoft.Diagnostics.Runtime/Linux/ElfCoreFile.cs
@@ -8,6 +8,7 @@ using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading;
 
 namespace Microsoft.Diagnostics.Runtime.Utilities
 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Runtime/Linux/ElfCoreFile.cs
```diff
@@ -124,38 +145,54 @@ private IEnumerable<ElfNote> GetNotes(ElfNoteType type)
...
+            if (_auxvLoaded)
                 return;
 
-            ElfNote auxvNote = GetNotes(ElfNoteType.Aux).SingleOrDefault() ?? throw new BadImageFormatException($"No auxv entries in coredump");
-            ulong position = 0;
-            int count = 0;
-            while (true)
+            // Hold the lock for the full read+populate so concurrent first callers do
+            // not corrupt _auxvEntries via interleaved Add calls.
+            lock (_auxvLock)
```
brianrob, line 153: It seems like there is a mix of `lock` and `Interlocked.CompareExchange`.  Is this intentional?

brianrob (approved):
One non-blocking question below.

src/Microsoft.Diagnostics.Runtime/Linux/ElfCoreFile.cs
leculver: This is intentional.  Interlocked better fit the functions where it's used, and a full lock was the better fit in LoadAuxvTable.  I'm not trying to build a fully lockless set of classes, but the code above didn't need a full lock to be correct.

brianrob: Sounds good.

PR #1473, opened 2026-06-03
Serialize DAC access and materialize unsafe enumerations

leculver:
## Summary

Introduces and applies a single DAC serialization model for stateful DAC entry points.

The DAC is not safe to call concurrently across independent helpers/enumerators. This PR adds and applies the `DacLibrary`/`SOSDac` `SyncRoot` convention so DAC calls that can race native state are serialized and results are materialized before returning to callers.

Included here:

- shared DAC/SOS `SyncRoot` plumbing and documentation
- service singleton construction and DAC flush serialization
- heap service initialization under the DAC lock
- runtime/heap/thread cache population protection so partial native enumerations cannot be permanently cached
- stack walking materialized under the DAC lock with `maxFrames` enforced inside the DAC walk
- remaining DAC helper surfaces serialized through `SyncRoot`
- native heap traversal materialized under the lock instead of returning lazy DAC-backed enumerables

Part of microsoft/clrmd#420.

## Validation

- Built as part of the split branch verification: `dotnet build src/Microsoft.Diagnostics.Runtime/Microsoft.Diagnostics.Runtime.csproj --framework net10.0` -- passed with 0 warnings and 0 errors.
- This split branch was recombined with the other reduced branches and compared against `lockfree-mmf-threadsafe`; all non-stress paths matched the original combined branch.

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
+++ b/src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
@@ -26,6 +26,7 @@ namespace Microsoft.Diagnostics.Runtime
         private volatile ClrHeap? _heap;
         private ImmutableArray<ClrThread> _threads;
         private volatile DomainAndModules? _domainAndModules;
+        private ImmutableArray<ClrHandle> _handles;
 
         private IAbstractRuntime? _runtime;
         private IAbstractComHelpers? _comHelpers;
@@ -36,6 +37,7 @@ namespace Microsoft.Diagnostics.Runtime
         private string? _stressLogFailureReason;
         private bool _stressLogProbed;
         private readonly object _stressLogGate = new();
+        private readonly object _runtimeCacheLock = new();
 
         internal ClrRuntime(ClrInfo clrInfo, IServiceProvider services)
         {
@@ -61,6 +63,7 @@ namespace Microsoft.Diagnostics.Runtime
             _controller.Flush();
             _domainAndModules = null;
             _threads = default;
+            _handles = default;
             _heap = null;
 
             lock (_stressLogGate)
@@ -208,22 +211,29 @@ namespace Microsoft.Diagnostics.Runtime
                 if (!_threads.IsDefault)
                     return _threads;
 
-                IAbstractThreadHelpers? threadHelpers = GetService<IAbstractThreadHelpers>();
-                ImmutableArray<ClrThread>.Builder builder = ImmutableArray.CreateBuilder<ClrThread>();
-
-                int max = DataTarget.Options.Limits.MaxThreads;
-                foreach (ClrThreadInfo data in GetDacRuntime().EnumerateThreads())
+                // Serialize cache population: concurrent DAC EnumerateThreads calls can
+                // truncate each other and the CAS winner's partial result would become
+                // the permanent cached value.
+                lock (_runtimeCacheLock)
                 {
-                    builder.Add(new ClrThread(DataTarget.DataReader, this, threadHelpers, data));
+                    if (!_threads.IsDefault)
+                        return _threads;
 
-                    if (max-- == 0)
-                        break;
+                    IAbstractThreadHelpers? threadHelpers = GetService<IAbstractThreadHelpers>();
+                    ImmutableArray<ClrThread>.Builder builder = ImmutableArray.CreateBuilder<ClrThread>();
+
+                    int max = DataTarget.Options.Limits.MaxThreads;
+                    foreach (ClrThreadInfo data in GetDacRuntime().EnumerateThreads())
+                    {
+                        if (builder.Count >= max)
+                            break;
+
+                        builder.Add(new ClrThread(DataTarget.DataReader, this, threadHelpers, data));
+                    }
+
+                    _threads = builder.MoveOrCopyToImmutable();
+                    return _threads;
                 }
-
-                ImmutableArray<ClrThread> threads = builder.MoveOrCopyToImmutable();
-                ImmutableInterlocked.InterlockedCompareExchange(ref _threads, threads, _threads);
-
-                return _threads;
             }
         }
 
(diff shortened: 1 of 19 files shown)


src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
```diff
@@ -275,7 +285,28 @@ public ImmutableArray<ClrThread> Threads
...
+            lock (_runtimeCacheLock)
+            {
+                if (!_handles.IsDefault)
+                    return _handles;
+
+                ImmutableArray<ClrHandle>.Builder builder = ImmutableArray.CreateBuilder<ClrHandle>();
+                foreach (ClrHandleInfo info in GetDacRuntime().EnumerateHandles())
+                    builder.Add(new ClrHandle(this, info));
+
+                _handles = builder.MoveOrCopyToImmutable();
```
brianrob, line 322: This wasn't previously cached - is caching it intentional?

src/Microsoft.Diagnostics.Runtime/ClrThread.cs
```diff
@@ -244,29 +261,33 @@ public IEnumerable<ClrStackFrame> EnumerateStackTrace(bool includeContext, int m
...
-                yield return frame;
+                cache = _frameCache;
+                if (cache is not null && (!includeContext || cache.IncludedContext))
+                    return cache.Elements.Length <= maxFrames ? Array.AsReadOnly(cache.Elements) : cache.Elements.Take(maxFrames);
+
+                ClrStackFrame[] frames = threadHelpers.EnumerateStackTrace(OSThreadId, includeContext, maxFrames, traceErrors: IsAlive)
+                    .Select(r => CreateClrStackFrame(r))
+                    .Take(maxFrames)
+                    .ToArray();
+                if (frames.Length < maxFrames)
```
brianrob, line 286: I'm a bit confused on this line.  Is the goal to only cache if frames is less than maxFrames?  It can't be more than maxFrames because of the call to `Take(maxFrames)`.

src/Microsoft.Diagnostics.Runtime/ClrThread.cs
```diff
@@ -127,26 +129,32 @@ public IEnumerable<ClrStackRoot> EnumerateStackRoots()
...
-        private IEnumerable<ClrStackRoot> CacheAndReturnRoots(IEnumerable<ClrStackRoot> roots)
-        {
-            List<ClrStackRoot> cache = new();
-            foreach (ClrStackRoot root in roots)
             {
-                cache.Add(root);
-                yield return root;
+                ClrHeap heap = Runtime.Heap;
+                ClrStackFrame[] frames = GetFramesForRoots();
+                return threadHelpers.EnumerateStackRoots(OSThreadId, traceErrors: IsAlive).Select(r => CreateClrStackRoot(heap, frames, r)).Where(r => r is not null)!;
```
brianrob, line 136: Does the non-caching path need to be called under the `_rootCacheLock` as well?

src/Microsoft.Diagnostics.Runtime/DacImplementation/DacServiceProvider.cs
```diff
@@ -75,44 +76,111 @@ public void Dispose()
...
             {
                 IAbstractHeap? heap = _heapHelper;
                 if (heap is not null)
                     return heap;
 
-                if (_sos.GetGCHeapData(out GCInfo data) && _sos.GetCommonMethodTables(out CommonMethodTables mts) && !mts.ObjectMethodTable.IsNull)
-                    return _heapHelper = new DacHeap(_sos, _sos8, _sos12, _sos16, _dataReader, _dac.TargetProperties, data, mts);
-
-                return null;
+                lock (_serviceLock)
```
brianrob, line 111: I'm noticing that `_serviceLock` is distinct from `_dac.SyncRoot` and `_sos.SyncRoot`.  Just wanting to make sure that (a) we shouldn't be taking `_sos.SyncRoot` instead of `_dac.SyncRoot` since you call `_sos.GetGCHeapData` under the DAC SyncRoot, and (b) just wondering why you need an additional `_serviceLock` here.

brianrob (dismissed):
Had a few questions below, but if all is good, feel free to merge.

src/Microsoft.Diagnostics.Runtime/ClrThread.cs
leculver: This is correct because the if statement has `frames.Length < maxFrames`, not `maxFrames < frames.Length`.  The result is we ONLY caches the result if we are sure we got the complete result, not if it was truncated.

For example, we requested 16 frames and got 8, we can safely cache it because we know the result is complete.  If we requested 16 frames and got 16 back...we don't know if this was a full request or if it was truncated, so don't cache it...the caller may call back and request 32 frames and we don't want to hand back 16 frames not knowing if there were 17 in the list.

src/Microsoft.Diagnostics.Runtime/ClrThread.cs
leculver: No, because the dac enumeration itself is properly serialized.  The `_rootCache` lock is only to make sure we properly cache and publish a complete `_rootCache`.

src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
leculver: Yes, walking the handle table is expensive so I wanted to cache it.  But this did remind me that there's a correctness issue with Flush.  Using a volatile publish will fix this.  It's ugly, but it's fast.  I wish I hadn't chosen to use ImmutableArray so long ago.

src/Microsoft.Diagnostics.Runtime/DacImplementation/DacServiceProvider.cs
leculver: Service lock is intentional, as it's distinct for singleton publishing here.

_dac.SyncRoot and _sos.SyncRoot are the same object, so that isn't an issue, but it is confusing. I changed some of these over to make it more consistent in the code so it's easier to understand.

src/Microsoft.Diagnostics.Runtime/ClrRuntime.cs
brianrob: Sounds good.

src/Microsoft.Diagnostics.Runtime/ClrThread.cs
brianrob: Ah, thanks for clarifying.  I missed that.

src/Microsoft.Diagnostics.Runtime/ClrThread.cs
brianrob: Sounds good.

src/Microsoft.Diagnostics.Runtime/DacImplementation/DacServiceProvider.cs
brianrob: Sounds good.

brianrob (approved):

PR #1478, opened 2026-06-10
Delegate DAC signature verification decision to a predicate

hoyosjs:
Add a new option on the `DataTargetOptions` to allow hosts to control, on a per-runtime basis, whether ClrMD verifies the signature of a DAC before loading it. This provides more flexibility and security for hosts that may have their own trust mechanisms or need to bypass verification for certain DACs they have procured securely. The override takes priority over `VerifyDacOnWindows`.


leculver (approved):
Looks good to me.

PR #1482, opened 2026-06-12
Add IClrSymbolProvider hook + enable external IClrInfoProvider impls

max-charlamb:
Two related additions so external IClrInfoProvider implementations can drive a (c)DAC over ClrMD's standard plumbing, and so the (c)DAC can call back into a host-supplied symbol resolver:

**External IClrInfoProvider plumbing**

  * ClrInfo: ctor internal -> public. External providers need to construct a ClrInfo to return from ProvideClrInfoForModule.

  * DataTarget.CreateDacServices(ClrInfo, dacFilePath, verifySignature = false): public static factory that loads the DAC and wraps it in the standard IServiceProvider that ClrRuntime consumes. Avoids cascading internal types (DacLibrary, DacServiceProvider, DacDataTarget, ClrDataProcess, SOSDac*, ...) into the public surface.

**Host symbol provider hook**

  * Adds public IClrSymbolProvider that hosts (SOS / dotnet-dump) register via DataTargetOptions.SymbolProvider. ClrMD's COM data target wrapper always exposes ICLRSymbolProvider; vtable thunks forward to the registered provider and return E_FAIL when none is set.

  * ClrMD on netstandard2.0 has no PDB reader of its own, but hosts often do (dbghelp, symbol service, remote symbol server). Without this hook a (c)DAC loaded via ClrMD can't resolve addresses to module-qualified names.

  * Wired symmetrically on both CCW paths: netstandard2.0 / .NET Framework hosts (LegacyDacDataTargetWrapper / COMCallableIUnknown + VTableBuilder) and NET6+ hosts (DacDataTargetCOM / ComWrappers).

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
+++ b/src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
@@ -43,14 +43,14 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             wrappers[1].IID = DacDataTarget.IID_IMetadataLocator;
             wrappers[1].Vtable = IMetaDataLocatorVtbl.Create(qi, addRef, release);
 
-            wrappers[2].IID = DacDataTarget.IID_ICLRRuntimeLocator;
-            wrappers[2].Vtable = ICLRRuntimeLocatorVtbl.Create(qi, addRef, release);
+            wrappers[2].IID = DacDataTarget.IID_ICLRSymbolProvider;
+            wrappers[2].Vtable = ICLRSymbolProviderVtbl.Create(qi, addRef, release);
 
-            wrappers[3].IID = DacDataTarget.IID_ICLRContractLocator;
-            wrappers[3].Vtable = ICLRContractLocatorVtbl.Create(qi, addRef, release);
+            wrappers[3].IID = DacDataTarget.IID_ICLRRuntimeLocator;
+            wrappers[3].Vtable = ICLRRuntimeLocatorVtbl.Create(qi, addRef, release);
 
-            wrappers[4].IID = DacDataTarget.IID_ICLRSymbolProvider;
-            wrappers[4].Vtable = ICLRSymbolProviderVtbl.Create(qi, addRef, release);
+            wrappers[4].IID = DacDataTarget.IID_ICLRContractLocator;
+            wrappers[4].Vtable = ICLRContractLocatorVtbl.Create(qi, addRef, release);
 
             return wrappers;
         }
@@ -59,13 +59,10 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
         {
             DacDataTarget dacDataTarget = obj as DacDataTarget ?? throw new InvalidOperationException($"Expected {nameof(DacDataTarget)} but got {obj.GetType()}.");
 
-            // Interfaces exposed (in order): IDacDataTarget, IMetadataLocator, ICLRRuntimeLocator, ICLRContractLocator, ICLRSymbolProvider
-            // ICLRRuntimeLocator requires a valid RuntimeBaseAddress; ICLRContractLocator requires a valid ContractDescriptor.
-            // ICLRSymbolProvider is always exposed — the vtable thunks return E_FAIL when no IClrSymbolProvider is registered on the DataTarget.
-            count = 2;
+            count = 3;
             if (dacDataTarget.RuntimeBaseAddress != 0)
             {
-                count = 3;
+                count = 4;
             }
             if (dacDataTarget.ContractDescriptor != 0)
             {
@@ -292,26 +289,33 @@ namespace Microsoft.Diagnostics.Runtime.DacInterface
             [UnmanagedCallersOnly]
             private static int TryGetSymbolName(IntPtr self, ulong address, uint cchName, char* pName, uint* pcchNameActual, ulong* pDisplacement)
             {
-                DacDataTarget dacDataTarget = ComInterfaceDispatch.GetInstance<DacDataTarget>((ComInterfaceDispatch*)self);
-                IClrSymbolProvider? provider = dacDataTarget.SymbolProvider;
-                if (provider is null)
-                    return HResult.E_FAIL;
+                try
+                {
+                    DacDataTarget dacDataTarget = ComInterfaceDispatch.GetInstance<DacDataTarget>((ComInterfaceDispatch*)self);
+                    IClrSymbolProvider? provider = dacDataTarget.SymbolProvider;
+                    if (provider is null)
+                        return HResult.E_NOTIMPL;
 
-                if (!provider.TryGetSymbolName(address, out string? symbolName, out ulong displacement) || string.IsNullOrEmpty(symbolName))
-                    return HResult.E_FAIL;
+                    if (!provider.TryGetSymbolName(address, out string? symbolName, out ulong displacement))
+                        return HResult.E_FAIL;
+                    if (pcchNameActual != null)
+                        *pcchNameActual = (uint)symbolName.Length + 1;
+                    if (pDisplacement != null)
+                        *pDisplacement = displacement;
 
-                if (pcchNameActual != null)
-                    *pcchNameActual = (uint)symbolName!.Length;
-                if (pDisplacement != null)
-                    *pDisplacement = displacement;
+                    if (cchName == 0 || pName == null)
+                        return HResult.S_OK;
 
-                if (cchName == 0 || pName == null)
+                    int copy = Math.Min(symbolName.Length, (int)cchName - 1);
+                    for (int i = 0; i < copy; i++)
+                        pName[i] = symbolName[i];
+                    pName[copy] = '\0';
                     return HResult.S_OK;
-
-                int copy = Math.Min(symbolName!.Length, (int)cchName);
-                for (int i = 0; i < copy; i++)
-                    pName[i] = symbolName[i];
-                return HResult.S_OK;
+                }
+                catch
+                {
+                    return HResult.E_FAIL;
+                }
             }
 
             [UnmanagedCallersOnly]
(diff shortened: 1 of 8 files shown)


src/Microsoft.Diagnostics.Runtime/DacInterface/LegacyDacDataTargetWrapper.cs
```diff
@@ -113,6 +118,50 @@ private int GetContractDescriptor(IntPtr _, out ulong address)
             return address == 0 ? HResult.E_FAIL : HResult.S_OK;
         }
 
+        private int TryGetSymbolName(IntPtr _, ulong address, uint cchName, char* pName, uint* pcchNameActual, ulong* pDisplacement)
+        {
+            IClrSymbolProvider? provider = _dacDataTarget.SymbolProvider;
+            if (provider is null)
+                return HResult.E_FAIL;
```
leculver: E_NOIMPL would be better here.

src/Microsoft.Diagnostics.Runtime/DacInterface/LegacyDacDataTargetWrapper.cs
```diff
@@ -113,6 +118,50 @@ private int GetContractDescriptor(IntPtr _, out ulong address)
...
+
+            if (!provider.TryGetSymbolName(address, out string? symbolName, out ulong displacement) || string.IsNullOrEmpty(symbolName))
+                return HResult.E_FAIL;
+
+            if (pcchNameActual != null)
+                *pcchNameActual = (uint)symbolName!.Length;
```
leculver: Can `TryGetSymbolName` set `symbolName` with `[NotNullWhen(true)]`, so we don't need the `!`?

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
```diff
@@ -270,6 +274,70 @@ private static int GetContractDescriptor(IntPtr self, ulong* address)
...
+                return (IntPtr)vtblRaw;
+            }
+
+            [UnmanagedCallersOnly]
+            private static int TryGetSymbolName(IntPtr self, ulong address, uint cchName, char* pName, uint* pcchNameActual, ulong* pDisplacement)
+            {
+                DacDataTarget dacDataTarget = ComInterfaceDispatch.GetInstance<DacDataTarget>((ComInterfaceDispatch*)self);
+                IClrSymbolProvider? provider = dacDataTarget.SymbolProvider;
+                if (provider is null)
+                    return HResult.E_FAIL;
```
leculver: E_NOTIMPL?

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
```diff
@@ -270,6 +274,70 @@ private static int GetContractDescriptor(IntPtr self, ulong* address)
...
+                if (pDisplacement != null)
+                    *pDisplacement = displacement;
+
+                if (cchName == 0 || pName == null)
+                    return HResult.S_OK;
+
+                int copy = Math.Min(symbolName!.Length, (int)cchName);
+                for (int i = 0; i < copy; i++)
+                    pName[i] = symbolName[i];
+                return HResult.S_OK;
```
leculver: Two problems here.  If returning less than request, we should return S_FALSE.  Second, if we truncate, you need to write `\0`.

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
```diff
@@ -270,6 +274,70 @@ private static int GetContractDescriptor(IntPtr self, ulong* address)
...
+                vtblRaw[1] = addRef;
+                vtblRaw[2] = release;
+                vtblRaw[3] = (IntPtr)(delegate* unmanaged<IntPtr, ulong, uint, char*, uint*, ulong*, int>)&TryGetSymbolName;
+                vtblRaw[4] = (IntPtr)(delegate* unmanaged<IntPtr, IntPtr, ulong*, int>)&TryGetSymbolAddress;
+
+                return (IntPtr)vtblRaw;
+            }
+
+            [UnmanagedCallersOnly]
+            private static int TryGetSymbolName(IntPtr self, ulong address, uint cchName, char* pName, uint* pcchNameActual, ulong* pDisplacement)
```
leculver, line 291: Wrap TryGetSymbolName/TryGetSymbolAddress in a try/catch which returns E_FAIL, in case we hit any exceptions, otherwise this will bring down the host process.

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
```diff
@@ -49,23 +49,27 @@ public static IntPtr CreateIDacDataTarget(DacDataTarget dacData)
             wrappers[3].IID = DacDataTarget.IID_ICLRContractLocator;
             wrappers[3].Vtable = ICLRContractLocatorVtbl.Create(qi, addRef, release);
 
+            wrappers[4].IID = DacDataTarget.IID_ICLRSymbolProvider;
```
leculver: This is always exposed, so I would move this to be 2.  Then below instead of moving ContractDescriptor from 4->5 alone, you would move RuntimeBaseAddress to count 3->4.  In fact, why isn't the original code "count++"?  Something is already broken here if we have a contract descriptor and not a runtimebaseaddress.  Maybe that can't happen in practice but somethings really off here.

leculver (changes requested):
A few issues, but the idea of the change is good.

A unit test or two wouldn't hurt...  See if copilot can just generate a mock symbol interface (you don't have to spend time on tests if copilot struggles with it or if the shape of clrmd with unit tests doesn't make this feasible).  I worry that with symbol provider and the other "optional" interfaces, something is missed here.

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
```diff
@@ -270,6 +271,84 @@ private static int GetContractDescriptor(IntPtr self, ulong* address)
...
+                        return HResult.E_FAIL;
+                    if (pcchNameActual != null)
+                        *pcchNameActual = (uint)symbolName.Length + 1;
+                    if (pDisplacement != null)
+                        *pDisplacement = displacement;
+
+                    if (cchName == 0 || pName == null)
+                        return HResult.S_OK;
+
+                    int copy = Math.Min(symbolName.Length, (int)cchName - 1);
```
leculver, line 313: Can you return E_INVALIDARG (or similar) if `ccshName > int.MaxValue` at the top of the function?

src/Microsoft.Diagnostics.Runtime/DacInterface/LegacyDacDataTargetWrapper.cs
```diff
@@ -113,6 +118,68 @@ private int GetContractDescriptor(IntPtr _, out ulong address)
...
+
+                if (pcchNameActual != null)
+                    *pcchNameActual = (uint)symbolName.Length + 1;
+                if (pDisplacement != null)
+                    *pDisplacement = displacement;
+
+                if (cchName == 0 || pName == null)
+                    return HResult.S_OK;
+
+                int copy = Math.Min(symbolName.Length, (int)cchName - 1);
```
leculver, line 143: Can you return E_INVALIDARG (or similar) if `ccshName > int.MaxValue` at the top of the function?

leculver (commented):
Two quick overflows to fix, then it's ready.

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
max-charlamb: Switched to count++ in 8af9590. Agreed it's odd — kept the slot ordering with the optional interfaces (ICLRRuntimeLocator, ICLRContractLocator) appended at the end and added a comment noting that ContractDescriptor-without-RuntimeBaseAddress isn't representable. AFAICT that combination doesn't happen in practice today.

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
max-charlamb: Added a cchName > int.MaxValue E_INVALIDARG guard in 8af9590.

src/Microsoft.Diagnostics.Runtime/DacInterface/LegacyDacDataTargetWrapper.cs
max-charlamb: Added a cchName > int.MaxValue E_INVALIDARG guard in 8af9590.

max-charlamb:
Added `ClrSymbolProviderTests` in a52ceb7 — 12 tests that build the real `DacDataTargetCOM` CCW against a `SyntheticDataReader`-backed `DataTarget` with an in-memory `IClrSymbolProvider` mock, QI for `ICLRSymbolProvider`, and invoke the vtable thunks directly. Covers both `TryGetSymbolName` and `TryGetSymbolAddress` across: null provider (E_NOTIMPL), success, buffer-too-small (S_FALSE + null-terminator), zero-buffer required-size query, `cchName > int.MaxValue` (E_INVALIDARG), provider-returns-false (E_FAIL), and provider-throws (E_FAIL). All 12 pass locally.

leculver (dismissed):

hoyosjs (dismissed):

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
```diff
@@ -270,6 +272,87 @@ private static int GetContractDescriptor(IntPtr self, ulong* address)
...
+                vtblRaw[1] = addRef;
+                vtblRaw[2] = release;
+                vtblRaw[3] = (IntPtr)(delegate* unmanaged<IntPtr, ulong, ulong, uint, char*, uint*, ulong*, int>)&TryGetSymbolName;
+                vtblRaw[4] = (IntPtr)(delegate* unmanaged<IntPtr, ulong, IntPtr, ulong*, int>)&TryGetSymbolAddress;
+
+                return (IntPtr)vtblRaw;
+            }
+
+            [UnmanagedCallersOnly]
+            private static int TryGetSymbolName(IntPtr self, ulong moduleBase, ulong address, uint cchName, char* pName, uint* pcchNameActual, ulong* pDisplacement)
```
leculver: We shouldn't take a `moduleBase` AND an `address`.  It should be either `moduleBase` + `offset` (and offset has to be _somehow_ bounded by the module's contents, which we don't always have) or a pure `address` which the library resolves to the symbol.  The most common way to write this kind of method is to take a raw `address` with no `moduleBase`.

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
```diff
@@ -270,6 +272,87 @@ private static int GetContractDescriptor(IntPtr self, ulong* address)
...
+                    return copy < symbolName.Length ? HResult.S_FALSE : HResult.S_OK;
+                }
+                catch
+                {
+                    return HResult.E_FAIL;
+                }
+            }
+
+            [UnmanagedCallersOnly]
+            private static int TryGetSymbolAddress(IntPtr self, ulong moduleBase, IntPtr namePtr, ulong* pAddress)
```
leculver, line 326: Usually we would throw if `namePtr` contains a `!` if you also take in a `moduleBase`, but it's not required.

src/Microsoft.Diagnostics.Runtime/DacInterface/DacDataTargetCOM.cs
```diff
@@ -281,14 +281,14 @@ public static IntPtr Create(IntPtr qi, IntPtr addRef, IntPtr release)
...
-                vtblRaw[4] = (IntPtr)(delegate* unmanaged<IntPtr, IntPtr, ulong*, int>)&TryGetSymbolAddress;
+                vtblRaw[3] = (IntPtr)(delegate* unmanaged<IntPtr, ulong, ulong, uint, char*, uint*, ulong*, int>)&TryGetSymbolName;
+                vtblRaw[4] = (IntPtr)(delegate* unmanaged<IntPtr, ulong, IntPtr, ulong*, int>)&TryGetSymbolAddress;
 
                 return (IntPtr)vtblRaw;
             }
 
             [UnmanagedCallersOnly]
-            private static int TryGetSymbolName(IntPtr self, ulong address, uint cchName, char* pName, uint* pcchNameActual, ulong* pDisplacement)
+            private static int TryGetSymbolName(IntPtr self, ulong moduleBase, ulong address, uint cchName, char* pName, uint* pcchNameActual, ulong* pDisplacement)
```
hoyosjs: The only question is address here sounds like a VA address, so why do you need a module base?

leculver (changes requested):
I think `TryGetSymbolName` should rework the input params.

leculver (approved):
LGTM

PR #1498, opened 2026-07-17
fix(macos): int overflow when loading large segments

mircodz:
Fixes an integer overflow in `ReadMemory` caused by loading a dupm over ~2GB on MacOS platforms.

```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at Microsoft.Diagnostics.Runtime.MacOS.MachOCoreDump.ReadMemory(UInt64 address, Span`1 buffer)
   at Microsoft.Diagnostics.Runtime.MacOS.MachOCoreReader.Read(UInt64 address, Span`1 buffer)
   at Microsoft.Diagnostics.Runtime.DacInterface.DacDataTarget.ReadVirtual(ClrDataAddress cda, IntPtr buffer, Int32 bytesRequested, Int32& bytesRead)
   at Microsoft.Diagnostics.Runtime.DacInterface.DacDataTargetCOM.IDacDataTargetVtbl.ReadVirtual(IntPtr self, UInt64 address, IntPtr buffer, Int32 bytesRequested, Int32* pBytesRead)
```

### DIFF
--- a/src/Microsoft.Diagnostics.Runtime/MacOS/MachOCoreDump.cs
+++ b/src/Microsoft.Diagnostics.Runtime/MacOS/MachOCoreDump.cs
@@ -223,7 +223,7 @@ namespace Microsoft.Diagnostics.Runtime.MacOS
             while (buffer.Length > 0 && FindSegmentContaining(address, out MachOSegment seg))
             {
                 ulong offset = address - seg.Address;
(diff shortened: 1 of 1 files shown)


leculver (approved):
LGTM, thanks for the fix!

PR #1505, opened 2026-09-03
Bound Mach-O dylinker scan to segment file size

JoeRobich:
## Summary

- bound the first-pass Mach-O dylinker search by each segment's `FileSize`
- add a synthetic core dump regression test with a fake dylinker header outside the advertised segment data

## Problem

`FindDylinker(firstPass: true)` replaces the segment-bounded end offset with a fixed 32 KiB value. For a segment with less file data than that, ClrMD scans virtual offsets that are outside the segment's file-backed range. This can identify unrelated bytes as a Mach-O dylinker or fail while reading modules.

This was observed through `dotnet-dump analyze` on a macOS arm64 core, where module enumeration failed from `MachOCoreDump.ReadMemory`/`FindDylinker`.

## Validation

`dotnet test src/Microsoft.Diagnostics.Runtime.Tests/Microsoft.Diagnostics.Runtime.Tests.csproj -c Release --filter FullyQualifiedName~MachOCoreBoundsTests --nologo`

The regression test fails against the original implementation because it discovers the out-of-range fake dylinker header, and passes with this change.


leculver (approved):
Thank you for the fix!

PR #1508, opened 2026-09-10
Preserve exhausted allocation contexts during heap walks

max-charlamb:
## Summary

- Preserve non-null allocation contexts when the allocation pointer equals the allocation limit.
- Apply the same validity check to global, thread, and subheap allocation contexts.
- Update allocation-context tests to allow exhausted context markers.

## Problem

A Linux arm64 dump captured a thread after an exact-fit allocation, leaving `alloc_ptr == alloc_limit`. CoreCLR still reserves an aligned minimum-object-sized tail after the limit, and heap walkers must skip through `alloc_limit + Align(min_object_size)`.

ClrMD required `alloc_ptr < alloc_limit` when collecting allocation contexts, so it dropped the exhausted context. Heap enumeration then treated the reserved tail as an object, read its zero-filled first word as a method table, and `!verifyheap` reported false heap corruption with `InvalidMethodTable`.

CoreCLR's diagnostic heap walker retains any allocation context with a non-null pointer, including the equal pointer/limit case.

This was found by internal testing.

## Fix

Treat an allocation context as valid when its pointer is nonzero and `pointer <= limit`. ClrMD's existing allocation-context skip logic then advances over the reserved minimum-object tail correctly.


leculver (approved):
LGTM

